PR #29, opened 2019-01-07
using std::optional in favor of bool return

sbrkopac:
# Pull Request Template

## Description

Utilizing std::optional rather than a bool check for returning a user value. This allows the library to transform code from:
```
int value;
if(object->getValue("foo", value))
{
}
```
Into its c++17 equivalent:
```
auto value = object->getValue("foo");
if(value)
{
}
```

## Type of change

- [X] Breaking change (fix or feature that would cause existing functionality to not work as expected)

## Checklist:

### DIFF
--- a/include/vsg/core/Object.h
+++ b/include/vsg/core/Object.h
@@ -14,6 +14,7 @@ THE SOFTWARE IS PROVIDED "AS IS", WITHOUT WARRANTY OF ANY KIND, EXPRESS OR IMPLI
 
 #include <atomic>
 #include <string>
+#include <optional>
 
 #include <vsg/core/Export.h>
 #include <vsg/core/ref_ptr.h>
@@ -76,7 +77,7 @@ namespace vsg
         void setValue(const std::string& key, const char* value) { setValue(key, value ? std::string(value) : std::string()); }
 
         template<typename T>
-        bool getValue(const std::string& key, T& value) const;
+        std::optional<T> getValue(const std::string& key) const;
 
         void setObject(const std::string& key, Object* object);
         Object* getObject(const std::string& key);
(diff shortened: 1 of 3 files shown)


sbrkopac:
I also tagged in a fix for a typo. Let me know if I need to open a new PR for that.

openscenegraph:
This won't work.  I had to remove std::optionall use elsewhere due to portability issues.

sbrkopac:
Portability issues from where?

vsg-dev:
I think @tomhog found issues with std::optional on either the Android port.  The line of least resistance was simply to replace the std::optional usage.

Could you generate a PR for the typo fix. Thanks.

vsg-dev:
I'm closing this PR as it risks cross platform build.  I have applied the typo fix:

https://github.com/vsg-dev/VulkanSceneGraphPrototype/commit/5d05f3f13f93a762c06c89407ab4fdab180001eb

sbrkopac:
I know it's difficult with the speed at which you are prototyping, but is it possible to have a blurb somewhere about the different c++17 features that can't be used due to X reason?

vsg-dev:
I have wondered about writing a C++ usage document, from the perspective of what bits of C++17 etc. are used where, but adding what parts don't currently work well across platforms would be a good addition too.  I can't promise this right away but I'll continue thinking about it.

tomhog:
Hi @sbrkopac 

So the issue at the moment is on macOS std::optional and std::any are currently in the std::experimental namespace (I'm assuming it'll be the same for iOS). So for now we've avoided them to prevent the code requiring conditional compilation in lots of places.

Hopefully apply move these into the standard std namespace in the near future.

PR #53, opened 2019-04-01
Read write methods for states

tomhog:
## Description

Added read write methods for inputassembly and depthstencil state. Added a contructor to InputAssemblyState to pass topology

### DIFF
--- a/include/vsg/vk/GraphicsPipeline.h
+++ b/include/vsg/vk/GraphicsPipeline.h
@@ -200,7 +200,7 @@ namespace vsg
     {
     public:
         InputAssemblyState();
-        InputAssemblyState(const VkPrimitiveTopology& primitiveTopology, const bool& enablePrimitiveRestart = false);
+        InputAssemblyState(VkPrimitiveTopology primitiveTopology, bool enablePrimitiveRestart = false);
 
         void read(Input& input) override;
         void write(Output& output) const override;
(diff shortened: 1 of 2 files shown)


include/vsg/vk/GraphicsPipeline.h
```diff
@@ -200,6 +200,7 @@ namespace vsg
     {
     public:
         InputAssemblyState();
+        InputAssemblyState(const VkPrimitiveTopology& primitiveTopology, const bool& enablePrimitiveRestart = false);
```
robertosfield: I don't understand why you've chosen to pass const& rather than just as simple type.  It's only makes sense to pass large objects by reference.  Have a look at CppCoreGuidelines to see a discussion on this topic.

Could you change the code to remove the use of const and &.  I'll then merge it.  Thanks.

tomhog: Just changed it. Sorry about that.Just gone on auto pilot I think.

Regarding these contructors. I added one for InputAssembly as it's only got two variables. 
How about things like DepthStencil state, do you think they should have contructors with
the options? They have a fair few more variables so I didn't do it for now

robertosfield: Having useful constructors that help keep the scene graph creation code short and more descriptive is a win for me.  Too many parameters on a command line can end up unwieldy so one has to look at it on a case by case basis.

PR #61, opened 2019-05-13
expose index type, read/write support

tomhog:
# Pull Request Template

## Description

Added support to set the index format for Geometry, VertexDrawIndexed and BindIndexBuffer, also added read write support for the value in each.

Done against the state decoupling branch

### DIFF
--- a/include/vsg/nodes/Geometry.h
+++ b/include/vsg/nodes/Geometry.h
@@ -43,6 +43,7 @@ namespace vsg
         DataList _arrays;
         ref_ptr<Data> _indices;
         DrawCommands _commands;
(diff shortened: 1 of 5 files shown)


vsg-dev:
It's been my intention to automatically compute the VkIndexType from the Data value type rather than require it to be set/read/written explicitly.  So rather than merge this PR I've implemented automatic computation of the VkIndexType on assignment of Data/BufferData to the BindIndexBuffer:

  https://github.com/vsg-dev/VulkanSceneGraph/commit/28ce7b04c5a6e1c0d6bf64b4ec86fbd0e8ac37be

I'll be merging the StateDecoupling work with master of VSG, osg2vsg and vsgExamples next so all these changes will be part of master.

PR #67, opened 2019-05-20
Added 16 and 32 bit int vecs and vec array types

tomhog:
# Pull Request Template

## Description

I've added usvec2/3/4 (vecs of uint16) and uivec2/3/4 (vec of uint32), then I've added array and array 2ds of these to help match more vulkan formats.


tomhog:
Hey Robert

I've edited this commit to remove the static_casts, it just contains the ui and us vec types now

PR #76, opened 2019-06-10
Updated all.h, fix for DescriptorImages fixes

tomhog:
# Pull Request Template

## Description

Compile would always return early due to _imageInfos and _imageDataList both being zero. Fixed a copy and paste error in DescriptorImage compile

## Type of change

## How Has This Been Tested?

Tested with vsgUnity terrain export, loaded export into vsgviewer


vsg-dev:
I've applied similar changes in the DescriptorBuffer branch which are now checked into master so I'll close this PR. Thanks.

PR #86, opened 2019-07-21
Fix for "overrides a member function but is not marked 'override'" warning

olegded:


include/vsg/vk/Descriptor.h
```diff
@@ -160,7 +160,7 @@ namespace vsg
             }
         }
 
-        virtual bool assignTo(VkWriteDescriptorSet& wds, VkDescriptorSet descriptorSet) const
+        virtual bool assignTo(VkWriteDescriptorSet& wds, VkDescriptorSet descriptorSet) const override
```
vsg-dev: Good catch, though you forgot to remove the virtual ;-)

I'll do this once I've merged this PR.

PR #101, opened 2019-08-17
WIP: avoid using the >> operator as it truncates strings with spaces

### LINKED ISSUE #100: Command Line parsing truncates string parameters
**Describe the bug**
Command Line parsing truncates string parameters

**To Reproduce**
```
#include <vsg/all.h>

int main(int argc, char** argv)
{
    // set up defaults and read command line arguments to override them
    vsg::CommandLine arguments(&argc, argv);
    auto test = arguments.value(std::string(), { "--test" });
    if (arguments.errors()) return arguments.writeErrorMessages(std::cerr);

    std::cout << test << std::endl;

    return 0;
}
```

Ran with the following:
```
foo.exe --test "C:\This\Is\A Path"
```

**Expected behavior**
```
EXPECTED: test == C:\This\Is\A Path
ACTUAL: test == C:\This\Is\A
```

 - OS: Windows
 - Version: master

sbrkopac:
# Pull Request Template

## Description

Fixes #100

## Type of change

## How Has This Been Tested?

See #100 for tests and expected results

**Test Configuration**:
* Hardware: Windows 10 
* Toolchain: Visual Studio 2019 - 16.2.2

## Checklist:

### DIFF
--- a/include/vsg/utils/CommandLine.h
+++ b/include/vsg/utils/CommandLine.h
@@ -67,9 +67,16 @@ namespace vsg
 
                 _istr.str(str);
             }
-            
-            // don't use the >> operator here as it will truncate strings if a space is found
-            value = _istr.str();
+
+            // check if we are dealing with a string as the >> operator will truncate on a space
+            if constexpr (std::is_same_v<T, std::string>)
+            {
+                value = _istr.str();
+            }
+            else
+            {
+                _istr >> value;
+            }
 
             return (!_istr.fail());
         }
--- a/include/vsg/utils/CommandLine.h
+++ b/include/vsg/utils/CommandLine.h
@@ -67,7 +67,9 @@ namespace vsg
 
                 _istr.str(str);
             }
-            _istr >> value;
+            
+            // don't use the >> operator here as it will truncate strings if a space is found
+            value = _istr.str();
 
             return (!_istr.fail());
         }


openscenegraph:
I believe this fix will only compile for cases where users are reading std::string, it'll fail for other usage cases.  It's these other usage cases that are why the istringstream is used - it enables the parsing to read values to int, float, vec2,vec3, mat4 etc.

I haven't yet worked out how to fix it.  It might require a specific read(int&i , string&) implementation.  Such a specific implementation could be pretty simple as the code wont have to try and pack multiple command line parameters into the string, instead it should be able to just convert it.

I won't close this PR quite yet, give you an opportunity to chip and depend it in-case I've overlooked something.  Feel free to close it if you want.

sbrkopac:
Oh, that's a good point. A little over zealous with that fix. I'll leave it open for now and give it some more thought.

sbrkopac:
The above fix works on windows and I've updated the issue with a new test program.
```

openscenegraph:
Neat use of constexp in the new version :-)

However, I more inclined to do a wholly different handling of string that avoids the packing into the istringstream and just passing the _argv value directly to the string, a template specialization for string might be a more readable way to do this, though you could do it with constexpr done higher up into the code.

vsg-dev:
The above OSG github was just me (robert) forgetting to log out/log in after doing some OSG maintenance... 

I have just checked in the follow change, the diff looks more serious code changes that there really were:
https://github.com/vsg-dev/VulkanSceneGraph/commit/4f7b886d01a686e30647739aeff0ea6e50cc6060

All I did was add near the top of the read method was:

            if constexpr (std::is_same_v<T, std::string>)
            {
                value = _argv[i++];
                return true;
            }
            else
            {

And then indented the old code that now sits within the else block. 

I prefer this approach as it'll avoid the compiler parsing and generating more code that boils down to a non op operationally, and will run more efficiently at run time as well.

I've test with the suggested --test "string with spaces" and also with parsing of a vec2 to test the original code paths and both now work.

PR #105, opened 2019-08-22
fix compiler warning about size_t to unsigned int conversion.

LaurensVoerman:
fix compiler warning about size_t to unsigned int conversion;
changed Latch constructor to take unsigned int instead of uint_32_t to match std::atomic_uint

## Description

fix compiler warning about size_t to uint_32_t with a static_cast - assuming the number of files will never exceed MAX_UINT. Then changed the uint_32_t in the cast and the Latch constructor to match the underling type of std::atomic_uint (unsigned int) to minimize surprise when the size of uint_32_t and unsigned int are not identical.

Fixes # (issue)

## Type of change

- [ x] Bug fix (non-breaking change which fixes an issue)
## How Has This Been Tested?

Not tested; all my compilers have 32 bit int so binaries will be unchanged.

**Test Configuration**:
* Firmware version:
* Hardware:
* Toolchain: Visual Studio 2019 16.2.3
* SDK:

## Checklist:

- [x ] My code follows the style guidelines of this project
- [x ] I have performed a self-review of my own code

### DIFF
--- a/include/vsg/threading/OperationQueue.h
+++ b/include/vsg/threading/OperationQueue.h
@@ -35,7 +35,7 @@ namespace vsg
 
     struct Latch : public Object
     {
-        Latch(uint32_t num) :
+        Latch(unsigned int num) :
             count(num) {}
 
         void count_down()
--- a/src/vsg/io/read.cpp
+++ b/src/vsg/io/read.cpp
@@ -95,7 +95,7 @@ PathObjects vsg::read(const Paths& filenames, ref_ptr<const Options> options)
         };
 
         // use latch to syncronize this thread with the file reading threads
-        ref_ptr<Latch> latch(new Latch(filenames.size()));
+        ref_ptr<Latch> latch(new Latch(static_cast<unsigned int>(filenames.size())));
 
         // add operations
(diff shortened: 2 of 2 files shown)


vsg-dev:
I wonder if using atomic_size_t might be an alternative.  It's not necessary in terms of capabilities but might fit better to how the Latch would normally be used - passed in a size of container.  I will have a think.

LaurensVoerman:
I Agree, atomic_size_t is a better match. I created #PR106 so this pr can be closed.
Regards, Laurens.

vsg-dev:
This morning I wrote a test whether atomic_size_t has similar performance to atomic_uint and on my AMD system they look to essentially perform the same, 2-3 times faster than protecting a value using mutexes.  In all cases contention between threads trying to modify the atomic value has a big impact on performance.  

The main difference I saw between atomic_size_t, atomic_uint  and mutex was the size differences - 8, 4 and 40 bytes respectively.  In the case of the vsg::Latch changing to atomic_size_t didn't affect the overall size due to the alignment.  All in all it looks like the size_t Latch is the way to go so I've gone ahead and merged your other PR.

PR #109, opened 2019-08-26
Support for user defined extensions when creating device and instance

tomhog:
# Pull Request Template

## Description

User can now pass extention strings via Window Traits to be used when initalizing the instance and device.

DeviceMemory create function can now be passed an optional pNext value for using extensions

## Type of change

- [x ] New feature (non-breaking change which adds functionality)


src/vsg/io/read.cpp
```diff
@@ -95,7 +95,7 @@ PathObjects vsg::read(const Paths& filenames, ref_ptr<const Options> options)
         };
 
         // use latch to syncronize this thread with the file reading threads
-        ref_ptr<Latch> latch(new Latch(filenames.size()));
+        ref_ptr<Latch> latch(new Latch(static_cast<uint32_t>(filenames.size())));
```
vsg-dev: We've fixed this another way - changing the Latch to use an atomic_size_t.  Could you revert this commit and update to VSG master.

tomhog:
Reverted that change and updated to ensure things are still merging

PR #120, opened 2019-11-04
Support for multiple geometry instances in a tlas, improved compile. Basic build visitor

tomhog:


src/vsg/vk/Submit.cpp
```diff
@@ -24,10 +24,10 @@ Submit::~Submit()
 
 void Submit::submit(Queue& queue)
 {
-    VkSubmitInfo submitInfo = {};
+    /*VkSubmitInfo submitInfo = {};
```
vsg-dev: Is this a bit of debugging experimentation that wasn't meant to be checked in?

tomhog:
Hey Robert

It's a weird one I meant to mentioned that. The submit header suddenly
started throwing errors when I was compiling the example.
When I looked at the file it was a mix of Submit, Submission for the class
name and file and then the functions where using variables
that didn't exist so I just got it to compile and forgot about it..

Is it actually used? I'm not using it

vsg-dev:
I'll look into Submit, perhaps I ended up doing something half backed then didn't gone down route fully.

vsg-dev:
OK, looks like Submit was an experimental approach that I didn't follow through and shouldn't have check it in.  I'll merge this submission then remove this class.  Later we might want something like this but I can reimplement when it actually becomes needed.

PR #121, opened 2019-11-06
prevent the image class deleting the vkimages owned by the swapchain

tomhog:
So the issue is that if we want to use a vsg Image with the swap chain ImageViews the vsg Image
class will delete the VkImage which is owned by the swap chain and seemingly deleted by vkDestroySwapchainKHR.

Previously around line 230 ish of SwapChain.cpp the swapchain image was wrapped in an image view by directly passing the VkImage to the ImageView create function.

However doing that there is no way to get hold of the image and on top on that things like ImageMemoryBarrier expect a vsg Image.

So for now this release function seems like the simplest solution. Otherwise I think we'd need ImageView to be able to optionally store a raw VkImage and create different functions for ImageMemoryBarrier  etc etc

### DIFF
--- a/include/vsg/vk/Image.h
+++ b/include/vsg/vk/Image.h
@@ -31,6 +31,8 @@ namespace vsg
         Device* getDevice() { return _device; }
         const Device* getDevice() const { return _device; }
 
+        void releaseImage() { _image = VK_NULL_HANDLE; }
+
         VkResult bind(DeviceMemory* deviceMemory, VkDeviceSize memoryOffset)
         {
             VkResult result = vkBindImageMemory(*_device, _image, *deviceMemory, memoryOffset);
--- a/src/vsg/vk/Swapchain.cpp
+++ b/src/vsg/vk/Swapchain.cpp
@@ -136,6 +136,10 @@ Swapchain::Swapchain(VkSwapchainKHR swapchain, Device* device, Surface* surface,
 
 Swapchain::~Swapchain()
 {
+    for (auto& imgview : _imageViews)
+    {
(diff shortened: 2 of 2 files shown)


vsg-dev:
I have reviewed the change and looked into the way the Swapchain/ImageView/Image are now linked in the topic-raytracing branch.  The proposed fix will work but I'm a bit torn on whether it's the best solution.  

Logically it would seem like a vsg::Image should have some kind of mechanism for tracking who owns the VkImage that it's associated with.  In the case of Swapchain Images then it's the VkSwapchain.  You can have swapchains that share their presentable images:

    https://manned.org/vkCreateSharedSwapchainsKHR/e7c61d26

The VSG does support the above yet, not even sure where it'll be needed, but I guess this will require some kind of vsg::SharedSwapchain class...

Anyway a bit of distraction... for now need to decide what to do about the Image that Swapchain creates.  One possible route is to have a vsg::SwapchainImage/PresentableImage class that subclasses from Image and disables the vkDestroyImage.

I'll pond this some more.  Feedback welcome.

tomhog:
The vsg::SwapchainImage/PresentableImage image makes sense. I was pondering
some kind of owner ship but was thinking along the lines of how we release
the Array data.
A SwapChainImage class makes sense. Their may be other things that are
special' about swapchain images well need to account for in the future.

At least that last pull request made it clear what the issue was ha.

vsg-dev:
Glad you agree that a SwapchainImage subclass makes sense as I just implemented while and checked it in!  I'll close this PR now.

PR #123, opened 2019-11-08
Added setimagelayoutcommand

tomhog:
# Pull Request Template

## Description

When directly using an imageview we still need to set the layout

Also did the same merge of master as you I think, so got a little mixed but is working and the
code in viewer will go eventually anyway


src/vsg/viewer/Viewer.cpp
```diff
@@ -230,9 +230,9 @@ bool Viewer::submitNextFrame(std::vector<VkSemaphore> externalWaits, std::vector
         std::vector<VkSemaphore> waitSemaphores;
         waitSemaphores.insert(waitSemaphores.end(), externalWaits.begin(), externalWaits.end());
 
-        std::vector<VkPipelineStageFlags> waitStages;
-        waitStages.insert(waitStages.end(), externalWaitStages.begin(), externalWaitStages.end());
-        waitStages.insert(waitStages.end(), pdo.waitStages.begin(), pdo.waitStages.end());
+        std::vector<VkPipelineStageFlags> waitDstStageMasks;
+        waitDstStageMasks.insert(waitDstStageMasks.end(), externalWaitStages.begin(), externalWaitStages.end());
+        //waitDstStageMasks.insert(waitDstStageMasks.end(), pdo.waitStages.begin(), pdo.waitStages.end());
```
vsg-dev: Why is this line commented out?

tomhog:
For some reason i thought i saw you’re not using that style in master
anymore. I’ll double check but things are still working.

vsg-dev:
I've looked into the commented line, this line looks to be inappropriate anyway so it's correct to comment it.  In master and now topic-glinterop branch I've removed the pdo.waitStages variable completely as well as any code that sets it.  So please update.

PR #128, opened 2020-01-31
add a DAG to manage SubPasses and Dependencies

mp3butcher:
## Add a graph abstraction for RenderPass preparation
cool feature not to have to maintain syntaxic coherence in renderpass design 
but Not hardy tested and very inlined

### DIFF
--- a/include/vsg/vk/PassGraph.h
+++ b/include/vsg/vk/PassGraph.h
@@ -1,0 +1,361 @@
+#pragma once
+
+/* <editor-fold desc="MIT License">
+
+Copyright(c) 2018 Robert Osfield
+Copyright(c) 2020 Julien Valentin
+
+Permission is hereby granted, free of charge, to any person obtaining a copy of this software and associated documentation files (the "Software"), to deal in the Software without restriction, including without limitation the rights to use, copy, modify, merge, publish, distribute, sublicense, and/or sell copies of the Software, and to permit persons to whom the Software is furnished to do so, subject to the following conditions:
+
+The above copyright notice and this permission notice shall be included in all copies or substantial portions of the Software.
+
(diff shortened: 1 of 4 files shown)


robertosfield:
I have just done a quick review, not enough time with it yet to understand it, but seen enough to know that you probably need to be aware of some fundamental changes to how the Viewer related classes are managed/configured.  The work is almost ready to merge with master - I would have done it already if I hadn't been busy with OSG-3.6.5 release.  The work can be found in the branch:

    https://github.com/vsg-dev/VulkanSceneGraph/tree/raytracing-refactor

The big change is replacing the Stage/GraphicsStage with a new graph approach that allows you to manage the recording and submission of command graphs to queue in a more flexible way.  There are three new key classes, top level is the RecordAndSubmitTask, this encloses a list of CommandGraph, which enclose a straight graph of nodes or enclosed the new RenderGraph node. 

The CommandGraph manages the command buffer creation and record to, while the RenderGraph manages the RenderPass side of things required for graphics work.

I haven't tried it yet, but in theory it should be possible to have multiple RenderGraph with different RenderPass setting within a single CommandGraph, or have multiple CommandGraph each with a nested RenderGaph.  One can assign semaphores to coordinate between command buffer submissions down on the GPU and Fence to coordinate with the CPU side.  One can also insert Semaphore related commands into the scene graph as well.

These changes required vsgExamples to be updated, these changes are all in the raytracing-refactor branch of vsgExamples.   The vsgdraw is now a little different w.r.t viewer setup:

    https://github.com/vsg-dev/vsgExamples/blob/raytracing-refactor/Desktop/vsgdraw/vsgdraw.cpp

These changes bring the VSG a bit closer to how Vulkan itself manages things, rather than trying to shoe horn into a more OSG style approach that the VSG viewer setup had (although still pretty loose relationship.)

Now, I don't know how this might interact with the changes that you have made.  For you own changes I think it would be worth re-basing them on the raytracing-refactor branch.  The raytracing part isn't actually the important part for you moving across to this branch, so you can just focus on the changes to the viewer.  Supporting raytracing coherently is why I ended up tackling the viewer refactor right now, so this is why it's all bundled in this branch that combines Tom's work on ray tracing and my recent refactor of Viewer and changes to Tom's ray tracing work to fit with this.

I hope this make sense. 

If you want to wait till early next week then I can get the ratracing-refactor branch merged with master and you could then just get your work build against it.  

Cheers,
Robert.

mp3butcher:
at first glance changes seems othogonal with the raytrace_branch...

openscenegraph:
The best way forward on this PR would be for me to create a subpass branch on VulkanSceneGraph and vsgfExamples to merge your changes into, then have you provide an example in vsgExample that tests this functionality so we can evaluate how the API would look in usage and it's test it works as intended.

robertosfield:
I have created a subpass branch for the VulkanSceneGraph and vsgExamples based on the raytracing_refactor branch.

PR #129, opened 2020-02-02
add a Graph to manage SubPasses and Dependencies

mp3butcher:
Add a graph abstraction for RenderPass preparation

cool feature not to have to maintain syntaxic coherence in renderpass design
but Not hardy tested (only one pass tested) and very inlined


mp3butcher:
example here
https://github.com/robertosfield/vsgFramework/commit/22820c23fa1daa7a4d3c81140472981466ef26da

vsg-dev:
I have yet to establish and understanding or opinion on the approach taken here, but as it's a branch I'm happy to merge to see how it looks in practice.  My own goal will be to understand the problem you are trying to solve and why you've tackled it the way you have.  

Once I have a better understanding I will have a better ability to decide what the best way forward might be, it could be that merge with master as is, through to tackling the desired functionality in a completely different way.

mp3butcher:
According my opinion, only the 2 last commits are contraversals....
There's a design problem that must be addressed...subpasses are linked to secondary cmdbuffers which are linked to cmdbuffer production thread policy

vsg-dev:
I have to admit I haven't yet found my bearings with what you are doing with this work.  Nothing has yet just clicked as to it being the right way to do things, and I'm don't know where this work is heading.  I feel a bit like I neither have a map nor a compass bearing.

vsg-dev:
Now that the raytracing and viewer refactor is part of master I'll be moving over to getting multi-GPU support working, this is quite a big topic, so what I'll do is look to get the subpass functionality merged in some form over the next week before I stack the multi-GPU side.

I'll use your work as a reference, but as I haven't seen a working example of your approach yet this would be good to get working so I have something to work to recreate if I opt for a different approach to the one you took. i.e. I want to make sure the functionality required is in place, not matter the interface & implementation details.

mp3butcher:
I hacked a lot but i believe there's a bug with mesa and secondaries
command buffer...not sure perhaps i miss something, but it stopped me...
It should work but on my intelchipset, but  i have neither graphic result
neither validation warnings :/

If you could just try to run vsg subpass on your driver...and tell me if
something is displayed i would be glad

the hacked branch
https://github.com/mp3butcher/VulkanSceneGraphPrototype/tree/subpass
and https://github.com/mp3butcher/vsgExamples/tree/subpass

Cheers




Le sam. 8 févr. 2020 à 11:22, Vulkan made easy <notifications@github.com> a
écrit :

PR #130, opened 2020-02-02
api version check for vkGetPhysicalDeviceProperties2

mp3butcher:


src/vsg/vk/PhysicalDevice.cpp
```diff
@@ -30,12 +33,17 @@ PhysicalDevice::PhysicalDevice(Instance* instance, VkPhysicalDevice device, int
...
     _rayTracingProperties.pNext = nullptr;
-
-    // do we check for extension support?
-    VkPhysicalDeviceProperties2 deviceProps2;
-    deviceProps2.sType = VK_STRUCTURE_TYPE_PHYSICAL_DEVICE_PROPERTIES_2;
-    deviceProps2.pNext = &_rayTracingProperties;
-    vkGetPhysicalDeviceProperties2(_device, &deviceProps2);
+    
+    if( VK_VERSION_MINOR(_properties.apiVersion) > 1  ||
+       ( VK_VERSION_MINOR(_properties.apiVersion) == 1 && VK_VERSION_MINOR(_properties.apiVersion) >= 2 ))
```
openscenegraph: Did you mean to use something different there?? As it looks like it will always evaluate to false....

src/vsg/vk/PhysicalDevice.cpp
```diff
@@ -15,6 +15,9 @@ THE SOFTWARE IS PROVIDED "AS IS", WITHOUT WARRANTY OF ANY KIND, EXPRESS OR IMPLI
 #include <iostream>
 
 using namespace vsg;
+#define VK_VERSION_MAJOR(version) ((uint32_t)(version) >> 22)
+#define VK_VERSION_MINOR(version) (((uint32_t)(version) >> 12) & 0x3ff)
+#define VK_VERSION_PATCH(version) ((uint32_t)(version) & 0xfff)
```
robertosfield: Vulkan provides these defines in the vulkan_core.h header.

src/vsg/vk/PhysicalDevice.cpp
robertosfield: Looking into the headers and Khronos docs the right way to check if a version is equal or greater to a specified version would be to compare the apiVersion to VK_MAKE_VERSION(x,y,z)

mp3butcher:
I admit i only pr to point the issue....will fix it

openscenegraph:
On Mon, 3 Feb 2020 at 11:42, Julien Valentin <notifications@github.com>
wrote:

> I admit i only pr to point the issue....will fix it
>
That's what the issue tracker is for :-)

FYI, I'm review the set up of PhysicalDevice/Device/Window and Queues.

BTW, what hardware, OS, drivers are you working with?  I am currently using
NVidia's Linux drivers + VulkanSDK 1.2.131 from LunarG with a Geforce 2060.

mp3butcher:
ATM lowended :
Intel Chipset/mesa  
Vulkan v 1.1.126

robertosfield:
Curious, ATM?

I have a old Pixel chromebook with an iCore7 with Kubuntu install using Crouton, perhaps it's time I tried to get it work with Vulkan...

robertosfield:
What version of Mesa are you using?

mp3butcher:
ATM=At the moment (the old Fermi card on this computer didn't support vulkan..had to remove it)

Ubuntu18.04.1 / Wayland/ Mesa 19.2.8
Intel Chipset/mesa
Vulkan v 1.1.126

openscenegraph:
On Mon, 3 Feb 2020 at 15:33, Julien Valentin <notifications@github.com>
wrote:

> ATM=At the moment (the old Fermi card on this computer didn't support
> vulkan..had to remove it)
>
I too got caught out with an older perfectly decent Fermi card that I had
to put into early retirement to do the Vulkan work.

vsg-dev:
What errors are you seeing that this PR workaround?  I would like to wrap up the raytracing_refactor branch and get it checked into master but don't want to break support for driver/hardware combinations that I don't have available to test.

vsg-dev:
@mp3butcher I'm currently looking at the PhysicalDevice class, while my review of it is wider than just adding a version check for the raytracing functionality, I am considering how best to set up the various properties of a device.  In the case of the RTX functionality I don't think a test against a specific version of Vulkan would be appropriate, instead I think it needs to be done based on checking for the availability of extensions.

To help guide how to resolve this it would be useful to know what problems you've seen on your system that made you implement the above workaround, and how this workaround helped.

mp3butcher:
> What errors are you seeing that this PR workaround?
Don't have full knowledge of vsg ..so:
- The decouplign RenderPass/PassGraph is just here to minimize changes (it could be merged)
- The CreateRenderPassCallback is quite an arbitrary design
-expose NextSubPass command to the use may not be a good approach as Renderpass is vsg managed 


My guess is sceondaries commandbuffer/and or subpass may be scenegraph semantized (compile time) in order to give the attended command interleaving:
BeginRP 
BeginCmdBuffer 
//firstpass
BeginSecondaryCmdBuffer[0]
BeginSecondaryCmdBuffer[1..]
...
EndSecondaryCmdBuffer[1..]
EndSecondaryCmdBuffer[0]
(..implemented as vkCmdExecuteCommands..)
...
vkCmdNextSubPass
//otherpasses...

EndCmdBuffer
EndRenderPass

vsg-dev:
Thanks for the explanation of the render pass functionality, though it's attached to the wrong PR :-)  This explanation make it a bit clearer. FYI, begin/end of the CommandBuffer enclose the begin/end of the RenderPass. which is why the new vsg::CommandGraph (scene graph manager of CommandBuffer) contains/ecloses the RenderGraph (scene graph manager of RenderPass).   Once I've settled the PhysicalDevice AP/implenentation changes I have another look over the subpass branch.

Right now, I'm still working on changes to the PhysicalDevice class and don't yet know what specific issue you are seeing with the RayTracing functionality added to PhysicalDevice.  In this PR you've added a version check, which I'm pretty sure is not correct for what you want to achieve, but I'm no wiser on what specific problems you had on your system.  Was it a compile time or runtime problem?

This is crucial for me to understand so I can implement a scheme that works correctly across systems.

vsg-dev:
I am still working on PhysicalDevice, but have checked in a small cleanup of how the properties are set up so the PhysicalDevice,cpp now looks a bit simpler, so if we do need to do some extra checks it'll be more obvious what parts need to be tweaked.  The commit is:

https://github.com/vsg-dev/VulkanSceneGraph/commit/23c797d6c4de53d11fcfe121ad01734057da498d

PR #132, opened 2020-02-08
Fixed build break on macOS by calling `Instance::getPhysicalDevice()`

olegded:
I hope it's the right change, examples are running.

I think same change needs to happen here as well https://github.com/vsg-dev/VulkanSceneGraph/blob/master/src/vsg/platform/win32/Win32_Window.cpp#L462

### DIFF
--- a/src/vsg/platform/macos/MacOS_Window.mm
+++ b/src/vsg/platform/macos/MacOS_Window.mm
@@ -858,7 +858,7 @@ MacOS_Window::MacOS_Window(vsg::ref_ptr<vsg::Window::Traits> traits, vsg::Alloca
         // temporary hack to force vkGetPhysicalDeviceSurfaceSupportKHR to be called as the Vulkan
         // debug layer is complaining about vkGetPhysicalDeviceSurfaceSupportKHR not being called
         // for this _surface prior to swap chain creation
-        vsg::ref_ptr<vsg::PhysicalDevice> physicalDevice = vsg::PhysicalDevice::create(traits->shareWindow->instance(), VK_QUEUE_GRAPHICS_BIT, surface);
+        vsg::ref_ptr<vsg::PhysicalDevice> physicalDevice = traits->shareWindow->instance()->getPhysicalDevice(VK_QUEUE_GRAPHICS_BIT, surface);
     }
     else
     {


vsg-dev:
Thanks for the fix. This bit of code is something that Tom added as a workaround back in 2018.  There is chance that the Vulkan driver is a bit more solid now and doesn't require this workaround.  Could you try commenting out this line completely?  If there are errors could you let me know what they are.

Unfortunately I don't have an OSX box to test this out myself.

olegded:
It seems to work, at least the examples start. However, right now I couldn't  really say if there some side effects or not.

vsg-dev:
Looking at the comment it looks like the issue was with the debug layer, so try running :

    vsgdraw -d

Also looking at the comment is suggests that it's the original vkGetPhysicalDeviceSurfaceSupportKHR done inside the PhysicalDevice::create() that was the bit that is the effective part of the workaround, PhysicalDevice now longer has this call so even calling Instance::getPhysicalDevice() won't provoke the above call.

If the debug layer isn't working then a different workaround will be required, such as explictly calling the vkGetPhysicalDeviceSurfaceSupportKHR in the MacOS_Window.mm code,

BTW, does vsgcompute work for you?  As there is Issue logged on vsgExamples:

   https://github.com/vsg-dev/vsgExamples/issues/22

vsgcompute just runs in the console and returns a couple of ms second later having used a compute shader to create a Mandelbrot image, this can be output to a .vsgb or .vsgt file using:

     osgcompute -o test.vsgt

vsg-dev:
While I don't have Win32 or MacOS to test I got the code attempted in this "fix" compiling under Linux so I just copied and pasted across:

https://github.com/vsg-dev/VulkanSceneGraph/commit/35fee142e594b707cad0044c5e849e690b67862c

Could you test out this change and see how you get on.  Fingers crossed build will be back working under Windows and MacOS once more.  BTW, the new changes avoids creating any new temporary objects on the heap, instead the query that we want is done ise the PhysicalDevice that is now constructed on creation of the vsg::/vkInstance.

olegded:
Ran two tests using `-d` switch: https://github.com/vsg-dev/VulkanSceneGraph/commit/35fee142e594b707cad0044c5e849e690b67862c and commented this line out. Same results - no messages on start, following in minimizing

```shell
VUID-VkSwapchainCreateInfoKHR-imageExtent-01689(ERROR / SPEC): msgNum: 0 - vkCreateSwapchainKHR(): pCreateInfo->imageExtent = (0, 0) which is illegal. The Vulkan spec states: imageExtent members width and height must both be non-zero (https://www.khronos.org/registry/vulkan/specs/1.1-extensions/html/vkspec.html#VUID-VkSwapchainCreateInfoKHR-imageExtent-01689)
    Objects: 1
        [0] 0x16b38850080, type: 3, name: NULL
VUID-VkImageCreateInfo-extent-00944(ERROR / SPEC): msgNum: 0 - vkCreateImage: parameter pCreateInfo->extent.width (= 0) is greater than 0 The Vulkan spec states: extent.width must be greater than 0. (https://www.khronos.org/registry/vulkan/specs/1.1-extensions/html/vkspec.html#VUID-VkImageCreateInfo-extent-00944)
    Objects: 1
        [0] 0, type: 10, name: NULL
VUID-VkImageCreateInfo-extent-00945(ERROR / SPEC): msgNum: 0 - vkCreateImage: parameter pCreateInfo->extent.height (= 0) is greater than 0 The Vulkan spec states: extent.height must be greater than 0. (https://www.khronos.org/registry/vulkan/specs/1.1-extensions/html/vkspec.html#VUID-VkImageCreateInfo-extent-00945)
    Objects: 1
        [0] 0, type: 10, name: NULL
VUID-VkFramebufferCreateInfo-width-00885(ERROR / SPEC): msgNum: 0 - vkCreateFramebuffer(): Requested VkFramebufferCreateInfo width must be greater than zero. The Vulkan spec states: width must be greater than 0. (https://www.khronos.org/registry/vulkan/specs/1.1-extensions/html/vkspec.html#VUID-VkFramebufferCreateInfo-width-00885)
    Objects: 1
        [0] 0, type: 0, name: NULL
VUID-VkFramebufferCreateInfo-height-00887(ERROR / SPEC): msgNum: 0 - vkCreateFramebuffer(): Requested VkFramebufferCreateInfo height must be greater than zero. The Vulkan spec states: height must be greater than 0. (https://www.khronos.org/registry/vulkan/specs/1.1-extensions/html/vkspec.html#VUID-VkFramebufferCreateInfo-height-00887)
    Objects: 1
        [0] 0, type: 0, name: NULL
VUID-VkFramebufferCreateInfo-width-00885(ERROR / SPEC): msgNum: 0 - vkCreateFramebuffer(): Requested VkFramebufferCreateInfo width must be greater than zero. The Vulkan spec states: width must be greater than 0. (https://www.khronos.org/registry/vulkan/specs/1.1-extensions/html/vkspec.html#VUID-VkFramebufferCreateInfo-width-00885)
    Objects: 1
        [0] 0, type: 0, name: NULL
VUID-VkFramebufferCreateInfo-height-00887(ERROR / SPEC): msgNum: 0 - vkCreateFramebuffer(): Requested VkFramebufferCreateInfo height must be greater than zero. The Vulkan spec states: height must be greater than 0. (https://www.khronos.org/registry/vulkan/specs/1.1-extensions/html/vkspec.html#VUID-VkFramebufferCreateInfo-height-00887)
    Objects: 1
        [0] 0, type: 0, name: NULL
VUID-VkFramebufferCreateInfo-width-00885(ERROR / SPEC): msgNum: 0 - vkCreateFramebuffer(): Requested VkFramebufferCreateInfo width must be greater than zero. The Vulkan spec states: width must be greater than 0. (https://www.khronos.org/registry/vulkan/specs/1.1-extensions/html/vkspec.html#VUID-VkFramebufferCreateInfo-width-00885)
    Objects: 1
        [0] 0, type: 0, name: NULL
VUID-VkFramebufferCreateInfo-height-00887(ERROR / SPEC): msgNum: 0 - vkCreateFramebuffer(): Requested VkFramebufferCreateInfo height must be greater than zero. The Vulkan spec states: height must be greater than 0. (https://www.khronos.org/registry/vulkan/specs/1.1-extensions/html/vkspec.html#VUID-VkFramebufferCreateInfo-height-00887)
    Objects: 1
        [0] 0, type: 0, name: NULL
VUID-VkViewport-width-01770(ERROR / SPEC): msgNum: 0 - vkCreateGraphicsPipelines: pCreateInfos[0].pViewportState->pViewports[0].width (=0.000000) is not greater than 0.0. The Vulkan spec states: width must be greater than 0.0 (https://www.khronos.org/registry/vulkan/specs/1.1-extensions/html/vkspec.html#VUID-VkViewport-width-01770)
    Objects: 1
        [0] 0, type: 19, name: NULL
VUID-VkViewport-height-01772(ERROR / SPEC): msgNum: 0 - vkCreateGraphicsPipelines: pCreateInfos[0].pViewportState->pViewports[0].height (=0.000000) is not greater 0.0. The Vulkan spec states: height must be greater than 0.0 (https://www.khronos.org/registry/vulkan/specs/1.0/html/vkspec.html#VUID-VkViewport-height-01772)
    Objects: 1
        [0] 0, type: 19, name: NULL
close window
Calling DestroyWindow(_window);
```

Probably more tests needed, but it looks like this line could be removed

olegded:
Will try to test `osgcompute` in the next couple of days :-) It worked.

vsg-dev:
Looking at the code block with the comment, it's for sharing instances etc. between multiple windows, so it could be that you aren't testing that code block.

The minimizing producing warnings is an interesting one, this will be an entirely different issue that we'll need to address separately.  Could you create an Issue for it and example your used in testing and how to reproduce the issue.  I'll have try now to see if it occurs under Linux.

You comment "Will try to test osgcompute in the next couple of days :-) It worked." is a bit confusing... have you tested osgcompute already under MacOS? Did it work?

olegded:
"The minimizing producing warnings is an interesting one, this will be an entirely different issue that we'll need to address separately. Could you create an Issue for it and example your used in testing and how to reproduce the issue. I'll have try now to see if it occurs under Linux."

---> Sure

olegded:
You comment "Will try to test osgcompute in the next couple of days :-) It worked." is a bit confusing... have you tested osgcompute already under MacOS? Did it work?

---> Not today, it worked last time I tested it, don't know when exactly. For sure not last week.

olegded:
Hi Robert, please see my latest findings for macOS: https://github.com/vsg-dev/VulkanSceneGraph/issues/137

It happens with your latest change https://github.com/vsg-dev/VulkanSceneGraph/commit/35fee142e594b707cad0044c5e849e690b67862c, my proposed changes in this PR and commented out line in question. 

Looks like I was simply too stupid to find right `vsgdraw` binary in the noon today or install step failed and I missed that. In any case I have started an outdated binary on my mac for testing (however, the compile step with my proposed change really succeeded).

So, there are crashes right now on macOS, for both `vsgdraw` and `vsgcompute`, further debugging needed.

PR #133, opened 2020-02-08
api version check (vk1.2) against vkGetPhysicalDeviceProperties2

mp3butcher:
vkGetPhysicalDeviceProperties2 require at least vk 1.2

### DIFF
--- a/src/vsg/vk/PhysicalDevice.cpp
+++ b/src/vsg/vk/PhysicalDevice.cpp
@@ -19,8 +19,8 @@ PhysicalDevice::PhysicalDevice(Instance* instance, VkPhysicalDevice device) :
     _instance(instance)
 {
     vkGetPhysicalDeviceProperties(_device, &_properties);
-
-    _rayTracingProperties = getProperties<VkPhysicalDeviceRayTracingPropertiesNV, VK_STRUCTURE_TYPE_PHYSICAL_DEVICE_RAY_TRACING_PROPERTIES_NV>();
+    if( _properties.apiVersion >= VK_MAKE_VERSION(1, 2, 0) )
+        _rayTracingProperties = getProperties<VkPhysicalDeviceRayTracingPropertiesNV, VK_STRUCTURE_TYPE_PHYSICAL_DEVICE_RAY_TRACING_PROPERTIES_NV>();
 
     uint32_t queueFamilyCount = 0;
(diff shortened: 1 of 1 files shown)


openscenegraph:
I don't understand the version check, vkGetPhysicalDeviceProperies2 isn't a 1.2 feature a far as I can tell.

What error are you getting?

vsg-dev:
FYI, the ray tracing functionality has been developed prior to 1.2 being released, for sure we don't need to restrict it to 1.2.

There are two elements, the availability of the RTX structures in headers, the availability of vkGetPhysicalDeviceProperies2 and the runtime support for RTX.  I need feedback on what the problem you are seeing so we can work out what needs to be changed.

mp3butcher:
perhaps a bug in mesa 19...

mp3butcher:
> I don't understand the version check, vkGetPhysicalDeviceProperies2 isn't a 1.2 feature a far as I can tell.
> 
> What error are you getting?

The entry point of the function vkGetPhysicalDeviceProperies2 seems good but call lead to crash with mesa 19

vsg-dev:
Perhaps Mesa isn't recognizing the NV data-structure and crashing rather than ignoring it.  Does sounds like a Mesa bug too me.

A workaround on the VSG end might be to check for the extension support prior to calling the getProperties().  There is an extension check elsewhere that Tom added, but I'd need to look this up to see how we might pass this into the PhysicalDevice setup.

mp3butcher:
any call to vkGetPhysicalDeviceProperties2 leads to crash so it really
seems mesa fault..

Le sam. 8 févr. 2020 à 18:25, Vulkan made easy <notifications@github.com> a
écrit :

mp3butcher:
any call to vkGetPhysicalDeviceProperties2 leads to crash so it really
seems mesa fault..
for reference mesa vulkan api version is 1.1.102

vsg-dev:
For oow just stick with your change to keep things working.

Tomorrow I'll have a look at moving the code for getting the ray tracing properties to code that directly depends upon it.

vsg-dev:
Changes now checked in, hopefully this should avoid all the issues you've been seeing:

https://github.com/vsg-dev/VulkanSceneGraph/commit/9da364a848ea2991c2e037e16af4b8d578a7b454

PR #136, opened 2020-02-08
bugfixes

mp3butcher:
This is the cleaner i can get until now..

I hacked a lot but i believe there's a bug with mesa 19 and secondaries command buffer...not sure perhaps i miss something, but it stopped me...
It should work but on my intelchipset, but  i have neither graphic result neither validation warnings :/
If someone could validate that  vsgsubpass work on your other driver...and tell me if something is displayed i would be glad

the hacked branch https://github.com/mp3butcher/VulkanSceneGraphPrototype/tree/subpass
and https://github.com/mp3butcher/vsgExamples/tree/subpass

### DIFF
--- a/include/vsg/vk/PassGraph.h
+++ b/include/vsg/vk/PassGraph.h
@@ -34,13 +34,14 @@ namespace vsg
         {
             AttachmentDescription()
             {
-                samples= VK_SAMPLE_COUNT_1_BIT;
+                flags = 0;
+                samples = VK_SAMPLE_COUNT_1_BIT;
                 loadOp  = VK_ATTACHMENT_LOAD_OP_DONT_CARE;
                 storeOp  = VK_ATTACHMENT_STORE_OP_DONT_CARE;
                 stencilLoadOp = VK_ATTACHMENT_LOAD_OP_DONT_CARE;
                 stencilStoreOp  = VK_ATTACHMENT_STORE_OP_DONT_CARE;
-                initialLayout= VK_IMAGE_LAYOUT_UNDEFINED;
-                finalLayout= VK_IMAGE_LAYOUT_UNDEFINED;
+                initialLayout = VK_IMAGE_LAYOUT_UNDEFINED;
+                finalLayout = VK_IMAGE_LAYOUT_UNDEFINED;
             }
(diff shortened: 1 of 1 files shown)


vsg-dev:
I am working on making the RenderPass/Subpass support in the VSG more flexible, using your branch as inspiration.  The branch of this experimental work is:

    https://github.com/vsg-dev/VulkanSceneGraph/tree/alternate_subpass

From your subpass branch I've pulled in the NextSubPass class and tweaked it to make it more flexible and added IO support.  I have also added a ref_ptr<RenderPass> to the Window::Traits so we can set up the RenderPass we want to use then pass it in when creating the Window.  That's the theory at least.  This approach should mean we don't need to have a callback class or implementation.

My next step is to implement the subpass example test program.  I will have a look at your vsgsubpass example for inspiration.  What end result should I expect from your example?

vsg-dev:
@mp3butcher,  Looking at this example there are two variables:

    vsg::ref_ptr< vsg::GraphicsPipeline > graphicsPipeline;
    vsg::ref_ptr< vsg::GraphicsPipeline > graphicsdepthPipeline;

But only the graphicsPipeline version is set up and used, the graphicsdepthPipeline remains unset and unused.

Is this intended?

Was this example based on a 3rd party example that was ported over?  It would be useful for me to have a look at this to see how things are mapping across.

mp3butcher:
sorry for my late answer...was offline
@vsg-dev i think you're lookiing the wrong version of the subpass example
The last version is:
https://github.com/mp3butcher/vsgExamples/blob/67ea94eba607ca05ed9a585cef220d16d77b2819/Desktop/vsgsubpass/vsgsubpass.cpp
However I just tested to launch the example with nvidia-driver-440 on a GTX660 and ..... nothing happens....(no validations errors, ,no display) :/ so it's not a mesa bug but something i'm doing wrong....will dig it..

mp3butcher:
I have succed in implement what i want subpass+secondaries buffer

it's poor design and there is lot of noisy code but it do the trick

the hacked branch https://github.com/mp3butcher/VulkanSceneGraphPrototype/tree/subpass
and https://github.com/mp3butcher/vsgExamples/tree/subpass

PR #138, opened 2020-02-11
Improved flexibility of RenderPass/WindowTraits setup

vsg-dev:


mp3butcher:
I'm quite dubious about design chosen with window in
https://github.com/vsg-dev/VulkanSceneGraph/pull/138/commits/710ba671977ff6087ad66a148cb43e4315fb506f
In my opinion, querying surface for supported presentation imageformat should be something exposed to the user during his renderpass description
That's why i used a callback instead of a simple _renderpass attribute in my branch....

vsg-dev:
On Sun, 23 Feb 2020 at 21:48, Julien Valentin <notifications@github.com>
wrote:

> I'm quite dubious about design chosen with window in
> 710ba67
> <https://github.com/vsg-dev/VulkanSceneGraph/commit/710ba671977ff6087ad66a148cb43e4315fb506f>
> In my opinion, querying surface for supported presentation imageformat
> should be something exposed to the user during his renderpass description
> That's why i used a callback instead of a simple _renderpass attribute in
> my branch....
>
My approach is the least complex, see how the functionality is works in
practice and if we really do need something more complex because of issues
that arise in real practice then that will be the time for adding extra
complexity.

mp3butcher:
> On Sun, 23 Feb 2020 at 21:48, Julien Valentin ***@***.***> wrote: I'm quite dubious about design chosen with window in [710ba67](https://github.com/vsg-dev/VulkanSceneGraph/commit/710ba671977ff6087ad66a148cb43e4315fb506f) <[710ba67](https://github.com/vsg-dev/VulkanSceneGraph/commit/710ba671977ff6087ad66a148cb43e4315fb506f)> In my opinion, querying surface for supported presentation imageformat should be something exposed to the user during his renderpass description That's why i used a callback instead of a simple _renderpass attribute in my branch....
> My approach is the least complex, see how the functionality is works in practice and if we really do need something more complex because of issues that arise in real practice then that will be the time for adding extra complexity.

Your design seems doubious:
To create a RenderPass we require Device which is created in Window::create so Not accessible before Window is created

mp3butcher:
arf sorry havent seen change in Traits

PR #139, opened 2020-03-03
Subpass + secondaries

mp3butcher:


robertosfield:
I've merged this PR to let you get on with the subpass branch, there are features I can see as being generally required so will implement some of these directly rather than pulling in from the subpass branch where I more VSG standard implementation would fit better.  For instance I've just checked in to the multi_gpu branch that will be master quite soon:

https://github.com/vsg-dev/VulkanSceneGraph/commit/ea24b8d6491de69990ba38de664cb62f4b851ccf

As I merge this work it would be worth keeping track of at least master to see how it impacts your code.  This could be quite significant in some places as I've made some pretty major changes to how vulkan objects are managed already as part of the multi_gpu work.  Through this month I'm keen to simplify the Vulkan object set up a bit.

For your wok it may be worth creating a standalone PR for specific classes if you think they are ready to go into master.

mp3butcher:
it's working and mergeable with master but just for review purpose.. but i have to think of better design before thinking of a merging perpsective:/
Further a memory leak arised i have to fix it

mp3butcher:
@robertosfield 
is the introduction of a SecondaryCommandGraph class inherited from CommandGraph seems sensible to you?

robertosfield:
On Wed, 4 Mar 2020 at 16:17, Julien Valentin <notifications@github.com>
wrote:

> @robertosfield <https://github.com/robertosfield>
> is the introduction of a SecondaryCommandGraph class inherited from
> CommandGraph seems sensible to you?
>
I did a brief look over your additions when I did the merge but didn't
attempt to QA the design as I'll need to put aside a few hours to properly
review the functionality.  Merging quickly on this branch allows the code
to be tested using your vsgExamples example so having it all checked in
makes testing easier.

I don't have time right now to do the full review and only then can I give
an definitive answers.  I am currently doing a sprint on a client project
(VSG related :-) once this is over I'll pop back over to more general VSG
support work and will then do a proper review of the subpass
functionality.  If I don't get to it this week I should next.

PR #144, opened 2020-03-12
Subpass on Master

mp3butcher:
for community reviews


vsg-dev:
Follow your suggestions I've got the changes merged and have also run build_all_h and updated the vsgExamples subpass branch to keep everything compiling.  I have also merged master VSG and vsgExamples into subpass branch respectively and this went smoothly.

When I run vsgsubpass I now get a single triangle in the foreground that is fixed in relative to the window and behind the vsgdraw scene rotating smoothly.  Is this what I should expect?

mp3butcher:
> 
> 
> Follow your suggestions I've got the changes merged and have also run build_all_h and updated the vsgExamples subpass branch to keep everything compiling. I have also merged master VSG and vsgExamples into subpass branch respectively and this went smoothly.
> 
> When I run vsgsubpass I now get a single triangle in the foreground that is fixed in relative to the window and behind the vsgdraw scene rotating smoothly. Is this what I should expect?

Yes the sample renderpass is pretty simple 
1st pass is normal 2nd omit the depth attachement
it yield in a non zfighting triangles

mp3butcher:
however i must draw your attention on an issue i had merging with multigpu changes:
You're storing multiple windows in the command graph, 
this seems to forbid windows to have different renderpass if i'm right ... 
To fix conflict i considered windows[0]->renderpass to be the one
but it doesn't fix the design issue

vsg-dev:
I'm away from my dev machine so can't look up and review the details, but off the top of my head you'd just need to use multiple CommandGraph if you have multiple Windows that have different RenderPass.  This constraint is currently just implicit.  It may be necessary to add some extra checks to provide useful debug feedback or evolve the design to handle a wider range of usage cases.

mp3butcher:
> I'm away from my dev machine so can't look up and review the details, but off the top of my head you'd just need to use multiple CommandGraph if you have multiple Windows that have different RenderPass. This constraint is currently just implicit. It may be necessary to add some extra checks to provide useful debug feedback or evolve the design to handle a wider range of usage cases.

It's just i don't understand why a commandgraph is associated with multiple windows:
https://github.com/vsg-dev/VulkanSceneGraph/blob/master/src/vsg/viewer/CommandGraph.cpp#L33

vsg-dev:
I don't recall the exact reasoning, I know at one point I was struggling to get some of the new classes to work nicely together as at different points you need to have access to different bits of the data to coordinate things, having multiple Windows may have been the easiest way to get things to work.

When I move on to implementing multi-threading I am expecting to have to further evolve all these viewer related classes to make it work in a coherent way, so what you see now is likely to be refined or perhaps even replaced completely.  How to do this type of stuff with Vulkan isn;'t an established science, we just have to try different things, learn what works, what doesn't, change things when they don't work well.  Eventually as the usage widens we'll settle on a scheme that works across a wide range of usage cases.

PR #149, opened 2020-03-21
Command Buffers Record Threading and Sync

mp3butcher:
Perhaps we should wait threading before merging this

### DIFF
--- a/src/vsg/viewer/Viewer.cpp
+++ b/src/vsg/viewer/Viewer.cpp
@@ -211,7 +211,8 @@ void Viewer::handleEvents()
 class CollectSecondaryCommandGraph : public ConstVisitor
 {
 public:
-    vsg::CommandGraphs _secondaries;
+    vsg::CommandGraphs secondaries;
+    std::vector<std::shared_ptr<std::mutex> > execCommandMuters;
     void apply(const Group& group) override
     {
         group.traverse(*this);
@@ -220,9 +221,10 @@ public:
         const vsg::ExecuteCommands *exec = dynamic_cast<const vsg::ExecuteCommands*>(&cmd);
         if(exec)
         {
-            for( auto g :exec->_cmdgraphs)
+            for( vsg::ExecuteCommands::Secondaries::const_iterator gm = exec->_cmdgraphs.begin(); gm != exec->_cmdgraphs.end(); ++gm)
             {
-                _secondaries.emplace_back(g);
+                secondaries.emplace_back(*gm);
+                execCommandMuters.emplace_back(std::move(exec->getCommandGraphMutex(*gm)));
             }
         }
     }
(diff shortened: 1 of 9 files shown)


src/vsg/viewer/Viewer.cpp
```diff
@@ -212,6 +212,7 @@ class CollectSecondaryCommandGraph : public ConstVisitor
 {
 public:
     vsg::CommandGraphs _secondaries;
+    std::vector<std::shared_ptr<std::mutex> > _execcommandmuters;
```
vsg-dev: A quick general comment, names need to be human readable ratherthanallbuncheduptogethermakingithardertoreadonaglance.  The standard way to make it readable isToUseCapitalization or_underscore_separated.

mp3butcher: indeed...fixed

robertosfield:
Thanks for the capitalisation changes, really helps readability,

I have just done a another online pass over the changes but still haven't a clear enough idea of where you are going with it to know if it's the right approach.  Have you attempted a multi-threading run with these changes?

I haven't done any work on the design side of mulit-threaded viewer since I introduced the RecrodAndSubmitTask/CommandGraph/RenderGraph classes so can't yet provide guidance on how I'll go about this, and therefore how it might dovetail into the changes you've made in this PR.  

The RecordAndSbumitTiask approach is a departure from how the OSG does things and is focused entirely on the needs for preparing and submitting data to Vulkan, this differences means we can't just adopt the threading model of the OSG's and transplant it the VSG.  We have plenty more opportunities on the VSG/Vulkan side for threading too, and will be able to record multiple command buffers in a parallel.  This will likely mean a thread per CommangGraph would be appropriate.  FYI, the vsg::DatabasePager does some mulit-threaded work already so I already have a little bit of experience of making it work.

For recording of secondary command buffers potentially this could be done via nesting, or perhaps as dependencies.  This obviously touches on the subpass functionality, so perhaps before merging any subpass work I need to figure out the threading model I want to go for and make sure the subpass merges cleanly with this.

To do do the subpass review/merge I think it would easiest for me to just merge the Vulkan functionality required and keep the thread safety functionality in this PR separate.  I can create a branch for this PR so enable this separation.

Workwise I'll be busy with a client's VSG project work this coming week, I have a milestone I need to achieve before I can jump back into core VSG work.  

I will need a clear run of time just focused on the multi-threading side to make sure I come up with a coherent design and implementation.  Sorry it's taking me so long to get the subpass work merged, it's come in during a period that I've been juggling lots of other VSG related work, my brain just isn't quite multi-thread safe enough for me to think deeply enough about disparate topics!

mp3butcher:
it's just a primitive NProducers(secondary)-1Consumer(primary) pattern implementation. I haven't tested it threaded but on the paper it should work.
I'll study a bit more C++-11 std::threading tools in order to make better design.

robertosfield:
I added a bit of extra threading functionality for the vsg::read threading and vsg::DatabasePager, have a look at include/vsg/threading.  These add a little extra functionality on top of std::thread.

I can create a subpass-threading branch if you want to continue to experiment with combining subpass and threading.

mp3butcher:
> I can create a subpass-threading branch if you want to continue to experiment with combining subpass and threading.

It would be a good idea, I've began Implementing threading for CommandGraph but something is wrong...Can't figure why yet..

mp3butcher:
Ready for an other branch

vsg-dev:
I have created a subpass_threading off subpass:

    https://github.com/vsg-dev/VulkanSceneGraph/tree/subpass_threading

PR #151, opened 2020-03-23
Command Buffers Record Threading and Sync

mp3butcher:
following https://github.com/vsg-dev/VulkanSceneGraph/pull/149
tested and ready for an other branch subpass-threading...(perhaps issue with DataPager sharing...)

### DIFF
--- a/src/vsg/viewer/Viewer.cpp
+++ b/src/vsg/viewer/Viewer.cpp
@@ -289,12 +289,12 @@ void Viewer::compile(BufferPreferences bufferPreferences)
             auto& deviceResource = deviceResourceMap[commandGraph->_device];
             commandGraph->_maxSlot = deviceResource.collectStats.maxSlot;
 
-            if(commandGraph->_primary.valid())
+            if(commandGraph->_masterCommandGraph.valid() &&  commandGraph->_masterCommandGraph->_commandBuffersLevel == VK_COMMAND_BUFFER_LEVEL_PRIMARY)
             {
-                deviceResource.compile->context.renderPass = static_cast<RenderGraph*>(commandGraph->_primary->getChild(0))->window->renderPass();
-                deviceResource.compile->context.viewport = static_cast<RenderGraph*>(commandGraph->_primary->getChild(0))->camera->getViewportState();
-                deviceResource.compile->context.viewport->getViewport().width = static_cast<RenderGraph*>(commandGraph->_primary->getChild(0))->window->extent2D().width;
-                deviceResource.compile->context.viewport->getViewport().height = static_cast<RenderGraph*>(commandGraph->_primary->getChild(0))->window->extent2D().height;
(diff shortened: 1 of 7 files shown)


src/vsg/threading/OperationThreads.cpp
```diff
@@ -30,10 +30,26 @@ OperationThreads::OperationThreads(uint32_t numThreads, ref_ptr<Active> in_activ
             }
         }
     };
+    auto run = [](ref_ptr<OperationQueue> q, ref_ptr<Active> a) {
```
openscenegraph: Why have two run functions? The new one will essentially be spin lock and just sit their consuming 
CPU resources regardless of whether there is any load or not.  

The implementation of the take_when_available is pretty efficient and shouldn't result in threads waiting long times before waking up to do work.

If the general solution is more efficient on CPU resources and work fast enough then it's something we should strongly prefer over spin locks.  If users want to do something undesirable then I would generally try to steer them in the direction of doing something that is generally better.

If doing the right then for normal usage and keeping the code simple is something I'd use as the default, adding complexity for very niche usage cases is something we should avoid, especially if it can be misused by developers who might not realized what they are doing can be lead to significant inefficiencies.

src/vsg/viewer/Viewer.cpp
```diff
@@ -211,18 +212,21 @@ void Viewer::handleEvents()
 class CollectSecondaryCommandGraph : public ConstVisitor
 {
 public:
-    vsg::CommandGraphs _secondaries;
+    vsg::CommandGraphs secondaries;
+    std::vector<std::shared_ptr<std::mutex> > execCommandMuters;
```
openscenegraph: Are you sure you want to use the name Muters?

The dictionary definition of Muter is:

"adjective, mut·er, mut·est.
silent; refraining from speech or utterance.
not emitting or having sound of any kind."

If you are looking for plural of mutex then mutices or mutexes, or move the plural elsewhere listOfExceCommandMutex.

mp3butcher: as a mutex mutes the thread, it's not absurd to call it a muter...However if you want me to change this, i don't mind

mp3butcher: done

openscenegraph: Halting a thread isn't muting it.  Muting would be something like redirect console output to /dev/null, or muting the sound of video while the video continues playing.

If you have a list of mutex then just call it that, or call it by a proper plural.

In terms of design it might be that sharing a std::mutex is not the ideal thing anyway so this discussion might a "mute" point :-)

However, as I haven't done a long enough review of your changes and thought about the wider topic of viewer threading I can't provide guidance yet on whether the direction you are going is ideal.

I have other work to complete this week before I can dedicate any time to the topic of getting subpass integrated and to look at viewer threading.

src/vsg/threading/OperationThreads.cpp
mp3butcher: it was mainly for test purpose but I admit I haven't seen any difference

PR #152, opened 2020-03-26
Allow multiple master command graph consumers

mp3butcher:
allow multiple master command graph consumers


mp3butcher:
i think i'd work not tested :/

PR #153, opened 2020-03-27
Allow multiple consumers command graphs per producer

mp3butcher:
Not sure but should work on the paper :/

### DIFF
--- a/include/vsg/viewer/CommandGraph.h
+++ b/include/vsg/viewer/CommandGraph.h
@@ -3,6 +3,7 @@
 /* <editor-fold desc="MIT License">
 
 Copyright(c) 2018 Robert Osfield
+Copyright(c) 2020 Julien Valentin
 
 Permission is hereby granted, free of charge, to any person obtaining a copy of this software and associated documentation files (the "Software"), to deal in the Software without restriction, including without limitation the rights to use, copy, modify, merge, publish, distribute, sublicense, and/or sell copies of the Software, and to permit persons to whom the Software is furnished to do so, subject to the following conditions:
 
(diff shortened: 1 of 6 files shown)


vsg-dev:
Have you not tested it?  This really needs to be  pre-requisit generating generating a PR.

mp3butcher:
tested ok

mp3butcher:
Can anyone could think of a use case where a commandbuffer are consumed by multiple primary?....If not this pr is useless

robertosfield:
What usage case did you have in mind when you wrote it?

I haven't had a chance to review your changes properly yet, not had time to think deeply on the topic of viewer level threading.  I'll throw out a few quick thoughts in case this is useful, they may be resolved in your PR already.

Recordeding a subgraph to a secondary command buffer and then using this within multiple primary command buffers, or across multiple frames is the reason why secondary command buffers exist.  One obviously needs to have fully recorded a secondary command buffer before using it in a primary command buffer.  

One could manage this at a high level at CommandGraph level or lower level at the ExecuteCommands level.  The later would allow greater parallelism in the case of it being possible to record multiple CommandGraphs at once.  In this case one would need to  avoid conditions where a primary command buffer is being recorded, but is waiting on results from a secondary command buffer, but this comment buffer has yet to execute due to both the primary and secondary being schedule on the same thread.

mp3butcher:
ok so this pr still have some interest....
Then use case i want to adress was primary that call dynamic secondary (transform) that call static secondary(recorded once and for all)...However that's not a way it could work (only one level)

There is still a design problem in this pr i have to manage before merging:
something to know which primary commandbuffer is secondary commandbuffer consumer..will reopen pr one it's done

PR #154, opened 2020-04-01
Allow multiple consumers command graphs per secondary

mp3butcher:
works but **mandates to duplicate ExecuteCommands in primaries**
I made an simple 2 primary sharing 1secondary example
https://github.com/vsg-dev/vsgExamples/pull/42

### DIFF
--- a/src/vsg/viewer/Viewer.cpp
+++ b/src/vsg/viewer/Viewer.cpp
@@ -289,12 +289,12 @@ void Viewer::compile(BufferPreferences bufferPreferences)
             auto& deviceResource = deviceResourceMap[commandGraph->_device];
             commandGraph->_maxSlot = deviceResource.collectStats.maxSlot;
 
-            if(commandGraph->_masterCommandGraph.valid() &&  commandGraph->_masterCommandGraph->_commandBuffersLevel == VK_COMMAND_BUFFER_LEVEL_PRIMARY)
+            if(commandGraph->_primary.valid())
             {
-                deviceResource.compile->context.renderPass = static_cast<RenderGraph*>(commandGraph->_masterCommandGraph->getChild(0))->window->renderPass();
-                deviceResource.compile->context.viewport = static_cast<RenderGraph*>(commandGraph->_masterCommandGraph->getChild(0))->camera->getViewportState();
-                deviceResource.compile->context.viewport->getViewport().width = static_cast<RenderGraph*>(commandGraph->_masterCommandGraph->getChild(0))->window->extent2D().width;
-                deviceResource.compile->context.viewport->getViewport().height = static_cast<RenderGraph*>(commandGraph->_masterCommandGraph->getChild(0))->window->extent2D().height;
+                deviceResource.compile->context.renderPass = static_cast<RenderGraph*>(commandGraph->_primary->getChild(0))->window->renderPass();
+                deviceResource.compile->context.viewport = static_cast<RenderGraph*>(commandGraph->_primary->getChild(0))->camera->getViewportState();
+                deviceResource.compile->context.viewport->getViewport().width = static_cast<RenderGraph*>(commandGraph->_primary->getChild(0))->window->extent2D().width;
+                deviceResource.compile->context.viewport->getViewport().height = static_cast<RenderGraph*>(commandGraph->_primary->getChild(0))->window->extent2D().height;
             }
 
             commandGraph->accept(*deviceResource.compile);
@@ -332,20 +332,6 @@ void Viewer::compile(BufferPreferences bufferPreferences)
             task->databasePager->start();
         }
     }
-}
-
-void recursivCollectSlaveGraphs(CommandGraphs& collectedgraphs, ref_ptr<CommandGraph> & mastergraph)
-{
-    CollectSecondaryCommandGraph collector;
-    mastergraph->accept(collector);
-    auto mutexit = collector.execCommandMutices.begin();
-    for( auto slavegraph : collector.secondaries )
-    {
-        slavegraph->_masterCommandGraph = mastergraph;
-        slavegraph->_masterCommandBufferMutex = (*mutexit++);
-        recursivCollectSlaveGraphs(collectedgraphs, slavegraph);
-    }
-    collectedgraphs.insert(std::end(collectedgraphs), std::begin(collector.secondaries), std::end(collector.secondaries));
 }
 
 void Viewer::assignRecordAndSubmitTaskAndPresentation(CommandGraphs in_commandGraphs, DatabasePager* databasePager)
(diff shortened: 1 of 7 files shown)


mp3butcher:
yhty

src/vsg/viewer/Viewer.cpp
```diff
@@ -282,21 +291,12 @@ void Viewer::compile(BufferPreferences bufferPreferences)
     {
         std::set<Device*> devices;
 
-        for (auto& commandGraph : task->commandGraphs)
+        for (auto& commandGraph : reverse(task->commandGraphs))
```
robertosfield, line 294: Why is the reverse required here?

mp3butcher: to create pipeline in primary before used in secondaries

mp3butcher: I would have push primaries front but the current chosen design in submitTask mandates secondaries to be treated first (to track their primaries)

mp3butcher: i just fixed this poor design

robertosfield:
FYI, I've been pondering a little on the issue of CommandGraph/CommandBuffer/ExecututeCommand and and threading, not enough to have a coherent opinion yet, but pondering still.  

Once approach that came to me after merging and testing your changes was whether we can simplify things by having the CommandGraph act as means of recording primary CommandBuffer for submission to a queue and for executing via recording secondary CommandBuffer for the purposes of being used in VkExecuteCommand.  

If a CommandGraph is placed within the subgraph of top level CommandGraph then it will implicitly be a secondary CommandGraph that is executed via VkExecuteCommand so that it's result is available to the enclosing CommandGraph.  If nested/secondary CommandGraph is shared between multiple top level CommandGraph then the nested/secondary CommandGraph's recorded secondary VkCommandBuffer will simply be reused.

If a hardware/driver system (some phones) doesn't support secondary VkCommandBuffer then the nested system would simply traverse the nested/secondary CommandGraph subgraph each time it's needed rather than have record once and execute many times that is supported by secondary VkCommandBuffer.

The CommandGraph would need to have a dirty system for knowing when a subgraph needs to be recorded.  The CommandGraph would also have a local mutex to prevent it's contents from being used and recorded at the same time.  You may also want to generalize the ability of single CommandGraph to manage multiple recorded VkCommandBuffer so that it can handle different input conditions (i.e. view matrix etc.) as well as double/triple buffering that is required due to viewer/windowing double/triple buffering.

If this approach is viable then it'd avoid the need for sharing of mutexes across the nodes in the scene graph and avoid the need to explicitly have ExceuteCommands.  I'll let this topic gestate a bit longer before making any decisions.  If I do try out the above idea then it'd be via a new CommandGraph branch from master.

mp3butcher:
> FYI, I've been pondering a little on the issue of CommandGraph/CommandBuffer/ExecututeCommand and and threading, not enough to have a coherent opinion yet, but pondering still.
> 
> Once approach that came to me after merging and testing your changes was whether we can simplify things by having the CommandGraph act as means of recording primary CommandBuffer for submission to a queue and for executing via recording secondary CommandBuffer for the purposes of being used in VkExecuteCommand.
> 

I'm thinking the same: I  was poking around differents usecase of vkExecuteCommand (ie: integrate them in the infrastructure would prescribe flexibility of the current design), but conclude that calling them from primaries might be the only one, so this simplification might be good.

> If a CommandGraph is placed within the subgraph of top level CommandGraph then it will implicitly be a secondary CommandGraph that is executed via VkExecuteCommand so that it's result is available to the enclosing CommandGraph. If nested/secondary CommandGraph is shared between multiple top level CommandGraph then the nested/secondary CommandGraph's recorded secondary VkCommandBuffer will simply be reused.
> 
> If a hardware/driver system (some phones) doesn't support secondary VkCommandBuffer then the nested system would simply traverse the nested/secondary CommandGraph subgraph each time it's needed rather than have record once and execute many times that is supported by secondary VkCommandBuffer.

Do some smart phone doesn't support secondaries?
It would be strange ... if some manufacturers didn't implement them they must comply with vulkan spec and hide it under the wood driver side..no?

> 
> The CommandGraph would need to have a dirty system for knowing when a subgraph needs to be recorded. The CommandGraph would also have a local mutex to prevent it's contents from being used and recorded at the same time. You may also want to generalize the ability of single CommandGraph to manage multiple recorded VkCommandBuffer so that it can handle different input conditions (i.e. view matrix etc.) as well as double/triple buffering that is required due to viewer/windowing double/triple buffering.
> 
> If this approach is viable then it'd avoid the need for sharing of mutexes across the nodes in the scene graph and avoid the need to explicitly have ExceuteCommands. I'll let this topic gestate a bit longer before making any decisions. If I do try out the above idea then it'd be via a new CommandGraph branch from master.

arg I've began the simplification on my branch :/

mp3butcher:
I Think it would be wiser to treat this simplication (if praticable) in conjonction with subpass paradigm:
I forgot that a primary with secondaries may use NextSubPass command, it forbid the integration of ExecuteCommand in CommandGraph infrastructure (haven't found a trick to overcome this yet)

mp3butcher:
I Think it would be wiser to treat this simplification (if praticable) in conjonction with subpass paradigm:
I forgot that a primary with secondaries may use NextSubPass command, it forbid the integration of ExecuteCommand in CommandGraph infrastructure.
I haven't found a satisfying trick to overcome this:
@robertosfield 
one solution would be to integrate NextSubPass in CommandGraph infrastructure but it would prevent user to submit subpass inlined (without secondaries)
Would this be acceptable to you?

mp3butcher:
I have a working branch here (more stuff should be removed but the idea is here)
https://github.com/mp3butcher/VulkanSceneGraphPrototype/tree/subpass_threading_simp

robertosfield:
Thanks for the refactor. I've done a first pass review.  Just looking at
diffs doesn't really do the whole feature justice so I'll need to return to
it and study it properly.

I'm currently working through client work, my plan is to tie up this work
and then jump back over to pure VSG work next week.  Once of the items on
the VSG side that I'll be focusing on it how to handle
subpass/nextsubpass/CommandGraph/RenderPass and threading.  By the end of
next week I'd really like for us to have resolved the best path forwards.
Really don't know how aspirational/duluded this is.  Will find out :-)

mp3butcher:
> Thanks for the refactor. I've done a first pass review. Just looking at diffs doesn't really do the whole feature justice so I'll need to return to it and study it properly. 

?! All "feature justice" is done in this branch!...I follow most of you guidance... I just omit the dirty flag on CommandGraph...
Concerning shared_ptr mutex there just shortcut and can't be removed

Further You don't reply my question about design acceptability of NextSubPass integration
I will pr on the alternate_subpass branch after few cleaning and you'll see that next week when you have time....

robertosfield:
On Thu, 9 Apr 2020 at 10:12, Julien Valentin <notifications@github.com>
wrote:

> Thanks for the refactor. I've done a first pass review. Just looking at
> diffs doesn't really do the whole feature justice so I'll need to return to
> it and study it properly.
>
> ?! All "feature justice" is done in this branch!...I follow most of you
> guidance... I just omit the dirty flag on CommandGraph...
> Concerning shared_ptr mutex there just shortcut and can't be removed
>
> Further You don't reply my question about design acceptability of
> NextSubPass integration
> I will pr on the alternate_subpass branch after few cleaning and you'll
> see that next week when you have time....
>
I can't provide replies on stuff I haven't researched problem.  You've been
exploring areas of Vulkan/VSG that I haven't yet had time to really think
about.  There are plenty bits of your work that I haven't immediately
understood and won't until I've spent the time looking at fully.  I just
need to clear time to look at the issues you are tackling and your solution
to them.

Once I've had time to look the topic and your approaches then I'll be able
to provide guidance.  It may well be that I see a path through which is
different to the approaches you've taken so far.  It may be significant
difference, or just small differences, or no difference at all.

What I can't do though is engage in a deep discussion about it as this
point in time.  I'm clearing the deck for me to look into it next week.

PR #155, opened 2020-04-05
Add CI using github actions

gupascal:
# Add CI using github actions

## Description

Hello,
This pull request adds automatized builds on push and pull request events, using github actions facilities. Builds are run on Linux, Windows and Mac OS and exit in failure in case of any error. (ATM, build warnings do nothing and I hope to improve that someday but probably not in a v1.)

For build triggered by push on master, documentation is built and published as build artifact.

Vulkan SDK version to use can be set using an environment variable set at the beginning of the workflow yml file.

Currently, the build fails on Mac OS because of a missing namespace in code. I kept this unchanged for now as it is good illustration of this CI workflow.

I hope this help. Any feedback is welcome.

## How Has This Been Tested?

- With cache missing (first run) or with cache hit
- After changing Vulkan SDK version
- Building or not documentation, depending on trigger

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -1,0 +1,97 @@
+name: CI
+on:
+  push:
+  pull_request:
+    branches:
+      - master
+env:
+  VulkanSDKVersion: 1.2.131.2
+jobs:
+  linux-build:
+    runs-on: ubuntu-latest
+    env:
+      BuildDocEnabled: ${{github.event_name == 'push' && github.ref == 'refs/heads/master'}}
+      VULKAN_SDK: $GITHUB_WORKSPACE/../$VulkanSDKVersion/x86_64
+    steps:
+    - uses: actions/checkout@v2
+    - name: Cache
+      id: cache
+      uses: actions/cache@v1.1.2
+      with:
+        path: ${{env.VULKAN_SDK}}
+        key: VulkanSdk${{env.VulkanSDKVersion}}ExtractedLinux
+    - name: Download & Extract Vulkan SDK
+      if: steps.cache.outputs.cache-hit != 'true'
+      run: |
+        wget --no-cookies -O ../vulkansdk-linux-x86_64-${{env.VulkanSDKVersion}}.tar.gz https://sdk.lunarg.com/sdk/download/${{env.VulkanSDKVersion}}/linux/vulkansdk-linux-x86_64-${{env.VulkanSDKVersion}}.tar.gz?u=
+        tar -zxf ../vulkansdk-linux-x86_64-${{env.VulkanSDKVersion}}.tar.gz -C ../
+    - name: Install Doxygen
+      if: env.BuildDocEnabled == 'true'
+      run: sudo apt install doxygen graphviz
+    - name: CMake
+      run: |
+        cmake -D Vulkan_INCLUDE_DIR="${{env.VULKAN_SDK}}/include" \
+          -D Vulkan_LIBRARY="${{env.VULKAN_SDK}}/lib/libvulkan.so" .
+    - run: make -j 2 
+    - if: env.BuildDocEnabled == 'true'
+      run: make docs
+    - name: Publish doc
+      if: env.BuildDocEnabled == 'true'
+      uses: actions/upload-artifact@v1
+      with:
+        name: VSG Documentation
+        path: html
+
+  windows-build:
+    runs-on: windows-latest
+    steps:
+    - uses: actions/checkout@v2
+    - name: Set env
+      run: echo "::set-env name=VULKAN_SDK::C:\VulkanSDK\${{env.VulkanSDKVersion}}"
+    - name: Cache
+      id: cache
+      uses: actions/cache@v1.1.2
+      with:
+        path: ${{env.VULKAN_SDK}}
+        key: VulkanSdk${{env.VulkanSDKVersion}}WindowsExtracted2
+    - name: Download & Install Vulkan SDK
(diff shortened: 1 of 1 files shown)


vsg-dev:
Awesome addition, thanks for you efforts ;-)

vsg-dev:
I haven't used Actions before so it's good to see it deployed.  I see that on merging this PR a build under WIndows, Linux and MacOS was kicked off with the later failing as you expected.  I've looked at the error and  to me it looks to be an issue on the handling of the .yml script in some way.  The error is:

```

Run wget --no-cookies -O ../vulkansdk-macos-1.2.131.2.tar.gz https://sdk.lunarg.com/sdk/download/1.2.131.2/mac/vulkansdk-macos-1.2.131.2.tar.gz?u=
--2020-04-06 06:28:47--  https://sdk.lunarg.com/sdk/download/1.2.131.2/mac/vulkansdk-macos-1.2.131.2.tar.gz?u=
Resolving sdk.lunarg.com (sdk.lunarg.com)... 209.126.117.218
Connecting to sdk.lunarg.com (sdk.lunarg.com)|209.126.117.218|:443... connected.
HTTP request sent, awaiting response... 412 Precondition Failed
2020-04-06 06:28:48 ERROR 412: Precondition Failed.

##[error]Process completed with exit code 8.

```
Running the wget line on my Linux box tells me that the file itself downloads fine.

Any ideas?

gupascal:
Well, this "412: Precondition Failed." error is quite difficult to deal with (and was not the one I expected and refered to). It is a strange security or cookies error. The first time I got it, I though I had solved it with the --no-cookies but it seems it was just randomness... I just tried various new things with no success.
The only solution I can think about now, would be to store yourself the macos archive somewhere and set the right url (cache in workflow will avoid relying on it to much). Maybe a proper way to do this would be to create an empty vsg-dev/vulkan-sdk repo, create a github release 1.2.131.2, upload the macos archive as asset and use this asset url. What do you think of it?

BTW, I think builds should have triggered on this pull. I will remove the 'master' branch filter for build on pull requests because it doesn't seems to work as expected with cross repo pull request (and it should not be a bad idea to build all PRs).

vsg-dev:
The failure on MacOS looks to be consistent - I have only seen failures so far.  Any chance that the MacOS system has some facilities blocked on the wget being a bit buggy/different than Linux?

I could create a vulkan-sdk repo but this might confuse the developers coming to the VSG project.  Potentially I could create a vulkan-sdk on my personal github to host this, this way we can experiment without it being an official pathway.

I am open to suggests as at this point I really don't have knowledge of the problem - I much more of coder than an server admin.

robertosfield:
I had a bash at uploading the vulkan-macos to a dedicate github account but it falls foul of github's file size limit of 100.00MB as the file is 146.MB.

Github does has large filesize support but I haven't used it before, I don't know how difficult it would be to set up.

gupascal:
Well the randomness of the failure is a bit strange. Since I started to work on the workflow, the build always succeeded for a period, then always fail for another period and finally always succeeded again on my experimental branch (without any change related to the download part). I continue to think about how to fix it, but I do not have any idea for now.

Regarding the re-upload of the archive, I was thinking about the github Release functionality which allow big assets. Not git commits which have indeed size limitations.
I tried uploading it on my side here: https://github.com/gupascal/vulkan-sdk/releases Then the workflow succeeded to download it using the new URL (https://github.com/gupascal/vulkan-sdk/releases/download/v1.2.131.2/vulkansdk-macos-1.2.131.2.tar.gz)
I do not know if you want me to update the PR with this. It could be better if you have your own upload on a repo you own (eg. for future versions). What do you prefer?

robertosfield:
On Mon, 6 Apr 2020 at 20:36, gupascal <notifications@github.com> wrote:

> Well the randomness of the failure is a bit strange. Since I started to
> work on the workflow, the build always succeeded for a period, then always
> fail for another period and finally always succeeded again on my
> experimental branch (without any change related to the download part). I
> continue to think about how to fix it, but I do not have any idea for now.
>
> Regarding the re-upload of the archive, I was thinking about the github
> Release functionality which allow big assets. Not git commits which have
> indeed size limitations.
> I tried uploading it on my side here:
> https://github.com/gupascal/vulkan-sdk/releases Then the workflow
> succeeded to download it using the new URL (
> https://github.com/gupascal/vulkan-sdk/releases/download/v1.2.131.2/vulkansdk-macos-1.2.131.2.tar.gz
> )
> I do not know if you want me to update the PR with this. It could be
> better if you have your own upload on a repo you own (eg. for future
> versions). What do you prefer?
>

Could you generate a PR with it working against your vulkan-sdk so we have
a successful build right away.  Then we can reflect on how to proceed
longer term.

robertosfield:
Looks like we've entered the twilight zone where the macOS automated build has progressed past the Vulkan_SDK download and finally got the build error relating to the missing vsg:: namespace.  

I've had a bash at fixing the issue, but without a macOS box I can't directly test so fingers crossed it'll just magically build once more....

robertosfield:
The macOS build has now completed successfully so I'm hoping that now the Vulkan-SDK has downloaded once it'll be cached and we won't see further problems with the download of the SDK.  If this does remain the case then I don't think there is a need to provide our own copies of the SDK.  So for now I don't think you'll need to generate a PR to get the automated build to point anywhere other than the LunarG SDK.

Thanks for your efforts implementing this feature, much appreciated :-)

gupascal:
Great to see it working :-)
I am not sure that cache is shared between push triggered builds and PR.
We will see...

Issue #160, opened 2020-04-10
Threading issue with OperationThreadQueue::Latch


mp3butcher:
**Describe the bug**
MT Freeze (Latch stays locked after count reach 0)

**To Reproduce**
[video](https://streamable.com/gsqyld)(vgssubpass exhibits same issue)
The bug emerged when i used Latch in vsg-dev:subpass_threading branch...
Hard to reproduce, but launching lot of instances of the **vsgexecutecommandbuffers** example yield application freeze in some windows 
(launch 6 instances and swap between windows seems to be a good way to provoke the bug)

**Desktop**
 - OS: Ubuntu 18.04.1

**Additional context**
The issue might affect DataPager but harder to reproduce
I fix it locally with https://github.com/vsg-dev/VulkanSceneGraph/commit/b4429c414c0fa10bc81e92e89dcbdec53dabda14 using a mutex instead of a condition_variable

robertosfield:
The way you test multi-threading is my running multiple threads within a single application, not by running lots of separate programs.  See a problem behavior when you are thrashing your window manager is a stress test of completely different kind.  It could be that you are hitting a window manager issue, or a driver issue, or CPU bug or a memory bug.

The way to create a stress test for the VSG threads is to write a single application with multiple threads that contend over the same resources.

I'm closing this "Issue" as it's a dubious test, on an experimental branch that has no relevance to master.  Any of the work that may make in from the experimental branch will be thoroughly vetted.  If then we find problems we need to establish it with a single application. multi-threaded test.

mp3butcher:
If it was a driver bug or a window manager issue my fix wouldn't work. 
But it does... You're bad faith

robertosfield:
Threading problems are often a very difficult to pin down and fix.  Some
"fixes" can just avoid a race condition but aren't fixes at all.  In this
instance you "fixes" are sledge hammer hacks that change the conditions
without actual pinpointing and fixing the actual bug if an actual bug
actually exists.

What you are reading as bad faith is years of experience of dealing with
threading problems and how it's most effective to deal with them.  Me
saying you aren't helping progress things forward on this is not some
random statement.  You are doing stuff that it's clear you don't yet have
the experience and insight to tackle.

PR #162, opened 2020-04-10
repair the Latch 

mp3butcher:
Fixes # (issue)
https://github.com/vsg-dev/VulkanSceneGraph/issues/160
## Type of change

Muteces instead of atomic int and condition

### DIFF
--- a/include/vsg/threading/OperationQueue.h
+++ b/include/vsg/threading/OperationQueue.h
@@ -37,21 +37,41 @@ namespace vsg
     class Latch : public Inherit<Object, Latch>
     {
     public:
-        Latch(size_t num) :
-            _count(num) {}
+        Latch(int num) :
+            _count(num)
+        {
+            _mutex.lock();
+        }
+
+        virtual void reset(const int value)
+        {
+            std::scoped_lock lock(_incMutex);
+            _count = value;
+            if(_mutex.try_lock()) _mutex.lock();
+        }
 
         void count_up()
         {
-            ++_count;
+             std::scoped_lock lock(_incMutex);
+             _count++;
+        }
+
+        int atomicCompIncAndReturn(const int inc = 1, const int sig = 1)
+        {
+            std::scoped_lock lock(_incMutex);
+            int result = _count, temp = result + inc;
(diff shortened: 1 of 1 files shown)


robertosfield:
The Latch implementation was based inspired by examples of latches implemented elsewhere.  Resorting to using mutexes is not good for performance.  I'd much rather find out the cause of the problem and address it than resort to what look like heavy handed fixes.

The first step has to be create a small discrete example that illustrates the bug then hammer this with a tool like valgrind.  The provided Issue #160 is related to an experimental branch which itself may contain problems that are leading to what you think is the latch failing, or it could simply be that the driver is failing in the suggest usage.

mp3butcher:
> The Latch implementation was based inspired by examples of latches implemented elsewhere. Resorting to using mutexes is not good for performance. I'd much rather find out the cause of the problem and address it than resort to what look like heavy handed fixes.
> 

Can't escape theses write access...

> The first step has to be create a small discrete example that illustrates the bug then hammer this with a tool like valgrind. The provided Issue #160 is related to an experimental branch which itself may contain problems that are leading to what you think is the latch failing, or it could simply be that the driver is failing in the suggest usage.

It's strange that 2 of my "experiments" branches exhibit both the same issue without having the same synchronisation design...

robertosfield:
Sorry but the changes just aren't good enough.  We can't just go throwing a sledge hammer around in the core VSG on a whim because some experimental branch did something odd when a used in a non standard way.

Next week I'll write a stress test for OperationQiueue and it's \Latch and see what I come up with.

FYI, I have a 16 core machine, multi GPU system to work with, paged databases.  I have nearly two decades experience of working on mulit-threaded OSG.  You just need to wait till I get on to this stuff.

As a general comment, for performance multi-threading in the VSG/Vulkan is far less important than it was with the OSG as it's some much faster.  If you have project you want to get on with then for now just do them single threaded, if hit a bottle neck then post a question/explanation to vsg-users.

If you just doing the multi-threaded stuff for fun then please play on you own repo for the time being, as I'm afraid having to deal with this half backed submissions is just taking me away from other work.

PR #163, opened 2020-04-10
repair the Barrier taking account read accesses

mp3butcher:
following https://github.com/vsg-dev/VulkanSceneGraph/pull/162

## Fixes
https://github.com/vsg-dev/VulkanSceneGraph/issues/160

## Type of change

Mutex instead of condition

### DIFF
--- a/include/vsg/threading/OperationQueue.h
+++ b/include/vsg/threading/OperationQueue.h
@@ -38,7 +38,15 @@ namespace vsg
     {
     public:
         Latch(size_t num) :
-            _count(num) {}
+            _count(num)
+        {
+            _mutex.lock();
+        }
+
+        virtual void reset(const int value)
+        {
+            _count = value;
+        }
 
         void count_up()
         {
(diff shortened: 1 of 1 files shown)


mp3butcher:
invoking the abstract machine, this solution is not good enough but seems not to exhibit same issue

mp3butcher:
oups too drunk

mp3butcher:
not perfect in regards to the Abstract Machine but working

robertosfield:
And efficient latch is designed around a condition variable, just throwing this away is not a solution.

mp3butcher:
your problem is that condition variable is waited after it has been notified...It's obvious in the code you have copy paste from internet

mp3butcher:
the only way current design could be used safely is by launching a waiting thread before the workers

robertosfield:
If there is a usage case that is problem then create a simple test case
that illustrates the problem.  I will then look at it.  If you don't do it
I'll write it myself when I have a clear run at it.

mp3butcher:
test case:
```
mainthread init latch to 1
mainthread lauching 1 worker with latch->count_down-)
mainthread latch.wait
```
for i in `seq 0 2000000`; do  ./a.out 
In some cases worker will have finished before wait and latch wait forever

PR #169, opened 2020-05-01
a small fix to CMAKE files to allow proper include paths when VSG is an external lib

neshume:
# Pull Request Template

## Description

Please include a summary of the change and which issue is fixed. Please also include relevant motivation and context. List any dependencies that are required for this change.

Fixes # (issue)

## Type of change

- [ ] build system related Bug fix (non-breaking change which fixes an issue)

## How Has This Been Tested?
- VulkanSceneGraph was added as a sub-module and used externally.


robertosfield:
Do you have an example that illustrates the problem this change resolves for you?  I'd like to recreate the problem and see the cause.  Thanks.

neshume:
I have shared a private repo with you that points to my fork .. I can create a public repo that points to the original VSG and that will illustrate the issue. Will that be OK Robert?

PR #180, opened 2020-05-18
Complete multisample support in context and compile traversal

timoore:
Change context::graphicsPipelineStates to two vectors,
defaultPipelineStates and overridePipelineStates, which are applied
before and after the states of a GraphicPipeline's states. Multisample
state is determined by the FrameAssembly object and is put in the
override list so that the rasterizationSamples value will be valid for
the current render pass.

This patch has a problem: it overrides all the MultisampleState in a GraphicsPipeline e.g., alpha to coverage. I've realized that a GraphicsPipelineState object doesn't need to set a complete Vulkan pipeline state; there could be an override object that only sets the rasterizationSamples and leaves the rest alone.

# Pull Request Template

## Description

Better multisampling handling.

## Type of change

## How Has This Been Tested?

Run a bunch of examples

**Test Configuration**:
* Firmware version:
* Hardware:
* Toolchain:
* SDK:

## Checklist:


src/vsg/state/GraphicsPipeline.cpp
```diff
@@ -398,15 +398,20 @@ void RasterizationState::apply(VkGraphicsPipelineCreateInfo& pipelineInfo) const
 //
 // MultisampleState
 //
-MultisampleState::MultisampleState() :
+MultisampleState::MultisampleState(VkSampleCountFlagBits rasterizationSamples_) :
```
robertosfield, line 401: The trailing VAR_ looks like typo in the context of the rest of the VSG that doesn't use that convention for function parameter,  I'd rather be explicit with an preceding in_VAR or just a simple renamed like samples.

Also use of two constructors here is a bit overkill when a default value of VkSampleCountFlagBits would have sufficed.  

I will merge this PR and then fix these issues.  I mention here as note for future changes.

src/vsg/state/GraphicsPipeline.cpp
```diff
@@ -107,9 +104,12 @@ GraphicsPipeline::Implementation::Implementation(Context& context, Device* devic
     _renderPass(renderPass),
     _pipelineLayout(pipelineLayout),
     _shaderStages(shaderStages),
-    _pipelineStates(pipelineStates),
+    _pipelineStates(context.defaultPipelineStates),
     _allocator(allocator)
 {
+    _pipelineStates.insert(_pipelineStates.end(), pipelineStates.begin(), pipelineStates.end());
```
robertosfield, line 110: Moving the pipeline states into the GraphicsPipeline::Implemention is messier than the parent code doing the combining.  I don't know why you just didn't add to the original code.  I will merge and then revert to the original code that does the combining in a more linear step by step way rather than mixing things up in the constructor.

timoore:
Not a typo, but a personal convention when a class / struct field doesn't
have an underscore or other member designator on it. Do whatever you like
there.

Tim

timoore:
My intention is/was to skip the combining and just iterate over the
different pipeline state lists; I don't see that _pipelineStates is used
anywhere else. For this initial work I was trying to keep my changes to a
minimum, just to make sure that we were on the right track.

Tim

robertosfield:
On Tue, 19 May 2020 at 18:25, timoore <notifications@github.com> wrote:

> My intention is/was to skip the combining and just iterate over the
> different pipeline state lists; I don't see that _pipelineStates is used
> anywhere else. For this initial work I was trying to keep my changes to a
> minimum, just to make sure that we were on the right track.
>

At a later date I'll do a review of what parts need to be cached in the
various *::Implementation classes.  The intent is to cache local objects to
prevent the associated Vulkan objects being deleted that the
*::Implementation require to exist for them to function correctly.

As for keeping changes to a minimum, it changed the way it was working and
would have required just two line of code to be changed to handle your
changes to Context *pipelineStates.

As a general comment, where code is rarely called I prioritize readability
over efficiency.

W,r,t use of naming conventions, while it isn't entirely solidified yet
across the whole VSG, introducing new conventions in an adhoc, personal
preference way, is a step towards inchorency in the code base.  This is why
I flag it up as trying to fix inconsistencies after they've been introduced
is a lot more effort and likely to miss problems than not introducing them
in the first place.

For naming right now, public members and public methods all have no prefix
or suffix.  Protected members have a proceeding _underscore.  Parameters
that share a number with a member variable I have been using in_name to
avoid collision.  I have never felt the postfix _ is helpful, it's just a
workaround to a problem that almost never exists in realworld code, and a
workaround that a natural human language meaning, the case of preceeding _
is essentially a shorthand for m_name which is shorthand for
membervariable_name, having name_ be short hand for in_name, or m_name I
find counter intuitive.

Once I formalize the naming conventions I'll write all this up.  So far
I've been testing out a few different approaches to see how that look in
practice in a live applications.  As I go I find certainly things don't
seem to fit and I change them.   For instance in an early experiment with
file extensions I tried out both .h and .hpp and the later just felt
cumbersome and counter-intuitive, i.e. .h is a header file, while .hpp is a
header plus plus?  .cpp is a c plus plus file so that makes intuitive
sense.  Some odd conventions have appeared around C++ over the years, some
that are widely adopted are just plain odd if you take a step back.

PR #193, opened 2020-06-09
add mouse wheel event

wadivision:
# Pull Request Template

## Description

I  found that the trackball manipulator  didn't respode to the mouse wheel events, so I added this functionality. 
I only modified the win32 part because  I worked on Windows  operating system, and was not familiar with other operating systems. 

## Type of change

## How Has This Been Tested?

- [ ] Test A
- [ ] Test B

**Test Configuration**:
* Firmware version:
* Hardware:
* Toolchain:
* SDK:

## Checklist:

### DIFF
--- a/include/vsg/ui/PointerEvent.h
+++ b/include/vsg/ui/PointerEvent.h
@@ -73,4 +73,15 @@ namespace vsg
             Inherit(in_window, in_time, in_x, in_y, in_buttonMask) {}
     };
 
+    VSG_type_name(vsg::WheelEvent);
+    class WheelEvent : public Inherit<PointerEvent, WheelEvent>
+    {
+    public:
+        WheelEvent(Window* in_window, time_point in_time, int32_t in_x, int32_t in_y, ButtonMask in_buttonMask, short in_delta) :
+            Inherit(in_window, in_time, in_x, in_y, in_buttonMask),
+            delta(in_delta) {}
+        
+        short delta;// multiple of WHEEL_DELTA (120)
+    };
+
 } // namespace vsg
--- a/src/vsg/platform/win32/Win32_Window.cpp
+++ b/src/vsg/platform/win32/Win32_Window.cpp
@@ -568,8 +568,8 @@ LRESULT Win32_Window::handleWin32Messages(UINT msg, WPARAM wParam, LPARAM lParam
         uint32_t mx = GET_X_LPARAM(lParam);
         uint32_t my = GET_Y_LPARAM(lParam);
 
-        uint32_t button = msg == WM_LBUTTONDOWN ? 1 : (msg == WM_RBUTTONDOWN ? 2 : msg == WM_MBUTTONDOWN ? 3 : (msg == WM_XBUTTONDOWN ? 4 : 0)); // need to determine x1, x2
-        _bufferedEvents.emplace_back(new vsg::ButtonReleaseEvent(this, event_time, mx, my, getButtonMask(wParam), getButtonEventDetail(msg)));
+        uint32_t button = msg == WM_LBUTTONUP ? 1 : (msg == WM_RBUTTONUP ? 2 : msg == WM_MBUTTONUP ? 3 : (msg == WM_XBUTTONUP ? 4 : 0)); // need to determine x1, x2
+        _bufferedEvents.emplace_back(new vsg::ButtonReleaseEvent(this, event_time, mx, my, getButtonMask(wParam), button));
 
         //::ReleaseCapture(); // should only release once all mouse buttons are released ??
         break;
@@ -582,7 +582,14 @@ LRESULT Win32_Window::handleWin32Messages(UINT msg, WPARAM wParam, LPARAM lParam
     }
     break;
     case WM_MOUSEWHEEL:
+    {
+        short fwKey = LOWORD(wParam);
+        short zDelta = HIWORD(wParam);
+        short mx = GET_X_LPARAM(lParam);
+        short my = GET_Y_LPARAM(lParam);
+        _bufferedEvents.emplace_back(new vsg::WheelEvent(this, event_time, mx, my, getButtonMask(wParam), zDelta));
         break;
+    }
     case WM_MOVE:
     case WM_SIZE:
     {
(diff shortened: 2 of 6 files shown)


include/vsg/ui/PointerEvent.h
```diff
@@ -73,4 +73,15 @@ namespace vsg
...
 
+    VSG_type_name(vsg::WheelEvent);
+    class WheelEvent : public Inherit<PointerEvent, WheelEvent>
+    {
+    public:
+        WheelEvent(Window* in_window, time_point in_time, int32_t in_x, int32_t in_y, ButtonMask in_buttonMask, short in_delta) :
+            Inherit(in_window, in_time, in_x, in_y, in_buttonMask),
+            delta(in_delta) {}
+        
+        short delta;// multiple of WHEEL_DELTA (120)
```
robertosfield, line 84: Where does the 120 come from?

wadivision: Hi，Robert. There is some information  about WM_MOUSEWHEEL in MSDN.

[MSDN](https://docs.microsoft.com/en-us/windows/win32/inputdev/wm-mousewheel)


> #define WM_MOUSEWHEEL                   0x020A
> 
> Parameters
> wParam
> 
> The high-order word indicates the distance the wheel is rotated, expressed in multiples or divisions of WHEEL_DELTA, which is 120. A positive value indicates that the wheel was rotated forward, away from the user; a negative value indicates that the wheel was rotated backward, toward the user.

robertosfield: That's quite bizarre interval. I think it would be best to just scale the delta to an interval of one prior to passing as an event, or for us to pass the interval with the even as well.  The aim should be to present a coherent set of values across platforms.  

Once I've got osg2vsg building against VulkanSDK 1.2.141.0 I'll come back to this issue and see what the other platforms do.  I'll also check what the OSG to see how it tries to avoid platform specific scaling.

src/vsg/platform/win32/Win32_Window.cpp
```diff
@@ -568,8 +568,8 @@ LRESULT Win32_Window::handleWin32Messages(UINT msg, WPARAM wParam, LPARAM lParam
         uint32_t mx = GET_X_LPARAM(lParam);
         uint32_t my = GET_Y_LPARAM(lParam);
 
-        uint32_t button = msg == WM_LBUTTONDOWN ? 1 : (msg == WM_RBUTTONDOWN ? 2 : msg == WM_MBUTTONDOWN ? 3 : (msg == WM_XBUTTONDOWN ? 4 : 0)); // need to determine x1, x2
-        _bufferedEvents.emplace_back(new vsg::ButtonReleaseEvent(this, event_time, mx, my, getButtonMask(wParam), getButtonEventDetail(msg)));
+        uint32_t button = msg == WM_LBUTTONUP ? 1 : (msg == WM_RBUTTONUP ? 2 : msg == WM_MBUTTONUP ? 3 : (msg == WM_XBUTTONUP ? 4 : 0)); // need to determine x1, x2
+        _bufferedEvents.emplace_back(new vsg::ButtonReleaseEvent(this, event_time, mx, my, getButtonMask(wParam), button));
```
wadivision, line 572: There may be a typo.Please check it.
@robertosfield

include/vsg/ui/PointerEvent.h
wadivision: Yeah,you are right.

src/vsg/platform/win32/Win32_Window.cpp
robertosfield: A typo in your revision or @tomhog's original work on Win32_Window.cpp?

wadivision: A typo in vsg master branch.I think the 'msg' should be 'WM_XXXUP' rather than 'WM_XXXDOWN' for button release event.I fixed it and tested in my revision.

wadivision:
@robertosfield 
Hi，Robert!

I checked the implementation of mouse Wheel event  in OSG, 

```c++
   case WM_MOUSEWHEEL :
        getEventQueue()->mouseScroll(GET_WHEEL_DELTA_WPARAM(wParam)<0 ? osgGA::GUIEventAdapter::SCROLL_DOWN :
                                                                            osgGA::GUIEventAdapter::SCROLL_UP,
                                         eventTime);
            break;

 enum ScrollingMotion
        {
            SCROLL_NONE,
            SCROLL_LEFT,
            SCROLL_RIGHT,
            SCROLL_UP,
            SCROLL_DOWN,
            SCROLL_2D
        };
```

It looks like a better way to focus on the  type of event rather than delta value.
I optimized my code for this  function. haha

robertosfield:
I have started looking into the topic on mouse wheel support.  Curiously I happened to add handling of the X11 approach to handling mouse wheel into Trackball.  X11 maps the mouse motion to button 4 and 5 and sends a button press and release - this is not done by the VSG, it's actually X11/Xcb that is sending these events and the VSG's Xcb_Window.cpp just passes it along oblivious that it's a hacky implementing of mouse wheel under the hood.

Adopting the X11 hacky for the VSG and amending Windows and macOS to work this way would be daft so I think it would be better for us to come up with mods to Xcb_Window.cpp to catch the hack and re-direct it to a more relevant mouse wheel event.

There are some mice with multiple scroll wheels.  Then there is possibility of different directions as well like the OSG's ScrollingMotion eludes to.

I need to continue looking into this topic as there doesn't yet seem a clear route through.

src/vsg/platform/win32/Win32_Window.cpp
robertosfield: Just had a look at the changes code.  The button value isn't used anywhere, so while the code looks wrong, it's also not used.  I'll just delete the line and check this into master.

robertosfield: So I've remove the offending line and the automated build just went through under Windows so no problems.  My guess is that Tom originally put the line in as a copy and paste from other button handling and as the implementation evolved removed the original place as part of the clean up process but missed this one.

https://github.com/vsg-dev/VulkanSceneGraph/commit/c86b88f2e3b2ddb8794b65c554460b005fb528a9

robertosfield:
@wadivision I have now merged the ScrollWheel branch with master.  If Windows isn't working then let me know via the vsg-users googlegroup or via the Issue tracker

PR #235, opened 2020-12-19
added a cmake option to disable msvc iterator checking in debug mode

sbrkopac:
# Pull Request Template

## Description

Added a CMake option to disable iterator checking in debug mode. This open is currently flagged as `PUBLIC` so it gets inherited by projects that import VulkanSceneGraph. This is important because if the target applications settings don't have this flag - users will get a crash.

## Type of change

## How Has This Been Tested?

I created a very simple CMake project to ensure the flag was getting imported:

```
cmake_minimum_required(VERSION 3.7)

project(vsg-test)

find_package(vsg REQUIRED)

add_executable (vsg-test main.cpp)
target_link_libraries (vsg-test vsg::vsg)
```

I then validated it was in the `Preprocess Definitions` in Visual Studio: ```WIN32;_WINDOWS;_ITERATOR_DEBUG_LEVEL=0;CMAKE_INTDIR="Debug";%(PreprocessorDefinitions)```

**Test Configuration**:
* Hardware: MSVC 2019

## Checklist:

### DIFF
--- a/src/vsg/CMakeLists.txt
+++ b/src/vsg/CMakeLists.txt
@@ -253,7 +253,7 @@ if(MSVC)
         target_compile_options(vsg PRIVATE "/MP")
     endif()
 
-    option(DISABLE_CHECKED_ITERATORS "Turning on this option will disable checked iterators in debug mode for visual studio. All imported projects must adhere to this requirement")
+    option(DISABLE_CHECKED_ITERATORS "Turning on this option will disable checked iterators in debug mode for visual studio. All imported projects must adhere to this requirement" OFF)
 
     if(DISABLE_CHECKED_ITERATORS)
         target_compile_definitions(vsg PUBLIC "_ITERATOR_DEBUG_LEVEL=0")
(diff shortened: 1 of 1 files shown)


src/vsg/CMakeLists.txt
```diff
@@ -246,13 +246,19 @@ if(MSVC)
...
-	
+
     option(ENABLE_MP_FLAG "Turning on this option will add the multi-processor flag in MSVC for VSG and it's included projects" ON)
-	
+
     if(ENABLE_MP_FLAG)
         target_compile_options(vsg PRIVATE "/MP")
     endif()
 
+    option(DISABLE_CHECKED_ITERATORS "Turning on this option will disable checked iterators in debug mode for visual studio. All imported projects must adhere to this requirement")
```
robertosfield, line 256: There isn't a default option supplied, what will cmake use in this case?

sbrkopac: OFF is the always the default

https://cmake.org/cmake/help/v3.0/command/option.html

robertosfield: Thanks for the clarification.  Could you add OFF to the option line, I think it's better to explicitly set defaults so that users reading the cmake scripts can read default behaviour without knowing cmake in deoth,  Thanks,

sbrkopac: Agreed. The PR should be updated with OFF.

PR #241, opened 2021-01-19
Added stream support to vsg::read and vsg::read_cast

anormann1974:


robertosfield:
The reason why I didn't previously implement the istream versions of the vsg::read() methods is that which ReaderWriter to use is normally determined by the file name extension, without this all ReaderWriter's would need to assume that they can handle it, but that then the wrong ReaderWriter could consume the istream prematurely before it's had a chance to figure out the data isn't of the required type.

Merging this submissions would leave this whole issue to resolve but without having discussed how we might go about it or whether it's easily resolvable at all.

I'm curious, what was the motivation behind the changes?  Were you aware of the above issue and had something in mind to resolve it?

anormann1974:
You know I am working on an assimp importer. The glTF-Format can have embedded images. That is the reason why I needed the stream support. 

One solution could be to extend vsg::Options with a "extensionHint" or "formatHint".

anormann1974:
vsg::ObjectCache also depends on a filename. This is also a problem. Maybe a hash value would solve the problem. This could be used for filenames and raw data.

robertosfield:
With the OSG the approach is to get the ReaderWriter to invoke from the osgDB::Registry, then application code calls the ReaderWriter directly.  My assumption has to use this approach with the VSG, though I haven't done any work in that direction yet.

You suggestions about Options hints is something I have considered too, I am open to this. As a general guide I want to keep the code as simple as it can be and still provide the required functionality.   The usage case you describe for gl-TF is very much part of required functionality so I'm on board with finding some solution.

What I can do is create one or more branches of VSG and vsgXchange for us to experiment with so we can get something working without worrying too much about the code being in a final form.  I'll have a think about an appropriate branch name, suggestions welcome :-)

robertosfield:
I've create a [ReadStream](https://github.com/vsg-dev/VulkanSceneGraph/tree/ReadStream) for the core VSG.

And an [Assimp](https://github.com/vsg-dev/vsgXchange/tree/Assimp) branch for vsgXchange.

robertosfield:
@anormann1974 I created a PR from VSG repo to merge the changes into ReadStream branch:

   https://github.com/vsg-dev/VulkanSceneGraph/tree/ReadStream

I'll close the PR.  Any further work on it can go into the ReadStream branch.

anormann1974:
Ok, I will checkout the new branch.

PR #247, opened 2021-01-29
Needed changes for cubemap support

anormann1974:
Hi Robert,

in this pull request you will find the needed changes in vsg to support cubemaps. I have created a new class ArrayCube to store the image data. Currently only uncompressed formats are supported. A pull request for compressed data will follow later.

I also merged vsg master branch into the ReadStream branch to be up to date.

### DIFF
--- a/include/vsg/all.h
+++ b/include/vsg/all.h
@@ -17,6 +17,7 @@ THE SOFTWARE IS PROVIDED "AS IS", WITHOUT WARRANTY OF ANY KIND, EXPRESS OR IMPLI
 #include <vsg/core/Array.h>
 #include <vsg/core/Array2D.h>
 #include <vsg/core/Array3D.h>
+#include <vsg/core/ArrayCube.h>
 #include <vsg/core/Auxiliary.h>
 #include <vsg/core/ConstVisitor.h>
 #include <vsg/core/Data.h>
--- a/include/vsg/core/ArrayCube.h
+++ b/include/vsg/core/ArrayCube.h
@@ -1,0 +1,325 @@
+#pragma once
+
+/* <editor-fold desc="MIT License">
+
+Copyright(c) 2018 Robert Osfield
+
+Permission is hereby granted, free of charge, to any person obtaining a copy of this software and associated documentation files (the "Software"), to deal in the Software without restriction, including without limitation the rights to use, copy, modify, merge, publish, distribute, sublicense, and/or sell copies of the Software, and to permit persons to whom the Software is furnished to do so, subject to the following conditions:
+
+The above copyright notice and this permission notice shall be included in all copies or substantial portions of the Software.
+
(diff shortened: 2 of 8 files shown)


robertosfield:
Thanks Norman.  Just done a first pass review.  So that I could see the key differences I've merged master into the vsg-dev/VulkanSceneGraph ReadStream.  I now just see the 8 files changed.

Originally I had assumed we could reuse Array3D as the basis for doing cubemaps.  I see that there are differences in how data is passed to Vulkan that require different treatment with setting up of Layers rather than depth.  I'm now wondering if a Texture2DArray equivalent will need the same treatment.

I will need to do another pass over looking in detail at the changes before setting upon a path.  Currently I'm inclined towards merging this branch with with vsg-dev/VulkanScenerGraph/ReadStream and doing a code review in situ.  I'm also still curious about the path reuse Array3D for doing cubemaps and texture 2d arrays so may create a parallel branch and see how it all just hangs together, then chose which ever implementation feels the more natural to use.

anormann1974:
I also tried to reuse Array3D, but I got some troubles. That was the reason why I create the ArrayCube class. One other solution might be to use Array2D and add an array member for handling faces.

...
 uint32_t m_width; 
uint32_t m_height; 
uint32_t m_faces;
...

Then the ArrayCube class will bee obsolete.

robertosfield:
Hi Andre,

On Fri, 29 Jan 2021 at 09:26, Andre Normann <notifications@github.com>
wrote:

> I also tried to reuse Array3D, but I got some troubles.
>
What problems did you come across?


> That was the reason why I create the ArrayCube class. One other solution
> might be to use Array2D and add an array member for handling faces.
>

I find the name ArrayCube a bit awkward.  Perhaps it'd be OK just to use
Cubemap.

> ...
> uint32_t m_width;
> uint32_t m_height;
> uint32_t m_faces;
>
That's a novel approach.  It would mean users of Array2D would need to
check for faces to make sure they process it correctly.

I've merged and will review those changes.  I will do a eview.  I am also
working on a new vsgGIS lib today so will have to juggle my time between
these tasks.

anormann1974:
> On Fri, 29 Jan 2021 at 09:26, Andre Normann ***@***.***> wrote: I also tried to reuse Array3D, but I got some troubles.
> What problems did you come across?

Computing mipmap size was one problem. I stored the num of arrays (for cubemaps always 6) inside the depth component of vsg::Array3D. But this will be taken into mipmap calculation. In vsg::Image the imageType is set to VK_IMAGE_TYPE_3D, but for cubemap it must be VK_IMAGE_TYPE_2D. This can be handled by the use application, but it would be nice to have it set automatically. 

-André

anormann1974:
> That's a novel approach. It would mean users of Array2D would need to check for faces to make sure they process it correctly. I've merged and will review those changes. I will do a eview. I am also working on a new vsgGIS lib today so will have to juggle my time between these tasks.

vsgGIS is very interesting. I work for an engineering office and have developed an OSG based program with GIS/CAD functionality. It is used for very large terrain visualization, which can embed GIS and CAD data as well. So any information about vsgGIS is welcome :-). Maybe best via email.

-André

PR #249, opened 2021-01-29
Cleanup and basic compression support

anormann1974:
Hi Robert,

I have cleaned up the CopyAndReleaseImage::CopyData::record method and added basic compression support.


robertosfield:
Something looks to have broken on texture upload with these changes.  See #250 for screenshot of issue when rendering a paged database that contains compressed mipmapped images.  I can't look into it further right now.  When I can I'll create a small example that reproduces the issue.

anormann1974:
Can you share this dataset?

robertosfield:
On Sat, 30 Jan 2021 at 14:14, Andre Normann <notifications@github.com>
wrote:

> Can you share this dataset?
>
I can't share this particular one.  I'll build a new database that doesn't
include proprietary data.

robertosfield:
I've posted a paged dataset, and created a standalone test.vsgt file that illustrates the regression when handling.  To the #250 Issue report I've added all the links and further info about the Vulkan debug and API output of VSG master which works without problems and ReadStream that is failing.

PR #265, opened 2021-03-30
build fix for android

rollastre:
there is also a need to remove the call to pthread_setaffinity_np(thread_native_handle, sizeof(cpu_set_t), &cpuset) in src/vsg/threading/Affinity.cpp (line 126). That function does not exist in the NDK

Android build is broken in 3 places. This fixes 2 of them. The other one remains open as per Robert's request. He'll fix it.

I'm writing an application using VSG and I tested these changes to work fine in it. Without these changes, VSG doesn't build at all

### DIFF
--- a/src/vsg/platform/android/Android_Window.cpp
+++ b/src/vsg/platform/android/Android_Window.cpp
@@ -320,8 +320,7 @@ Android_Window::Android_Window(vsg::ref_ptr<WindowTraits> traits) :
 {
     _keyboard = new KeyboardMap;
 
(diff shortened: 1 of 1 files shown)


robertosfield:
Thanks @rollastre.  I have implement non op versions of the vsg::setAffinity(..) functions that should hopefully compile OK.  Could you test and let me know if it works fine. #5af97a72dac81eb24acbbf81df5edd57555d73ff

rollastre:
Thanks @robertosfield, I ran my build and I confirm it now works without patching the code. So this is ok. Hopefully, we find a solution for setting the thread affinity. Performance is very welcome. Especially in mobile.

PR #271, opened 2021-04-24
Corrections for Multiview

karolmateusznowak:
## Description

A few fixes that make the Multiview feature to work as expected in context of single vsg::RenderGraph.

**NOTE**: Unfortunately the change in "RenderGraph.cpp" screws up the multiview feature in case of **many** vsg::RenderGraph(s)


robertosfield:
Could you provide suggestions on how to spot the problems that these two changes address.  I'd like to reproduce the problems so I can see the source of the errors and how the changes fix them.  

Tomorrow I'll have a bash at fixing the RenderGraph cases as well.

PR #279, opened 2021-05-08
RenderGraph: force to update viewports, scissors, renderArea and clears

karolmateusznowak:
Intention of these changes is to allow an user to force to update viewports, scissors, renderArea and clears. It is needed when user changes size of view otherwise new size is not applied.

Pseudo Code (event handler):

      auto width = _window->extent2D().width;
      auto height = _window->extent2D().height;
     
      _camera1->getViewportState()->set(0, 0, width, height);
      
      // nothing changed, view internally has new size but it's presented with old one
      
      // to solve this, force rendergraph to update views
      _renderGraph->forceUpdate = true;

### DIFF
--- a/include/vsg/viewer/RenderGraph.h
+++ b/include/vsg/viewer/RenderGraph.h
@@ -51,13 +51,16 @@ namespace vsg
         /// Subpass contents stetting passed to vkCmdBeginRenderPass
         VkSubpassContents contents = VK_SUBPASS_CONTENTS_INLINE;
 
-        /// Callback used to automatically update viewports, scissor, renderAraa and clears when the window is resized.
+        /// Callback used to automatically update viewports, scissors, renderArea and clears when the window is resized.
         /// By default is null so no resize handling is done.
         ref_ptr<WindowResizeHandler> windowResizeHandler;
 
         /// window extent at previous frame, used to track window resizes
         const uint32_t invalid_dimension = std::numeric_limits<uint32_t>::max();
         mutable VkExtent2D previous_extent = VkExtent2D{invalid_dimension, invalid_dimension};
+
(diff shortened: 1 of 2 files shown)


robertosfield:
I've done a first pass review and am torn about how best to get things to work better for your usage case.   The introduction of forceUpdate flag is a plausible solution, though it does require knowledge from end users to know what forceUpdate relates to, when to use it, what state of the RenderGraph/subgraph is updated.

So I'm wondering if we need evolve RenderGraph and related classes to better handle resizes and other changes, ideally have things automatically update when settings are changed.  Dirty counts might be one way of handling this, or caching the previously applied values and then checking for changes relative to this.

The approach I've taken so far with the VSG has been to implement the simple mechanisms to get the ball rolling, then refactoring the simple approaches that cause problems in usage. 

To figure out how best to evolve RenderGraph etc. to handle your usage case I think the best thing would be to flesh out the usage case in a dedicated vsgExample and use this as a testbed/unit test.  Would could move this PR to a branch so that it can be used in conjunction with this example.

I am in the middle of some changes to vsgExamples/vsgXchange/VSG related to the positional state work, I'd like to get that work to a point where I it's at a good point to set aside, jump over the the multiview functionality that you've been working on, get that improvements to the VSG to allow you to move on your your own work, then for me to jump back to the positional state work.

It's weekend so I'm mostly doing DIY around the house rather than coding so realistically it'll be Monday before I can wrap up the PositionalState work sufficient to jump over the multiview work.

robertosfield:
I have merged using the branch suggested by github:

    karolmateusznowak-RenderGraph-ForceUpdate

I will experiment to see what may be the best long term solution.

PR #280, opened 2021-05-12
KeySymbol: missing keys added

karolmateusznowak:
Missing key symbols added

### DIFF
--- a/include/vsg/ui/KeyEvent.h
+++ b/include/vsg/ui/KeyEvent.h
@@ -94,6 +94,7 @@ namespace vsg
         KEY_Quotedbl = 0x22,
         KEY_Hash = 0x23,
         KEY_Dollar = 0x24,
+        KEY_Percent = 0x25,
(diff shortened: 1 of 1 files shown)


robertosfield:
Thanks for the PR.  I have done quick review, bit will need to follow up tomorrow.

What were using as a guide for the naming used?

My inclination is that have a look through  X11/keysymdef.h installed on my system for inspiration.  For instance it has:

#define XK_braceleft                     0x007b  /* U+007B LEFT CURLY BRACKET */
#define XK_bar                           0x007c  /* U+007C VERTICAL LINE */
#define XK_braceright                    0x007d  /* U+007D RIGHT CURLY BRACKET */
#define XK_asciitilde                    0x007e  /* U+007E TILDE */

The original list that the VSG has KeySymbol was inspired by what the OSG has as well as X11.  It might be that we need to update both the OSG and VSG :-)

PR #310, opened 2021-09-03
Add macro vsg_add_cmake_support_files()

rhabacker:
## Description
This pr adds a new cmake macro for the central generation of cmake support files. This patch was inspired by https://github.com/vsg-dev/vsgXchange/issues/45.

## Type of change
- [x] Partial revision of the cmake build system to reduce maintenance effort

## How Has This Been Tested?
- [x] Compiled and installed VulkanSceneGraph with the patches from this pr applied
- [x]  Compiled and installed vsgXchange with a patch applied to use this new macro

**Test Configuration**:
* OS: OpenSUSE Linux openSUSE 15.2

## Checklist:

### DIFF
--- a/src/vsg/CMakeLists.txt
+++ b/src/vsg/CMakeLists.txt
@@ -343,18 +343,6 @@ if (glslang_FOUND)
     set(FIND_DEPENDENCY_GLSLANG "find_dependency(vsg_glslang)")
 endif()
 
-configure_file("${CMAKE_SOURCE_DIR}/src/vsg/vsgConfig.cmake.in" "${CMAKE_BINARY_DIR}/src/vsg/vsgConfig.cmake" @ONLY)
-
-# [==[
-install(EXPORT vsgTargets
-    FILE vsgTargets.cmake
-    NAMESPACE vsg::
-    DESTINATION ${CMAKE_INSTALL_LIBDIR}/cmake/vsg
+vsg_add_cmake_support_files(
+    CONFIG_TEMPLATE vsgConfig.cmake.in
 )
-
-include(CMakePackageConfigHelpers)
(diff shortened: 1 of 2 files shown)


robertosfield:
Reminds me of the C++/C programming recommendation of breaking big functions down into small functions that do specific tasks :-)

How much more work do you think there could be on the vsgMacros work?  Potentially I could marge with master in the next few days if you feel things are ready.

rhabacker:
After merging the currently open pr's a few things may be added/changed:
1. in the macro add_option_maintainer() the PREFIX argument can be made optional and removed from all repos except VulkanSceneGraph, since all repos now have a top-level call to project() and the prefix can be extracted from the project name (similar to what is implemented in vsg_add_cmake_support_files()).
2. currently, basic cmake targets like `clobber`, `uninstall` and `add_option_maintainer()`[1] need to be added to all top-level CMakeLists.txt. I am thinking about adding calls for them in vsgMacros.cmake and removing them from all other places so they are available by default in all repos. This way, new basic macros can be made available by default in the future.

[1] add_feature_summary() probably can't be added this way, since it has to be done after the find_package() calls for all features to be present, but I haven't checked that yet.

robertosfield:
I will do a review of the respective vsgMacros.cmake branches this afternoon with a view to merging them with the respective masters.  If there are further development there after this will be fine - they can happen on the vsgMacros.cmake branches and be later merged.  I can hold off if you want.

rhabacker:
I think that the commits in the vsgMacros.cmake branches should work, as I have not encountered any problems building and installing binaries from these branches on openSUSE Leap 15.2 and 15.3 https://build.opensuse.org/project/show/home:rhabacker:branches:games:vsgMacros.cmake, so please start the review - thanks for your support.

robertosfield:
I have merged the various vsgMacros.cmake branches with the respective master now.  I haven't deleted any of these branches in case further work needs to be done.

So far everything looks cleaner and works for me.  I have a few warnings to fix though...

PR #327, opened 2021-09-10
Fix for using template classes with MinGW compiler

rhabacker:
## Description
Fixes https://github.com/vsg-dev/vsgGIS/issues/13

## Type of change

## How Has This Been Tested?
On a local system using a mingw cross compiler and by a github CI pipeline triggered by this PR.  

## Checklist:
- [x] My changes generate no new warnings (I hope so)

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -94,4 +94,9 @@ install(FILES "${CMAKE_CURRENT_SOURCE_DIR}/cmake/vsgMacros.cmake" DESTINATION ${
 #
 add_subdirectory(src/vsg)
 
+#
+# tests directory contains testing applications for the vsg library
+#
+add_subdirectory(tests)
+
 vsg_add_feature_summary()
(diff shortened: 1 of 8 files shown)


robertosfield:
I feel that pushing all these suggested changes as one isn't helpful for figuring out what parts to merge when.  Whether to have unit tests in the source code base is also a matter for discussion.  

The windows-dll build is also failing due to the new tests, which seem only to exist to test this specific feature, which I understand is required because we don't yet have automated build testing of vsgXchange etc.  But this compile fail is on a PR which includes changes to include an export to the Value/Array* classes, which I believe is what you added to prevent these DLL issues.

robertosfield:
I am closing this PR as the addition of the exports to Value/Array* breaks the Windows DLL build, confirmed by my various experiments at fixing the compile with the test-linking-class-templates.cpp, the fix was to removed the VSG_DECLSPEC from Value/Array* and to add an include <vsg/core/ConstVisitor.h> and include <vsg/vore/Visitor.h>

PR #328, opened 2021-09-11
Add class template link test case

rhabacker:
## Description
Add test case for linking class templates on various platforms, see https://github.com/vsg-dev/vsgGIS/issues/13

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -94,4 +94,9 @@ install(FILES "${CMAKE_CURRENT_SOURCE_DIR}/cmake/vsgMacros.cmake" DESTINATION ${
 #
 add_subdirectory(src/vsg)
 
+#
+# tests directory contains testing applications for the vsg library
+#
+add_subdirectory(tests)
+
 vsg_add_feature_summary()
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -7,6 +7,7 @@ project(vsg
 )
 set(VSG_SOVERSION 2)
 SET(VSG_RELEASE_CANDIDATE 0)
+set(CMAKE_CXX_STANDARD 17)
 
 # create the version header
 set(VSG_MAX_DEVICES 1 CACHE STRING "Set maximum number of vk/vsg::Device that are supported.")
--- a/tests/CMakeLists.txt
+++ b/tests/CMakeLists.txt
@@ -1,0 +1,1 @@
+add_subdirectory(core)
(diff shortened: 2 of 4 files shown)


robertosfield:
As an experiment I've checked in this PR as a branch called rhabacker-AddClassTemplateLinkTestCase.  I've also just pushed the addition of the exports to see what the effect is on the test build under Windows.

robertosfield:
@rhabacker I've just merged a vsgExamples PR from @sbrkopac that adds automated build of both the VSG and vsgExamples together, and while there are 48 warnings (mostly float conversions) there isn't any build failures.  The examples do use vsg::Array etc.  This is the result from the DLL build:

[vsgExamples automated build](https://github.com/vsg-dev/vsgExamples/runs/3578797065)

Which is curious, I would have expected the same errors as in your test case.  Could this be an indication that the test case isn't correctly linking to the lib?

rhabacker:
> I would have expected the same errors as in your test case.

Did that build with gdal ? The mentioned error happens on linking symbols required by gdal_utils.cpp.

sbrkopac:
> > I would have expected the same errors as in your test case.
> 
> Did that build with gdal ? The mentioned error happens on linking symbols required by gdal_utils.cpp.

I have opened a [draft PR](https://github.com/vsg-dev/vsgGIS/pull/14) in vsgGIS to enable CI builds. Please ignore the macos error as I haven't finished that up yet.

You can see the error on [this run](https://github.com/sbrkopac/vsgGIS/runs/3579628412). @robertosfield sorry for any confusion regarding the naming of the CI tests.

robertosfield:
By removing the export for the Value and Array classes and adding the 

#include <vsg/core/Visitor.h>
#include <vsg/core/ConstVisitor.h>

To the test-linking-class-templates.cpp the Windows DLL now builds cleanly, check the top item out, where the build was finally fixed.
[
https://github.com/vsg-dev/VulkanSceneGraph/commits/rhabacker-AddClassTemplateLinkTestCase](
https://github.com/vsg-dev/VulkanSceneGraph/commits/rhabacker-AddClassTemplateLinkTestCase)

robertosfield:
I modified the test adding the include <vsg/core/ConstVisitor.h> and include <vsg/core/Visitor.h> and it compiles without errors.  No need for any exports in Value/Array, and if one does add them then it breaks the Windows DLL build.

rhabacker:
> and adding the
> #include <vsg/core/Visitor.h>
> #include <vsg/core/ConstVisitor.h>

That means any client code which want to use templates must include that header files ?

PR #348, opened 2021-10-31
Changed fileExtension to closer mimic cpp17 filesystem::path::extension

sbrkopac:
## Description

In my application I am using the filesystem provided by the cpp17 standard. There is a mismatch between how VSG returns an extension and how the standard returns the extension. This becomes a bit cumbersome because I have to erase the first character coming off extension in the application. 

While I understand that VSG is avoiding the cpp17 filesystem due to compiler issues, I think it's reasonable that the VSG provided functions return as close as possible to the standard without re-implementing it.

**I have not done the work on any of the downstream VSG projects yet**

## Type of change

- [X] Breaking change (fix or feature that would cause existing functionality to not work as expected)

## How Has This Been Tested?

I created a test application to test some basic features of the ```extension``` function provided by the standard. 

[vsg-fs-extension.txt](https://github.com/vsg-dev/VulkanSceneGraph/files/7448189/vsg-fs-extension.txt)

**Test Configuration**:
* Hardware: Windows 10 x64
* Toolchain: MSVC 2019 16.11.5
* SDK: Vulkan 1.2.182.0

## Checklist:


robertosfield:
This is a bit of awkward one, the VSG's fileExtension() method mirrors what OSG does, so applications that are porting from OSG will prefer the current incarnation, but those mixing with std::filestream  will find a incompatibility.

Could you give an example of code in your application is currently more awkward because of the VSG convention of stripping the .?  It might be that we can provide a method handles this automatically.

robertosfield:
FYI, the file path handling I've written so far is a first pass implementation, where I implemented what was required to get basic reading/writing of files with vsgXchange.

My plan has to be refactor the file path handling prior to 1.0, primarily to add support for utf8/wstring support to enable wide strings to be pass to Windows ifstream.  The OSG had a bit of this functionality but is a kludge, so I'd to improve upon this.  The std::filesystem::path also doesn't handle this area as well I'd like either, so I'd like to take a different route.

Exactly what the implementation details will be I don't know yet.  This work will likely need to wait till December.

sbrkopac:
I had some legacy stuff hanging around after porting to ```FindFileCallback``` so I've cleaned up that mess. After my cleanup it becomes more of a clarity issue than it does a usability issue.

For instance in the VSG code base - I'd like to keep it as VSG-esque as possible for things like PR and bug reports. This means in the loaders, I try to keep the checks as ```if (vsg::fileExtension(filename) != "map")```. However, if I mix in ```std::filesystem``` the checks everywhere else in the code base become ```if(path.extension() == ".map)"```.  Part of the motivation for this is to make sure that if I need VSG support - I don't brush up against the ```std::experimental``` issue.

W.R.T the FileSystem rewrite - I'm glad to hear that you will be taking another pass at this prior to 1.0. I currently use a very simplified FileSys interface because I swap between a VFS and the LocalFileSys. It would be really nice to see VSG support a thin interface to help facilitate this in the future. Perhaps extending out the FindFileCallback into a IFileSystem interface? This discussion can wait until December though.

robertosfield:
HI Sam,

Perhaps a matchExtension(file, extension) overloaded to handle
std::filesystem::[ath and vsg::Path would be a way of avoiding issues with
your code.

W.r.t IFileSystem interface, I'm curious about what this might entail,
pointers to existing implementations that do something similar will
probably be a good guide.  Leaving this discussion till the time I'm
looking at file system will be best as I have enough items to think about
right now.

Cheers,
Robert.

PR #356, opened 2021-11-10
Add support for finding a minimum version of the Vulkan package for cmake < 3.23

### LINKED ISSUE #312: Missing support to detect supported Vulkan SDK version
**Does your feature request relate to a problem? Please describe it.
vsg does not recognize the version of the Vulkan SDK being used, which would prevent version-related issues such as those reported at https://github.com/vsg-dev/VulkanSceneGraph/issues/311.

**Describe the solution you want**.
vsg should specify the minimum required version when finding the Vulkan SDK, which can be specified by an additional argument to the call to cmake find_package():
```
...
set(Vulkan_MIN_VERSION 1.2.162.0)
...
find_package(Vulkan ${Vulkan_MIN_VERSION} REQUIRED)
```
which causes cmake to report any problems finding the required version. 

**Additional context**.
Unfortunately, the currently available cmake module for finding [Vulkan libraries](https://github.com/Kitware/CMake/blob/86ee38d75678d68bb1a79b9fa443b42723b09f23/Modules/FindVulkan.cmake#L68) does not include version support.

rhabacker:
## Description

With cmake >=3.23, there will be support for checking the installed Vulkan version against a minimum required version. For older cmake versions, the commit of this mr adds an identical check.
 
Fixes #312

## Type of change

## How Has This Been Tested?

- [x] With cmake < 3.23 installed, I built VulkanSceneGraph from this branch, installed it locally, and then built the vsgGIS project using the previously installed vsg package after setting the minimum required vsg version to one not yet available. The configuration of vsgGIS then fails with the expected message that the package "vsg" could not be found.

**Test Configuration**:
* cmake 3.17

## Checklist:


robertosfield:
Thanks for the changes @rhabacker.  The automated builds are failing on all platforms.  Could you look into this?

rhabacker:
> Thanks for the changes @rhabacker. The automated builds are failing on all platforms. Could you look into this?

from https://github.com/vsg-dev/VulkanSceneGraph/runs/4162516978?check_suite_focus=true
```
CMake Error at cmake/vsgMacros.cmake:365 (message):
  Could NOT find Vulkan: Found unsuitable version "#define
  VK_HEADER_VERSION_COMPLETE VK_MAKE_VERSION(1, 2, VK_HEADER_VERSION)", but
  required is at least "1.2.162" (found
  /home/runner/work/VulkanSceneGraph/1.2.162.1/x86_64/lib/libvulkan.so)
```
The regex did not match

rhabacker:
> The regex did not match

The related line in the installed version 1.2.162 is `#define VK_HEADER_VERSION_COMPLETE VK_MAKE_VERSION(1, 2, VK_HEADER_VERSION)`
where expected is `"#define VK_HEADER_VERSION_COMPLETE VK_MAKE_API_VERSION(0, 1, 2, VK_HEADER_VERSION)"`

They changed that macro in version 1.2.175 - https://github.com/KhronosGroup/Vulkan-Headers/commit/e01b00657664ac34389d51b04c58dff3eb19a026

robertosfield:
On Wed, 10 Nov 2021 at 08:43, Ralf Habacker ***@***.***>
wrote:

> The regex did not match
>
> The related line in the installed version 1.2.162 is #define
> VK_HEADER_VERSION_COMPLETE VK_MAKE_VERSION(1, 2, VK_HEADER_VERSION)
> where expected is "#define VK_HEADER_VERSION_COMPLETE
> VK_MAKE_API_VERSION(0, 1, 2, VK_HEADER_VERSION)"
>
> They changed that macro in version 1.2.175 - KhronosGroup/Vulkan-Headers@
> e01b006
> <https://github.com/KhronosGroup/Vulkan-Headers/commit/e01b00657664ac34389d51b04c58dff3eb19a026>
>

But there isn't a reason other than this things shouldn't build against
1.2.162.  We can't have builds failing because of one marco.



>

rhabacker:
> 62. We can't have builds failing because of one marco.

The macro must be customized to match all versions - this is the disadvantage of using a non-"cmake"-standard method to detect a package version. The use of the form major.minor.patch corresponds to the conventions of the tags used on https://github.com/KhronosGroup/Vulkan-Loader/tags and the packages in distributions e.g. https://software.opensuse.org/package/vulkan-loader.

The more "stable" way would be to use the official version check support of the vulkan development package.

Currently available is pkgconfig support

```
$ rpm -q -l vulkan-devel
/usr/lib64/libvulkan.so
/usr/lib64/pkgconfig/vulkan.pc
```
```
$ cat /usr/lib64/pkgconfig/vulkan.pc
...
Version: 1.2.189
...
```
Possible, but not currently implemented, would be to add cmake support files to the Vulkan development package.

robertosfield:
With problems like these, or requirements to add extra 3rd party
dependencies it strikes me that it's not an ideal solution.

Checking the Vulkan version is nice thing to add, but we have to avoid
adding extra hurdles for this nicity.  We have to make sure the benefit is
with the pain, at this point it really doesn't feel like we are near this
point.

Another way would be to add a Cmake build against the Vulkan to test the
Vulkan version like we presently do for glslang, or make the header checks
handle these different versions more gracefully.

rhabacker:
I also pushed the fix upstream https://gitlab.kitware.com/cmake/cmake/-/merge_requests/6717

robertosfield:
Thanks for the fixes, shame the script got more complicated but looks to be working well :-)

PR #369, opened 2021-11-21
Convert DatabasePager to use CollectResourceRequirements.

rolandhill:
PagedLod subgraphs require resource requirements for the entire quad tree to be known ahead of time. Dynamically added PagedLods will struggle with this. This change allows resource requirements to be assessed on a leaf by leaf basis.

## Type of change
Applied the CollectResourceRequirements from Robert's DynamicLoadAndCompile code to the DatabasePager.

**Note: I'm not confident with this and you should consider any other implications that this might have. This PR is supplied on a 'seems to work' basis.**

- [ X?] Breaking change (fix or feature that would cause existing functionality to not work as expected)

## How Has This Been Tested?
Working in my own application with a dynamically added PagedLod subgraph (elevation model with air photo overlay). Looks to be OK without any obvious issues. The base Vulkan setup is added at the PagedLod root and the textures and DescriptorSets are added for each tile, based on the vsg PagedLod example.


robertosfield:
I understand the motivation behind these changes, it's a more flexible than the previous reliance on assignment of vsg::ResourceHints to scene graph, however, it's less efficient than preallocating the resources. 

Compromising efficiency for flexibility is something we need to think carefully about.  Compromises like this suggest that there is something missing for VSG design and implementation, and in this instance, it's actually one I've been aware of since the early days of the VSG - solving the need to dynamically allocation DescriptorPools on demand is something that needs to be done more seamlessly by the VSG.

I'm going to change this PR to a draft and leave it to one side till I do the review the dynamic resource management that I've been planning to get to.  This will need to be after Wednesday as my focus will be on client work till then,

rolandhill:
I suspected there would be a performance issue for doing this. For now,
this change works for me and allows me to proceed with my application but I
look forward to switching to your higher performance solution.

Regards,
Roland

robertosfield:
@rolandhill could you submit the vsg::quat changes as a separate PR?

rolandhill:
Hi Robert,

Sure. I didn't realise they would propagate to this draft PR.

Roland

robertosfield:
My plan this week is start tackling the backlog of issues reported on VSG, vsgXchange and vsgExamples.  This includes this DatabasePager one.

PR #391, opened 2022-01-09
LineSegmentIntersector crash

rolandhill:
# Pull Request Template

## Description

LineSegmentIntersector::intersectDrawIndexed(uint32_t firstIndex, uint32_t indexCount) can segfault if indexCount is not a multiple of 3. The file producing this crash also shows some artefacts that might be in the original file or might be due to the assimp/vsgXchange import process (the subject of further investigation), but these changes make VSG more robust and poorly formed files are part of the real world anyway.

1) return immediately if there are < 3 indices.
2) loop finishes before the last multiple of 3 indices to ensure that i+1 and i+2 do not go past the end of the array.

## How Has This Been Tested?
On a DXF file that was causing the crash

VSG 9/1/2022


rolandhill:
Note that applying the same fix to intersectDraw has not been tested! I only have a file that uses intersectDrawIndexed.

robertosfield:
I'll merge this workaround for bad data as it's such a lighweight fix - no more work than the original code.  However, the data that trips the intersection code up will certain cause problems elsewhere - you just might not see it.

Catching erroneous data is pretty open ended task, but might be worth exploring as a topic on vsg-users google group.  Perhaps some extra scene graph validate utilities would be appropriate.

rolandhill:
The cause was a DXF file containing lines and faces being interpreted as
all faces by vsgXchange. I've sent a PR to resolve that issue.

Regards,
Roland

PR #394, opened 2022-01-12
Added 64bit vsg::Mask

robertosfield:


sbrkopac:
On windows this starts to throw the following warning:

```1>C:\Programming\libraries\VulkanSceneGraph\bin\include\vsg/viewer/View.h(34,43): warning C4244: 'initializing': conversion from 'vsg::Mask' to 'uint32_t', possible loss of data
```

Not sure if it's something to be concerned about or not.

robertosfield:
On Wed, 12 Jan 2022 at 17:59, Sam Brkopac ***@***.***> wrote:

> On windows this starts to throw the following warning:
>
> Not sure if it's something to be concerned about or not.
>
> It's the end of the working day here so I'll check out the example and
have a tinker tomorrow.

PR #405, opened 2022-02-21
Add file system install related functions

rhabacker:
## Description

With this merge request some installation related functions are added to the vsg FileSystem module. They will be used for example by vsgvr to find installed model files.

## Type of change

## How Has This Been Tested?
I used the following test case to verify that the functions returns the expected values.

```
$ cat test.cpp
#include <iostream>

#include <vsg/io/FileSystem.h>

int main()
{
    vsg::Path path;
    path = vsg::executableFilePath();
    std::cout << path << std::endl;
    path = vsg::executablePath();
    std::cout << path << std::endl;
    path = vsg::sharedDataPath();
    std::cout << path << std::endl;
    path = vsg::sharedAppDataPath();
    std::cout << path << std::endl;
    return 0;
}
```
Linux
```
$ cd ~/src
$ git clone https://github.com/vsg-dev/VulkanSceneGraph.git
$ mkdir VulkanSceneGraph-build
$ cd VulkanSceneGraph-build
$ cmake ../VulkanSceneGraph
$ make
g++ -o bin/test test.cpp -I../VulkanSceneGraph/include  -lvsg -Llib
$ mkdir share
$ LD_LIBRARY_PATH=$PWD/lib ./bin/test
/home/xxx/src/VulkanSceneGraph-build/bin/test
/home/xxx/src/VulkanSceneGraph-build/bin
/home/xxx/src/VulkanSceneGraph-build/share
/home/xxx/src/VulkanSceneGraph-build/share/test
```
Windows
```
cd ~/src
$ mkdir VulkanSceneGraph-mingw32-build
$ cd VulkanSceneGraph-mingw32-build
$ mingw32-cmake ../VulkanSceneGraph
$ make
$ i686-w64-mingw32-g++ -o bin/test test.cpp -I../VulkanSceneGraph/include  -lvsg -Llib
$ mkdir share
$ wine bin/test.exe
Z:\home\xxx\src\VulkanSceneGraph-mingw32-build\bin\test.exe
Z:\home\xxx\src\VulkanSceneGraph-mingw32-build\bin
Z:\home\xxx\src\VulkanSceneGraph-mingw32-build\share
Z:\home\xxx\src\VulkanSceneGraph-mingw32-build\share\test
```

## Checklist:
- [x] I have added tests that prove my fix is effective or that my feature works -> see above
- [ ] New and existing unit tests pass locally with my changes -> needs advice where to add related unit test as vsgExamples only contains code examples

### DIFF
--- a/src/vsg/io/FileSystem.cpp
+++ b/src/vsg/io/FileSystem.cpp
@@ -19,11 +19,27 @@ THE SOFTWARE IS PROVIDED "AS IS", WITHOUT WARRANTY OF ANY KIND, EXPRESS OR IMPLI
 #    include <io.h>
 //	 cctype is needed for tolower()
 #    include <cctype>
+#    include <windows.h>
+
+#ifdef _MSC_VER
+#ifndef PATH_MAX
+#define PATH_MAX MAX_PATH
+#endif
+#endif
+
 #else
 #    include <errno.h>
 #    include <sys/stat.h>
 #    include <unistd.h>
 #endif
+
+#ifdef __APPLE__
+#    include <libgen.h>
+#    include <mach-o/dyld.h>
+#    include <TargetConditionals.h>
+#endif
+
+#include <limits.h>
 
 #include <iostream>
 
@@ -277,4 +293,89 @@ bool vsg::makeDirectory(const Path& path)
     }
 
     return true;
+}
+
+Path vsg::executableFilePath(void)
+{
+    Path path;
+    char buf[PATH_MAX+1];
+
+#if defined(WIN32)
+    DWORD result = GetModuleFileName(NULL, buf, sizeof(buf)-1);
+    if (result && result < sizeof(buf))
+        path = buf;
+#elif defined(__linux__)
+    ssize_t len = readlink("/proc/self/exe", buf, sizeof(buf)-1);
+    if (len != -1)
+        path = buf;
+#elif defined(__APPLE__)
+#if TARGET_OS_MAC
+    char realPathName[PATH_MAX+1];
+    uint32_t size = (uint32_t)sizeof(buf);
+
+    if(!_NSGetExecutablePath(buf, &size))
+    {
+        realpath(buf, realPathName);
+        path = realPathName;
+    }
+#elif TARGET_IPHONE_SIMULATOR
+    // iOS, tvOS, or watchOS Simulator
+    // Not currently implemented
+#elif TARGET_OS_MACCATALYST
+    // Mac's Catalyst (ports iOS API into Mac, like UIKit).
+    // Not currently implemented
+#elif TARGET_OS_IPHONE
+    // iOS, tvOS, or watchOS device
+    // Not currently implemented
+#else
+#   error "Unknown Apple platform"
+#endif
+#elif defined(__ANDROID__)
+    // Not currently implemented
+#endif
+    return path;
+}
+
+Path vsg::executablePath(void)
+{
+    return vsg::filePath(executableFilePath());
+}
+
+Path vsg::sharedDataPath()
+{
+    Path path;
+
+#if defined(WIN32) || defined(__linux__)
+    path = executablePath();
+    if (path.empty())
+        return {};
+    auto sharePath = concatPaths(path, std::string(CMAKE_INSTALL_DATADIR));
+    if (fileExists(sharePath))
+        return sharePath;
+    path = vsg::filePath(path);
+    sharePath = concatPaths(path, std::string(CMAKE_INSTALL_DATADIR));
+    if (fileExists(sharePath))
+        return sharePath;
+#elif defined(__APPLE__)
+    // Not currently implemented
+#elif defined(__ANDROID__)
+    // Not currently implemented
+#endif
+    return {};
+}
+
+Path vsg::sharedAppDataPath()
+{
+#if defined(WIN32) || defined(__linux__)
+    Path appName = simpleFilename(executableFilePath());
+    Path appDataDir = sharedDataPath();
+    if (appDataDir.empty())
+        return {};
+    return concatPaths(sharedDataPath(), appName);
+#elif defined(__APPLE__)
+    // Not currently implemented
+#elif defined(__ANDROID__)
+    // Not currently implemented
+#endif
+    return {};
 }
--- a/include/vsg/io/FileSystem.h
+++ b/include/vsg/io/FileSystem.h
@@ -69,6 +69,39 @@ namespace vsg
     /// make a directory, return true if path already exists or full path has been created successfully, return false on failure.
     extern VSG_DECLSPEC bool makeDirectory(const Path& path);
 
+    /// returns the path and filename of the currently executed file
+    extern VSG_DECLSPEC Path executableFilePath(void);
+
+    /// returns the path of the currently executed file
+    extern VSG_DECLSPEC Path executablePath(void);
+
+    /** @brief returns the absolute path to the shared data installation directory at runtime
+     *
+     * The supported path layouts:
+     * <pre>
+     *   <root>/
+     *     <executable>
+     *     share/ [1]
+     *
+     *   <root>/
+     *     bin/
+     *       <executable>
+     *     share/ [1]
+     * </pre>
+     *
+     * @return path to shared data installation directory [1] or empty path in case of unsupported path layout
+     */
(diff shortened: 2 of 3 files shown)


src/vsg/io/FileSystem.cpp
```diff
@@ -278,3 +279,47 @@ bool vsg::makeDirectory(const Path& path)
...
+
+Path vsg::executablePath(void)
+{
+    return vsg::filePath(executableFilePath());
+}
+
+Path vsg::sharedDataPath()
+{
+    Path path = executablePath();
+    Path sharePath = path + std::string(1, delimiterNative) + std::string("share");
```
robertosfield, line 308: Shuld be possible to replace with concatPaths(path, "share");

src/vsg/io/FileSystem.cpp
```diff
@@ -278,3 +279,47 @@ bool vsg::makeDirectory(const Path& path)
...
+    if (fileExists(sharePath))
+        return sharePath;
+    path = vsg::filePath(path);
+    sharePath = path + std::string(1, delimiterNative) + std::string("share");
+    if (fileExists(sharePath))
+        return sharePath;
+    return Path();
```
robertosfield, line 315: Should us {}

src/vsg/io/FileSystem.cpp
```diff
@@ -278,3 +279,47 @@ bool vsg::makeDirectory(const Path& path)
...
+{
+    Path path;
+    char buf[1024];
+
+#if defined(WIN32)
+    DWORD result = GetModuleFileName(NULL, buf, sizeof(buf)-1);
+    if (result && result < sizeof(buf))
+        path = buf;
+#else
+    ssize_t len = readlink("/proc/self/exe", buf, sizeof(buf)-1);
```
robertosfield, line 293: Will this work under Android , macOS,and iOS?

src/vsg/io/FileSystem.cpp
rhabacker: It only works using `concatPaths(path, std::string("share"));`

src/vsg/io/FileSystem.cpp
rhabacker: I added related cpp macros as place holder for currently unimplemented platforms. For mac osx I did found a hint how to implement that.

robertosfield:
Thanks for the updates.  I'm torn whether all of the new functions are core VSG features or not.  I can see the getting of the executable path as being universal, but the shared data path is more in realm of application specific convention.  

It may be better to make it easy for applications to set up the search path conventions they use, rather than encode the convention into a .cpp.

rhabacker:
This is the code fragment, that I was preparing to add to vsgvr code base:

example_vr.cpp
```
int main(int argc, char **argv) {
  try
  {
    // set up defaults and read command line arguments to override them
    vsg::CommandLine arguments(&argc, argv);

    // set up vsg::Options to pass in filepaths and ReaderWriter's and other IO
    // realted options to use when reading and writing files.
    auto options = vsg::Options::create();
    arguments.read(options);

    vsg::Paths searchPaths = vsg::getEnvPaths("VSG_FILE_PATH");
    // prepend executable path
    char buf[1024];
#ifdef __linux__
    ssize_t result = readlink("/proc/self/exe", buf, sizeof(buf));
    if (result != -1 && result < sizeof(buf))
        searchPaths.insert(searchPaths.begin(), vsg::filePath(buf));
#elif _WIN32
    DWORD result = GetModuleFileName(NULL, buf, sizeof(buf));
    if (result && result < sizeof(buf))
        searchPaths.insert(searchPaths.begin(), vsg::filePath(buf));
#endif
   // TODO: add Windows and app image support
    searchPaths.push_back(VSGVR_INSTALL_DATADIR);
    vsg::Path filename = vsg::findFile("world.vsgt", searchPaths);
```

In vsgvr code base there is in the toplevel CMakeLists.txt
```
# Copy models into the build dir
configure_file(models/world/world.vsgt ${CMAKE_CURRENT_BINARY_DIR}/world.vsgt COPYONLY)
configure_file(models/controller/controller.vsgt ${CMAKE_CURRENT_BINARY_DIR}/controller.vsgt COPYONLY)
configure_file(models/controller/controller2.vsgt ${CMAKE_CURRENT_BINARY_DIR}/controller2.vsgt COPYONLY)
```
which is required to run the application from the build dir and depends on 
```
    // prepend executable path
    char buf[1024];
#ifdef __linux__
    ssize_t result = readlink("/proc/self/exe", buf, sizeof(buf));
    if (result != -1 && result < sizeof(buf))
        searchPaths.insert(searchPaths.begin(), vsg::filePath(buf));
#elif _WIN32
    DWORD result = GetModuleFileName(NULL, buf, sizeof(buf));
    if (result && result < sizeof(buf))
        searchPaths.insert(searchPaths.begin(), vsg::filePath(buf));
#endif
 ```
This has been added to vsg by vsg::executablePath() and reduces the mentioned code to 
```
        searchPaths.insert(searchPaths.begin(), vsg::executablePath());
```
For finding the installed files I added to the toplevel CMakeLists.txt:
```
add_definitions(-DVSGVR_INSTALL_DATADIR="${CMAKE_INSTALL_FULL_DATADIR}/vsgvr")
...
install(FILES
    models/world/world.vsgt
    models/controller/controller.vsgt
    models/controller/controller2.vsgt
    DESTINATION ${CMAKE_INSTALL_DATADIR}/vsgvr
)
```
and to the code 
```
    searchPaths.push_back(VSGVR_INSTALL_DATADIR);
```
This only works on linux with regular distributions, but not on Windows and not for app images, where the packages needs to be portable, having a different install root at runtime compared with build time.

The default file path layout for vsg based installation packages on Windows, provided by the cmake build system, looks like this:
```
<root>
   bin/
   share/
       <appname>/
   lib/
      pkgconfig/
      cmake/
     ...
```
The only determinable path by api is the executable path, from which all other paths needs to be constructed, something like this: 
```
#if defined(WIN32) || <build-for-app-image>
    auto appName = simpleFilename(executableFilePath());
    auto path = vsg::filePath(executablePath());
    auto sharedPath = concatPaths(path, std::string("share"));
    auto sharedAppDataPath = concatPaths(sharedPath, appName);
    searchPaths.push_back(sharedAppDataPath);
#elif defined(__linux__)
    searchPaths.push_back(VSGVR_INSTALL_DATADIR);
#elif defined(__ANDROID__)
    ...
#endif
```
Any package which depends on installed read only files like shader, model or other files would need to implement such stuff by itself, which has been added to vsg by vsg::sharedAppDataPath() to reduce that burden.

Please note that using these functions is optional, but looking at current vsg based application I see vsgExamples, vsgGIS, vsgvr4 and maybe vsgQt in future as candidates and would benefit from this support function.

To support special needs, the mentioned function can be extended to get an optional parameter to override the default name like `vsg::sharedAppDataPath(const std::string &customAppName = {})`.

Also to avoid a hard coded relative data path name used in `auto sharedPath = concatPaths(path, std::string("share"));`, here CMAKE_INSTALL_DATADIR can be used.

robertosfield:
I have create a [executableFilePath](https://github.com/vsg-dev/VulkanSceneGraph/tree/executableFilePath) branch and add the Path vsg::executableFilePath() function from this PR.

I also modified the linux code path to add in the \0 c string ended as ::readlink doesn't do this and added a TODO note about PATH_MAX not being ideal.  With the \0 set the Linux path now oorrectly returns the full executable path without any random characters.

I only have Linux systems here so can't test the other code paths.

I haven't added the other functions from this PR as feel that are too application specific, and the CMAKE_INSTALL_DATADIR would be only appropriate to the build of the VSG library itself.  Application developers will have their own CMAKE_INSTALL_DATADIR value when building their own application, which really is the one they should be using.

If you think there is merit to adding some VSG library specific data path then please raise this as topic for general discussion on vsg-users.

As for the creation of paths relative to the executable path - it should be possible to do something like:

  auto appFilePath = vsg::executableFilePath();
  auto appName = vsg::simpleFilenem(appFilePath);
  auto appPath = vsg::filePath(exeFilePath);
  auto dataPath = vsg::concatPaths(executablePath, vsg::Path{"../share"}, appName);

This requirement seems very application specific and as it's easy to implement with the above should not be a big issue for those users that need it implementing it exactly how they want for their app needs.

robertosfield:
I have merged the executableFilePath branch with master.

PR #408, opened 2022-02-25
Changed ci to use a matrix approach

sbrkopac:
## Description

Cleaned up the continuous integration so its much cleaner and can be easily extended using the matrix.

## Type of change

## Checklist:

### DIFF
--- a/.github/workflows/build-docs.yml
+++ b/.github/workflows/build-docs.yml
@@ -1,0 +1,25 @@
+name: build-docs
+on:
+  push:
+    branches:
+      - main
+jobs:
+  build:
+    runs-on: ubuntu-latest
+    steps:
+      - uses: actions/checkout@v2
+      - name: Install Doxygen
+        run: sudo apt install doxygen graphviz
+      - name: Install Vulkan
+        uses: humbletim/install-vulkan-sdk@v1.1.1
+        with:
+          version: latest
+      - name: Configure VulkanSceneGraph
+        run: cmake -S . -B . -DCMAKE_BUILD_TYPE=Release
+      - name: Build Documentation
+        run: make -j 2
+      - name: Publish Docs
+        uses: actions/upload-artifact@v1
+        with:
+          name: VSG Documentation
+          path: html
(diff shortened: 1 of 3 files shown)


sbrkopac:
I believe so: 
```
-- The following OPTIONAL packages have been found:

 * vsg_glslang
```
Unless I need to enable it via a cmake flag? I just checked on my local windows machine and I can see that code path being compiled.

sbrkopac:
It looks like the way Khronos is packaging things has evolved a little bit. The macOS DMG is now an installer rather than just packaged files so I've asked for support in the here humbletim/install-vulkan-sdk#3.

@robertosfield What are your thoughts on bumping the minimum requirement to Vulkan Version ```1.2.176.1```? The other option is we could just bump the CI and handle exceptions as they come.

robertosfield:
On Fri, 25 Feb 2022 at 20:22, Sam Brkopac ***@***.***> wrote:

> It looks like the way Khronos is packaging things has evolved a little
> bit. The macOS DMG is now an installer rather than just packaged files so
> I've asked for support in the here humbletim/install-vulkan-sdk#3
> <https://github.com/humbletim/install-vulkan-sdk/issues/3>.
>
> @robertosfield <https://github.com/robertosfield> What are your thoughts
> on bumping the minimum requirement to Vulkan Version 1.2.176.1? The other
> option is we could just bump the CI and handle exceptions as they come.
> Message ID: ***@***.***>

Bump the min requirements for the automated build?  Is it something that
you could just do for macOS is that's causing problems?

robertosfield:
I have little expertise on  CI, the changes look cleaner, but not in a good position to provide a proper technical review so happy to defer to others who've spent more time with it.  @sbrkopac do you feel this work is ready to merge now?

sbrkopac:
Yes, I'm pretty happy with this. This covers all the functionality that we had before. @robertosfield I haven't had a chance to review the [Safety Critical](https://www.khronos.org/news/press/khronos-releases-vulkan-safety-critical-1.0-specification-to-deliver-safety-critical-graphics-compute) update yet. But we can get that in at a later date.

PR #414, opened 2022-03-08
Add CollectResourceRequirements to DatabasePager

rolandhill:
Resubmission of previous PR (I got my branches in a mess).

Resource requirements are calculated and applied per DatabasePager request, specifically so that PagedLOD tiles can be loaded without previous knowledge of the number of tiles.

### DIFF
--- a/src/vsg/io/DatabasePager.cpp
+++ b/src/vsg/io/DatabasePager.cpp
@@ -355,6 +355,15 @@ void DatabasePager::start()
                             // compiling subgraph
                             if (subgraph)
                             {
+                                vsg::CollectResourceRequirements collectRequirements;
+                                subgraph->accept(collectRequirements);
+
+                                auto maxSets = collectRequirements.requirements.computeNumDescriptorSets();
+                                auto descriptorPoolSizes = collectRequirements.requirements.computeDescriptorPoolSizes();
+
+                                // brute force allocation of new DescrptorPool for this subgraph, TODO : need to preallocate large DescritorPoil for multiple loaded subgraphs
+                                if (descriptorPoolSizes.size() > 0) ct->context.descriptorPool = vsg::DescriptorPool::create(ct->context.device, maxSets, descriptorPoolSizes);
+
                                 subgraph->accept(*ct);
                                 nodesCompiled.emplace_back(plod);
                             }


robertosfield:
My plan for next week is to refactor how DescriptorPools are managed in the VSG, with an objective of making their use more flexible and better able to handle dynamic use.  This may well avoid the need for your PR completely, but I'll leave it up as a draft until we've confirmed that it's no longer needed.

When testing the changes I'll make next week it would be useful to have concrete test cases that flesh out the problems that you've addressed in this PR.  Would it be possible to share a dataset that causes problems that this PR addressed?  I'm happy to sign NDA's if required.

rolandhill:
Hi Robert,

There is no one particular dataset that causes a problem. All datasets are
loaded dynamically for me. Some of these are regular graphics file (eg the
DXF file that I've sent to you before), some are generated dynamically from
CSV data or data from an SQLite database, and some are tiled surfaces
dynamically generated from a digital elevation model and satellite
image/air photo overlays (which I don't know the resolution, therefore tile
size/extent/maximum lod level until the user loads the files).

My VSG setup basically goes (some steps skipped)

_sceneRoot = vsg::Group::create();
> _view->addChild(_sceneRoot);
> _viewer->compile();


At which point, there is nothing in the scene. Then everything else
(including IMGUI) is added via a DynamicLoadAndCompile class based on one
of your examples, either within the setup function or later by the user

auto gui = vsgImGui::RenderImGui::create(_window);
> _dlc->compileRequest(gui, renderGraph); // (subgraph to compile,
> attachment point)


The DatabasePager is installed and started manually because the compile
method only loads it if there are PagedLODs present.

I didn't used to call ->compile at all, but I have it in my mind that
ViewDependantState needs a call to compile to get set up properly.

My DynamicLoadAndCompile class is attached. It has an additional method to
allow a subgraph to be added without loading it from disk.

The changes to DBPager in my PR aren't for any special problematic dataset,
they are for a dataset that you don't know its size ahead of time so can't
allocate sufficient resources.

Regards,
Roland

robertosfield:
Thanks for the explanation.

What I'll do is modify the handling of ResourceHints in the paged databases
I have so that preallocation doesn't happen like it was originally designed
to do.  This will break the loading pretty quick and be a reasonable proxy
for your usage approach.

robertosfield:
@rolandhill  I have successfully hacked ResoureRequirements.cpp to prevent the normal handling of the ResourceHints in paged database so I get a crash pretty quickly, which is then resolved if I apply your PR.  To provoke the crash I simple run:

   vsgviewer models/openstreetmap.vsgt

The changes I made to provoke the crash was:

$ git diff src/vsg/vk/ResourceRequirements.cpp
diff --git a/src/vsg/vk/ResourceRequirements.cpp b/src/vsg/vk/ResourceRequirements.cpp
index 7469b3c6..36aae820 100644
--- a/src/vsg/vk/ResourceRequirements.cpp
+++ b/src/vsg/vk/ResourceRequirements.cpp
@@ -84,6 +84,8 @@ void CollectResourceRequirements::apply(const Object& object)
 
 bool CollectResourceRequirements::checkForResourceHints(const Object& object)
 {
+    return false;
+
     auto resourceHints = object.getObject<ResourceHints>("ResourceHints");
     if (resourceHints)
     {

robertosfield:
I have checked in an experiment with tracking of available descriptors in vsg::DescriptorPool and automatic allocation of new DescriptorPool when resources run out.  @rolandhill could you try it out?

The changes are in the [DescriptorSetExperiement](https://github.com/vsg-dev/VulkanSceneGraph/tree/DescriptorSetExperiement) branch.

I'm also writing up this work on [vsg-users](https://groups.google.com/g/vsg-users/c/JJQZ-RN7jC0/m/dQkp-0neAwAJ?utm_medium=email&utm_source=footer).

vsg-dev:
I have now merged #451 the DynamicDescriptorPool work that makes this PR obsolete so after 6 weeks can finally close this PR :-)

PR #417, opened 2022-03-17
Fixing issue with window resizing and vkAcquireNextImageKHR eventuall…

valleymouth:
…y not returning VK_ERROR_OUT_OF_DATE_KHR.

# Pull Request Template

## Description

Apparently the behaviour of vkAcquireNextImageKHR is not consistent when window resizing happens, and sometimes it doesn't report VK_ERROR_OUT_OF_DATE_KHR. This forces window resizing without depending on vkAcquireNextImageKHR, just checking the current swapchain extent with the new resized extent.

## Type of change

## How Has This Been Tested?

- [x] Use vsgQt example vsgqtviewer, and make sure to run it with -d option. Resize the window drastically. You should see some vulkan errors being reported, basically complaining about the framebuffer size. You should also see visual artefacts caused by a corrupted framebuffer.

**Test Configuration**:
Tested on a Ubuntu 20.04, vsg 2a12449a305d9b1c1abf066174ae76b4bc0c3738, vsgQt b0eb106bf71fd0f561882a4af7d4df3366caf512, Qt 5.12.8, NVidia T500 (510.47.03 driver).

## Checklist:

### DIFF
--- a/src/vsg/viewer/Viewer.cpp
+++ b/src/vsg/viewer/Viewer.cpp
@@ -152,6 +152,14 @@ bool Viewer::acquireNextFrame()
     {
         if (!window->visible()) continue;
 
+        // This is a workaround to force resize without depending on vkAcquireNextImageKHR returning VK_ERROR_OUT_OF_DATE_KHR,
+        // because apparently it is not behaving consistently. Driver bug maybe?
+        if (window->getSwapchain() && (window->extent2D().width != window->getSwapchain()->getExtent().width ||
+                                       window->extent2D().height != window->getSwapchain()->getExtent().height))
+        {
+            window->resize();
+        }
+
         while ((result = window->acquireNextImage()) != VK_SUCCESS)
         {
             if (result == VK_ERROR_SURFACE_LOST_KHR ||


robertosfield:
I quick update, now that I've completed all my MUST HAVE feature items for 1.0 I'm now catching up with all the Issue/bug fixes.  I've been testing vsgqtviewer on my AMD5700G system and don't see any visual problems but when running with -d I do see errors being reported:

VUID-VkRenderPassBeginInfo-renderArea-02848(ERROR / SPEC): msgNum: -1615083365 - Validation Error: [ VUID-VkRenderPassBeginInfo-renderArea-02848 ] Object 0: handle = 0xfab64d0000000002, type = VK_OBJECT_TYPE_RENDER_PASS; | MessageID = 0x9fbbc89b | vkCmdBeginRenderPass: Cannot execute a render pass with renderArea not within the bound of the framebuffer, renderArea.offset.x (0) + renderArea.extent.width (2349) is greater than framebuffer width (2303). The Vulkan spec states: renderArea.offset.x + renderArea.extent.width must be less than or equal to VkFramebufferCreateInfo::width the framebuffer was created with (https://vulkan.lunarg.com/doc/view/1.3.204.1/linux/1.3-extensions/vkspec.html#VUID-VkRenderPassBeginInfo-renderArea-02848)

--

Curiously if I switch to using a vsg::Surface rather than a QSurface I don't see any of these warnings when resizing:

     vsgqtviewer models/lz.vsgt -d  --vsg

I will keep digging to see if I can figure out why the vsg::Surface's native implementation works fine but the QSurface doesn't.  My hope is that we won't need to put a workaround in the core VSG for an issue that is specific to Qt's wrapping of Vulkan.

@valleymouth if you've come across any more insights then this would be helpful in my investigation.

robertosfield:
@valleymouth I've ended up implementing a swaphain/window extents check like you did, but I've placed this check in Window::aquireNextImage(..) so that the Viewer::aquireNextImage() can be kept as is with no extra checks/resize handling. To keep the code clean I also added a != and == operators for VkExtent2D.  The commit is: ac6a8e3db32cc98098bfc638d6e63c57acc030e8

With this commit the vsgqtviewer example works reliably without emitting the Vulkan debug errors like it was so I believe this issue should now be resolved.

valleymouth:
Amazing Robert! Thanks for this!

PR #422, opened 2022-03-24
Window Position can be set now.

rainergericke:
Window traits position (x,y) are now used to set the window position.

It is the distance between lower left corner of the screen and the lower left corner of the window.

# Pull Request Template

## Description

Please include a summary of the change and which issue is fixed. Please also include relevant motivation and context. List any dependencies that are required for this change.

Fixes # (issue)

## Type of change

## How Has This Been Tested?

- [x] vsgbuilder, setting the x and y
- [x] chrono_vsg, setting per accessor method

**Test Configuration**:
* Firmware version: 7459.101.2
* Hardware: Apple iMac M1
* Toolchain: Apple Xcode 13.3
* SDK: Vulkan 1.3.204.0

## Checklist:

### DIFF
--- a/src/vsg/platform/macos/MacOS_Window.mm
+++ b/src/vsg/platform/macos/MacOS_Window.mm
@@ -797,8 +797,7 @@ MacOS_Window::MacOS_Window(vsg::ref_ptr<vsg::WindowTraits> traits) :
     [_window setOpaque:YES];
     [_window setBackgroundColor:[NSColor whiteColor]];
 
-    //[_window setFrame:CGRectMake(pos.x, ymax - pos.y, [_window frame].size.width , [_window frame].size.height) display:NO]; 
-   
+    
     // create view
     _view = [[vsg_MacOS_NSView alloc] initWithVsgWindow:this];
     [_view setWantsBestResolutionOpenGLSurface:_traits->hdpi];
--- a/src/vsg/platform/macos/MacOS_Window.mm
+++ b/src/vsg/platform/macos/MacOS_Window.mm
@@ -797,14 +797,14 @@ MacOS_Window::MacOS_Window(vsg::ref_ptr<vsg::WindowTraits> traits) :
     [_window setOpaque:YES];
     [_window setBackgroundColor:[NSColor whiteColor]];
 
-    // set the lower left corner window position as offset from the lower left corner of the screen
+    // set the top left corner window position as offset from the top left corner of the screen
     NSPoint pos; 
     int xmax = [[NSScreen mainScreen] frame].size.width - [_window frame].size.width;
     int ymax = [[NSScreen mainScreen] frame].size.height - [_window frame].size.height;
     pos.x = std::clamp(traits->x,0,xmax);
     pos.y = std::clamp(traits->y,0,ymax);
-    [_window setFrame:CGRectMake(pos.x, pos.y, [_window frame].size.width , [_window frame].size.height) display:YES];    
-
+   [_window setFrame:CGRectMake(pos.x, ymax - pos.y, [_window frame].size.width , [_window frame].size.height) display:NO]; 
+   
     // create view
     _view = [[vsg_MacOS_NSView alloc] initWithVsgWindow:this];
     [_view setWantsBestResolutionOpenGLSurface:_traits->hdpi];
--- a/src/vsg/platform/macos/MacOS_Window.mm
+++ b/src/vsg/platform/macos/MacOS_Window.mm
@@ -797,6 +797,14 @@ MacOS_Window::MacOS_Window(vsg::ref_ptr<vsg::WindowTraits> traits) :
     [_window setOpaque:YES];
     [_window setBackgroundColor:[NSColor whiteColor]];
 
+    // set the lower left corner window position as offset from the lower left corner of the screen
+    NSPoint pos; 
+    int xmax = [[NSScreen mainScreen] frame].size.width - [_window frame].size.width;
+    int ymax = [[NSScreen mainScreen] frame].size.height - [_window frame].size.height;
+    pos.x = std::clamp(traits->x,0,xmax);
+    pos.y = std::clamp(traits->y,0,ymax);
+    [_window setFrame:CGRectMake(pos.x, pos.y, [_window frame].size.width , [_window frame].size.height) display:YES];    
+
     // create view
     _view = [[vsg_MacOS_NSView alloc] initWithVsgWindow:this];
     [_view setWantsBestResolutionOpenGLSurface:_traits->hdpi];
--- a/src/vsg/platform/macos/MacOS_Window.mm
+++ b/src/vsg/platform/macos/MacOS_Window.mm
@@ -797,13 +797,7 @@ MacOS_Window::MacOS_Window(vsg::ref_ptr<vsg::WindowTraits> traits) :
     [_window setOpaque:YES];
     [_window setBackgroundColor:[NSColor whiteColor]];
 
-    // set the top left corner window position as offset from the top left corner of the screen
-    NSPoint pos; 
-    int xmax = [[NSScreen mainScreen] frame].size.width - [_window frame].size.width;
-    int ymax = [[NSScreen mainScreen] frame].size.height - [_window frame].size.height;
-    pos.x = std::clamp(traits->x,0,xmax);
-    pos.y = std::clamp(traits->y,0,ymax);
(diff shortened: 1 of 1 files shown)


rainergericke:
This PR is unique to MacOS.

src/vsg/platform/macos/MacOS_Window.mm
```diff
@@ -797,6 +797,14 @@ void createApplicationMenus(void)
     [_window setOpaque:YES];
     [_window setBackgroundColor:[NSColor whiteColor]];
 
+    // set the lower left corner window position as offset from the lower left corner of the screen
```
robertosfield, line 800: Window managers on Windows and unix assume that the origin is the top left, rather than lower left.

What is standard on the Mac?

rainergericke: Lower left.

rainergericke: Maybe it would be better to modify the Mac code to mimic Linux and Windows behavior to make it more logical to users with multiple platforms.

robertosfield: A top right origin is also consistent with Vulkan, while lower left is consistent with OpenGL.  My inclination is to standardized on top right.  Thought there is origin enum in include/vsg/core/Data.h:

    enum Origin : uint8_t
    {
        TOP_LEFT = 0,
        BOTTOM_LEFT = 2
    };

Perhaps this should be a setting in Traits?  Thought this is adding complexity.

There is also the issue of window managers ignoring the settings up you pass, this is something I'm seeing in Linux with Xcb.  I haven't spent enough time investigating how to resolve this to provide guidance on this.

For now I think just make things consistent across platforms and add a comment in WindowTraits to make it clear that the top left is used as an origin.

Cheers,
Robert.

rainergericke: Yes, I totally agree to you. Why should we make too many exceptions? I changed the relevant line, so the Mac should behave like Linux or Windows now. 

After some tests I also discovered a problem: 
_vsgskybox glTF-Sample-Models/2.0/FlightHelmet/glTF/FlightHelmet.gltf --skybox textures/skybox.dds_
crashes because the gltf model could not be loaded. After converting to vsgb it worked perfectly and faster:
_vsgskybox FlightHelmet.vsgb --skybox textures/skybox.dds_

The 2Cylinder model could also be loaded as gltf. Converting to vsgt/vsgb should be a workaround.

Best

Rainer

rainergericke: Could the original author of the Mac code help us?

vsg-dev: I will need to try out vsgskybox as suggested.  Have you tried the same thing on other platforms?

W.r.t the original author I haven't heard from @tomhog for a couple of years, he got a big client project that took all his time, perhaps he'll resurface again.

rainergericke: I just tried it on my windows machine:
_vsgskybox glTF-Sample-Models/2.0/FlightHelmet/glTF/FlightHelmet.gltf --skybox textures/skybox.dds_
is running well.

Running the same on the Mac leads to this error message:
2022-03-26 17:18:49.835 vsgskybox[6610:157690] -[__NSTaggedDate frame]: unrecognized selector sent to instance 0xc7a011b800000000
2022-03-26 17:18:49.843 vsgskybox[6610:157690] *** Terminating app due to uncaught exception 'NSInvalidArgumentException', reason: '-[__NSTaggedDate frame]: unrecognized selector sent to instance 0xc7a011b800000000'
*** First throw call stack:
(
	0   CoreFoundation                      0x000000019db32b08 __exceptionPreprocess + 240
	1   libobjc.A.dylib                     0x000000019d87de14 objc_exception_throw + 60
	2   CoreFoundation                      0x000000019dbc6a68 -[NSObject(NSObject) __retain_OA] + 0
	3   CoreFoundation                      0x000000019da91e14 ___forwarding___ + 1764
	4   CoreFoundation                      0x000000019da91670 _CF_forwarding_prep_0 + 96
	5   vsgskybox                           0x00000001004adfe4 -[vsg_MacOS_NSWindowDelegate handleFrameSizeChange] + 60
	6   CoreFoundation                      0x000000019daa6ddc __CFNOTIFICATIONCENTER_IS_CALLING_OUT_TO_AN_OBSERVER__ + 28
	7   CoreFoundation                      0x000000019db4b578 ___CFXRegistrationPost_block_invoke + 52
	8   CoreFoundation                      0x000000019db4b4e4 _CFXRegistrationPost + 456
	9   CoreFoundation                      0x000000019da75a74 _CFXNotificationPost + 732
	10  Foundation                          0x000000019e932e48 -[NSNotificationCenter postNotificationName:object:userInfo:] + 96
	11  AppKit                              0x00000001a0662560 -[NSWindow _setFrameCommon:display:fromServer:] + 2428
	12  vsgskybox                           0x00000001004af8ac _ZN8vsgMacOS12MacOS_WindowC2EN3vsg7ref_ptrINS1_12WindowTraitsEEE + 1048
	13  vsgskybox                           0x00000001004adce0 _ZN3vsg7InheritINS_6WindowEN8vsgMacOS12MacOS_WindowEE6createIJRNS_7ref_ptrINS_12WindowTraitsEEEEEENS6_IS3_EEDpOT_ + 76
	14  vsgskybox                           0x00000001004adc7c _ZN3vsg6Window6createENS_7ref_ptrINS_12WindowTraitsEEE + 28
	15  vsgskybox                           0x00000001003c6340 main + 5440
	16  dyld                                0x0000000101d79088 start + 516
)
libc++abi: terminating with uncaught exception of type NSException
zsh: abort      vsgskybox glTF-Sample-Models/2.0/FlightHelmet/glTF/FlightHelmet.gltf --skybox

Looks like my changes to 'MacOS_Window.mm' are suboptimal in some cases.

robertosfield:
What happens in vsgskybox if you comment out your changes? Does it still
crash?

Are their particular values that are causing the crash?

It's odd that windowing code would be sensitive to what you load, suggests
a memory error/uninitialized variable somewhere along the line.

rainergericke:
Yes it is strange. With the new code commented out I get this:
Bildschirmfoto 2022-03-26 um 18 52 56
That's what we want to get.

robertosfield:
On Sat, 26 Mar 2022 at 17:55, Rainer Gericke ***@***.***>
wrote:

> Yes it is strange. With the new code commented out I get this:
>

Looking at your changes, is it the line with:

   [_window setFrame:CGRectMak

that is causing the exception?

Could there be some order of initialization or threading sensitivity issue?

rainergericke:
Good news! You had the right idea. Shifting the code close to the end seems to resolve the problems. Additionally I use an alternative setting method for the position of the window. I am going to commit my last changes.

robertosfield:
On Thu, 24 Mar 2022 at 12:58, Rainer Gericke ***@***.***>
wrote:

> ***@***.**** commented on this pull request.
> ------------------------------
>
> In src/vsg/platform/macos/MacOS_Window.mm
> <https://github.com/vsg-dev/VulkanSceneGraph/pull/422#discussion_r834278483>
> :
>
> > @@ -797,6 +797,14 @@ void createApplicationMenus(void)
>      [_window setOpaque:YES];
>      [_window setBackgroundColor:[NSColor whiteColor]];
>
> +    // set the lower left corner window position as offset from the lower left corner of the screen
>
> Maybe it would be better to modify the Mac code to mimic Linux and Windows
> behavior to make it more logical to users with multiple platforms.
>
A top right origin is also consistent with Vulkan, while lower left is
consistent with OpenGL.  My inclination is to standardized on top right.
Thought there is origin enum in include/vsg/core/Data.h:

    enum Origin : uint8_t
    {
        TOP_LEFT = 0,
        BOTTOM_LEFT = 2
    };

Perhaps this should be a setting in Traits?  Thought this is adding
complexity.

There is also the issue of window managers ignoring the settings up you
pass, this is something I'm seeing in Linux with Xcb.  I haven't spent
enough time investigating how to resolve this to provide guidance on this.

Cheers,
Robet.

robertosfield:
For some reason github is reposting Issue replies that I made months ago, doing so without any prompting from me - like the one immediately above.  I have no clue why this is happening. A bug in github?

rolandhill:
Hi Robert,

I'm getting the same thing from other projects. Looks like it isn't
specific to VSG.

Cheers,
Roland

PR #426, opened 2022-04-11
Proposal to include UserData with LineSegmentIntersector intersections

rolandhill:
The data returned by LineSegmentIntersector Intersections is currently fairly limited to coordinates and the node path and there may be a need to include additional application specific information.

This PR introduces the UserData class that can be used as a base class for additional data to be attached to intersections and potentially other classes in the future. It is derived from vsg::Object so that ref_ptr can be used to manage memory. I considered just using a void *, but this is problematic with deleting objects.

LineSegmentIntersector has been modified to accept a UserData object.

This is working in my application, but I'm not sure if it's the best way to go about it.


robertosfield:
I don't a dedicated UserData class is necessary.  All vsg::Object subclasses have support for meta values/objects that you can assign to them.  Internally a vsg::Auxiliary object will be created for you when the Object::setValue(name, value) and Object::setValue(name, object) methods are called.  Essentially vsg::Aulixary works like a UserData container.

For LineSegmentIntersector all you should need to do to add custom data would be :

    lineSegmentIntersector->setValue("length", 10.0);
    lineSegmentIntersector->setObject("my_object", MyCustomObjec::create()t);

rolandhill:
Hi Robert,

Thanks for your comment. I'll convert my code to use the inbuilt
functionality.

Regards,
Roland

PR #427, opened 2022-04-12
Convert LineSegmentIntersector::Intersection to derive from vsg::Object

rolandhill:
This PR is the successor to PR 426.

The suggestion in PR 426 was to use the built in meta value functionality of vsg::Object to pass application specific data for intersections from LineSegmentIntersector. This isn't possible per intersection because each Intersection is held in a simple structure and not derived from vsg::Object. This PR converts LineSegmentIntersector::Intersection to be derived from vsg::Object set that the meta value function can be used.

Note that add() now returns a ref_ptr to the Intersection so that user values can be added afterward.

The main change for users is that Intersection members are now accessed via -> instead of dot.

The result can be used like this:

```
void PtInstanceGeom::accept(vsg::ConstVisitor& visitor) const
{
    if (auto v = dynamic_cast<LineIntersector*>(&visitor))
    {
            // We need to test if we have intersected a point
            for(int32_t index = 0; index < _pts->size(); index++)
            {
                vsg::dvec3 pt(_pts->at(index));
                if(v->intersects(vsg::dsphere(pt, 5.0)))
                {
                    auto is = v->add(pt, 0.0, {});
                    is->setValue("id", _ids[index]);
                }
            }
    }
    else
    {
        traverse(visitor);
    }
}
```

### DIFF
--- a/include/vsg/traversals/LineSegmentIntersector.h
+++ b/include/vsg/traversals/LineSegmentIntersector.h
@@ -33,25 +33,38 @@ namespace vsg
         LineSegmentIntersector(const dvec3& s, const dvec3& e, ref_ptr<ArrayState> initialArrayData = {});
         LineSegmentIntersector(const Camera& camera, int32_t x, int32_t y, ref_ptr<ArrayState> initialArrayData = {});
 
-        struct Intersection
+        class Intersection : public Inherit<Object, Intersection>
         {
-            dvec3 localIntersection;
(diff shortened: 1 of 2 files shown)


robertosfield:
I have merged and updated vsgExamples to build against these changes.  I have also merged into the StateComposer branch as I'm currently do further work in LineSegmentIntersector.

PR #431, opened 2022-04-19
Add vsg:: identifier for calls outside vsg namespace

rolandhill:
User defined values can only be defined within vsg namespace. This PR allows users to define their own values outside of vsg namespace.


robertosfield:
I merged but only consideration of applying this same change to the VSG_Array* macros realized that it sets the className to vsg::ClassName namespace which is inappropriate as they could be in a different namespace.

What would be required is a separate macro that you provide the namespace as well.  Or to simply change the main one to require a namespace and then update the use of this to keep things consistent.

PR #432, opened 2022-04-19
Phong lighting

rolandhill:
Phong shader was applying ambient light value directly without modifying it by the material alpha, leading to overly bright scenes.

The diffuse light calculations multiply by the light alpha instead of the (vertex colour * material) alpha. I think it should be the later - please check.

Note that I haven't looked at the other shaders as I'm only familiar with Phong.

### DIFF
--- a/src/vsg/utils/shaders/assimp_phong_frag.cpp
+++ b/src/vsg/utils/shaders/assimp_phong_frag.cpp
@@ -166,8 +166,8 @@ void main()
         // ambient lights
         for(int i = 0; i<numAmbientLights; ++i)
         {
-            vec4 ambient_color = lightData.values[index++];
-            color += ambient_color.rgb * ambient_color.a;
+            vec4 lightColor = lightData.values[index++];
+            color.rgb += ambientColor.rgb * lightColor.rgb * ambientColor.a;
         }
     }
 
--- a/src/vsg/utils/shaders/assimp_phong_frag.cpp
+++ b/src/vsg/utils/shaders/assimp_phong_frag.cpp
@@ -179,11 +179,11 @@ void main()
             vec4 lightColor = lightData.values[index++];
             vec3 direction = -lightData.values[index++].xyz;
             float diff = max(dot(direction, nd), 0.0);
-            color.rgb += (diffuseColor.rgb * lightColor.rgb) * (diff * lightColor.a);
+            color.rgb += (diffuseColor.rgb * lightColor.rgb) * (diff * diffuseColor.a);
             if (diff > 0.0)
             {
                 vec3 halfDir = normalize(direction + vd);
-                color.rgb += specularColor.rgb * (pow(max(dot(halfDir, nd), 0.0), shininess) * lightColor.a);
+                color.rgb += specularColor.rgb * (pow(max(dot(halfDir, nd), 0.0), shininess) * diffuseColor.a);
             }
         }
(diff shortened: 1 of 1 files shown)


robertosfield:
The .cpp shaders built into the core VSG are generated using vsgconv from the GLSL shaders in vsgExamples/data/shaders.  I'll have a look at your changes and the ones in vsgExamples and unify them,

robertosfield:
I'm currently reviewing the vsgExamples/data/shaders/assimp_phong.frag and comparing them to assimp_pbr.frag and see the phong variant currently ignores the material.ambientColor value, while is probably wrong, but also even that would probably be wrong as it would ignore any input from texturing on the object.  

Ignoring the texturing on the object would lead to a scene that uses the diffuse texture map to add colour and brightness to objects, areas that should be black shouldn't be flooded by ambient light.  

The pbr shader uses a baseColor that it uses the vertex colour and diffuse texture and then uses this in ambient colour contribution.

I'm thinking that phong shader should be closer to how the phr shader works and use the diffuse texture map as input.

A bit further down the line I want to see if I can unify the Phong and PBR materials, as well as the one use to support the OSG loader.  

Like you, I'm not the original author of the phong, obr or OSG->VSG shaders so I'm having read through and figure out the original intent, and whether than original intent is expressed in the shader and whether that intent is sound.

For today I think just fix the worse problems and they come back to it.

robertosfield:
I decided to check something temporary in but have ended up realizing that the non VSG_VIEW_LIGHT_DATA code path in assim_phong.frag relies upon the material ambient colour to encoded the ambient light level, with is rather mixing the light definition into the material definition.  

The VSG_VIEW_LIGHT_DATA code path can compute the ambient light contribution based on the vsg::AmbientLight values, so the material ambient property is more akin the ambient occlusion level. which while is at least a little more appropriate.  I don't think there is any easy solution to this - the old code path is broken and the only proper way to fix it is to fully embrace the new lighting functionality.

rolandhill:
I'm converting my code to fully embrace the new lighting, or at least trying to. I've got good results now using a diffuse material colour of 1.0, 1.0, 1.0 and ambient colour of 0, 0, 0, which doesn't wash out with white light. I'm wondering if this is all a work around to compensate for the ambient light not using the texture colour??

robertosfield:
In my version of assimp_phong.frag I've used the diffuseColor to modulate the ambientColor so it's a bit more realistic. The bit I'm tripping over is the non VSG_VIEW_LIGHT_DATA code path.

robertosfield:
I have checked in changes to the vsgExamples/data/shaders/assimp_phong.frag GLSL source and the built in assimp_phong_frag.cpp shaders to better represent the ambient contribution.   

It's not perfect but I think it's an improvement.

robertosfield:
Should add, changes at checked into the VSG and vsgExamples StateComposer branches.

rolandhill:
Thanks, Robert. It's getting late here and I probably won't be able to try it out until morning.

robertosfield:
On Wed, 20 Apr 2022 at 10:34, Roland Hill ***@***.***> wrote:

> The changes look much better!
>
> Good to hear.

I'm thinking that the shaders need a non LIGHT_DATA path where a fallback
of an ambient and directional light are encoded into the shader.  The aim
should be to unify the shader and make it clearer what the fallback is.

PR #436, opened 2022-04-29
Add EVSG_value to create values external to vsg namespace

rolandhill:
Update to PR 431 which was accepted then reverted.

Adds a new macro EVSG_value for use outside vsg namespace. This follows the pattern for VSG_type_name / EVSG_type_name.

### DIFF
--- a/include/vsg/core/Value.h
+++ b/include/vsg/core/Value.h
@@ -32,6 +32,12 @@ THE SOFTWARE IS PROVIDED "AS IS", WITHOUT WARRANTY OF ANY KIND, EXPRESS OR IMPLI
     template<>          \
     constexpr const char* type_name<N>() noexcept { return "vsg::" #N; }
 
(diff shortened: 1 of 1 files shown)


robertosfield:
To test out this PR I added a simple value struct in it's own namespace to the vsgvalues example, but couldn't coax your suggested EVSG_value to work correctly with the non vsg namespace.  This is what I ended up with:

    [vsgvalues mods](https://github.com/vsg-dev/vsgExamples/commit/a6e5d46bc9be8e0bcdc7e84ef6184a059eb15cda)

I think this is simple enough that the proposed macro isn't required.

robertosfield:
[Added comments and extra code to illustrate custom Value usage](https://github.com/vsg-dev/vsgExamples/commit/cc8cd3f725736d8e039b86249ac7f05e61471f92)

I'm going to close this PR now as it doesn't quite achieve the intended functionality, and setting up custom value is pretty straight forward without it,

PR #447, opened 2022-05-14
Added Cygwin and MinGW for ci

sbrkopac:
## Description

Added CI for both Cygwin and MinGW. Note that these have to be manually run rather than running during the typical CI process.

### DIFF
--- a/.github/workflows/win-cygwin-mingw.yml
+++ b/.github/workflows/win-cygwin-mingw.yml
@@ -1,0 +1,58 @@
+name: windows-mingw-cygwin
+on:
+  workflow_dispatch:
+jobs:
+  mingw:
+    runs-on: windows-latest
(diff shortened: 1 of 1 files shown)


robertosfield:
Thanks for the changes. I've merged, but how do you run Cygwin and MinGW targets manually?

sbrkopac:
If you click on Actions and then select the CI you want to run, it should be on the right.

robertosfield:
On Sun, 15 May 2022 at 18:16, Sam Brkopac ***@***.***> wrote:

> If you click on Actions and then select the CI you want to run, it should
> be on the right.
>
> It works, thanks for help.

Looks like MinGW is compiling, but Cywin is failing in the thread affinity
code, I'll investigate tomorrow.

PR #456, opened 2022-06-07
vsgvr - Allow inheritance from Instance and Device

geefr:
For the OpenXR integration I'm currently creating the Vulkan instance through the XR runtime - It's not the only approach, but what I've got for the moment, avoids the issue of selecting the device / parameters that the OpenXR runtime requires.

To do that though I need to be able to create the vsg/vk classes, while bypassing the normal creation logic.
For now I've inherited from a few classes, but needed to add default constructors in vsg to allow it.

Child classes are the following - Ideally they wouldn't exist but the base vsg classes would then need more changes:
* https://github.com/geefr/vsgvr/blob/openxr/vsgvr/src/vsgvr/OpenXRVkInstance.cpp
* https://github.com/geefr/vsgvr/blob/openxr/vsgvr/src/vsgvr/OpenXRVkPhysicalDevice.cpp
* https://github.com/geefr/vsgvr/blob/openxr/vsgvr/src/vsgvr/OpenXRVkDevice.cpp

I'm sure this isn't the best approach, any alternatives would be appreciated. Overall the integration is moving forward, so would be good to resolve this aspect (before thinking about a cleaner 'Viewer' implementation etc).

### DIFF
--- a/include/vsg/vk/Device.h
+++ b/include/vsg/vk/Device.h
@@ -35,6 +35,7 @@ namespace vsg
     class VSG_DECLSPEC Device : public Inherit<Object, Device>
     {
     public:
+        Device() = default;
         Device(PhysicalDevice* physicalDevice, const QueueSettings& queueSettings, const Names& layers, const Names& deviceExtensions, const DeviceFeatures* deviceFeatures = nullptr, AllocationCallbacks* allocator = nullptr);
 
         Instance* getInstance() { return _instance.get(); }
--- a/include/vsg/vk/Instance.h
+++ b/include/vsg/vk/Instance.h
@@ -33,6 +33,7 @@ namespace vsg
     class VSG_DECLSPEC Instance : public Inherit<Object, Instance>
     {
     public:
+        Instance() = default;
         Instance(const Names& instanceExtensions, const Names& layers, uint32_t vulkanApiVersion = VK_API_VERSION_1_0, AllocationCallbacks* allocator = nullptr);
 
         /// Vulkan apiVersion used when creating the VkInstaance


robertosfield:
I'm a bit dubious about these changes being the right way forward for you work.  Could you provide a link to example code that illustrates how you are wiring up the OpenVR provided Vulkan values to the VSG.

As a general statement, if we are to provide default constructors the Instance and Device to allow you to do the subclassing like you are currently trying then the way to do it would be it make the default constructors protected so that can't be called in normal usage.  This is crucial as we don't want normal users to attempt to create Instance and Device in invalid states.

geefr:
(Whoops, those were certainly meant to be protected..I missed something in cleanup)

Yes that's my thinking, there's probably a better approach for vsg, I'm not sure I can know what that is though.
The code path in this case is in https://github.com/geefr/vsgvr/blob/openxr/vsgvr/src/vsgvr/OpenXRGraphicsBindingVulkan2.cpp

With OpenXR there's 2 ways to initialise vulkan, I've chosen [XR_KHR_vulkan_enable2](https://www.khronos.org/registry/OpenXR/specs/1.0/man/html/XR_KHR_vulkan_enable2.html) for simplicity.

There is also XR_KHR_vulkan_enable, but a vsg change would also be needed in that case. The key restriction is that a specific physical device must be used due to the headset connection, and OpenXR requires various vulkan extensions and version restrictions.

The setup logic is roughly:
* Ask OpenXR what vulkan version / extensions are required, similar to regular vulkan setup
* Create vulkan instance, or ask OpenXR to create it
* Select the physical device that OpenXR requires, or query it from OpenXR
* Create vulkan device, including any requirements from OpenXR
* Set up the graphics binding, make swapchains, etc.

In vsg that equates to the ability to create a vsg::Instance, PhysicalDevice, and Device while being more specific about the construction:
* For vulkan_enable2 it would need to be a direct construction from vulkan types, A minimal change to allow inheritance, or a constructor that takes an existing Vulkan handle, and adopts it into the vsg type
* For vulkan_enable the original instance creation would stay within vsg (better for custom allocators?), but the device selection logic would need to take a known VkPhysicalDevice, rather than be based on application preferences.

robertosfield:
Thanks for the explanation and links.

I can't help but feel there's lots of leg work in their that really shouldn't be required as the VSG does 90% of what is required out of the box and is already customizable w.r.t Instance, PhysicalDevice & Device extensions.  

The VSG also has a robust Vulkan object lifetime scheme that the explicitly destroying of VkInstance/VkPhysicalDevice/VkDevice will break, so you are introducing a potential fragility that didn't previous exist.

It might be useful to run a simple OpenXR app  with Vulkan and simple model with the Vulkan API debug layer enabled and then capture the setup, first frame, then clean up.  This would tell you want the VSG would need to pass to Vulkan when setting up the Instance/PhysicalDevice/Device.

FYI, vsgExamples has a [vsgdeviceselection](https://github.com/vsg-dev/vsgExamples/blob/master/examples/viewer/vsgdeviceselection/vsgdeviceselection.cpp) example that illustrates query and selecting specific devices.  The [vsgmeshshader](https://github.com/vsg-dev/vsgExamples/blob/master/examples/rtx/vsgmeshshader/vsgmeshshader.cpp) example also illustrates querying and setting up for various Vulkan extensions.

If OpenEXR is adding it's own PhysicalDevice then it might be listed when you run:

    vsgdeviceselection --list

--

What is the status of OpenXR usage under Linux and emulating having the device?

geefr:
Spent the day looking into this, and you're right - Based on those examples the vulkan init can be done with the stock vsg classes, and needs much less code on the vsgvr side.

The stumbling block turned out to be SteamVR seemingly misbehaving, and a few errors in my code - Ended up not requesting the correct extensions, passing an invalid VkInstance to SteamVR, and triggering an opaque segfault.

So now things work fairly well, no changes to vsg needed, will close the PR, thanks for all the help here.
See [here](https://github.com/geefr/vsgvr/commit/cb984cf7aa18fb0cca96aecd8128cae435e2232a) for example usage..It's more involved than letting OpenXR init vulkan, but feels a lot more inline with the existing vsg approach.

Creating the instance through a Window does mean there's a brief flash of a desktop window, but should also make the desktop mirror-window simpler later on. Potentially there's vsg rework needed if the Oculus/Meta Quest headsets are supported eventually, as those probably don't have a desktop of any kind (Would fail when enumerating surface formats, openxr init would be fine as it's surface-less anyway).

PR #460, opened 2022-06-17
Link to Android log library

geefr:
## Description
Fixes link errors on Android, combined with vulkan compatibility changes should result in a functional library.

## Type of change

- Only impact is to Android builds

## How Has This Been Tested?

Tested so far by building shared libs, for android platform 24 (oldest supporting vulkan) and newer. Unable to test vsgExamples/Android at this time.

Build command: `cmake -DCMAKE_INSTALL_PREFIX=../../install_android/armeabi-v7a -DCMAKE_TOOLCHAIN_FILE=~/Android/Sdk/ndk/24.0.8215888/build/cmake/android.toolchain.cmake -DANDROID_ABI=armeabi-v7a -DANDROID_PLATFORM=android-24 -DBUILD_SHARED_LIBS=ON ../`

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -42,6 +42,7 @@ endif()
 
 if (ANDROID)
     find_library(AndroidLib android)
+    find_library(AndroidLogLib log)
     if(CMAKE_SYSTEM_VERSION GREATER 24)
         find_library(AndroidNativeWindowLib nativewindow)
     endif()
--- a/src/vsg/CMakeLists.txt
+++ b/src/vsg/CMakeLists.txt
@@ -275,9 +275,9 @@ if (ANDROID)
     set(SOURCES ${SOURCES} platform/android/Android_Window.cpp)
 
     if(CMAKE_SYSTEM_VERSION GREATER 24)
-        set(LIBRARIES ${LIBRARIES} PRIVATE ${AndroidLib} PRIVATE ${AndroidNativeWindowLib})
+        set(LIBRARIES ${LIBRARIES} PRIVATE ${AndroidLib} PRIVATE ${AndroidLogLib} PRIVATE ${AndroidNativeWindowLib})
     else()
-        set(LIBRARIES ${LIBRARIES} PRIVATE ${AndroidLib})
+        set(LIBRARIES ${LIBRARIES} PRIVATE ${AndroidLib} PRIVATE ${AndroidLogLib})
     endif()
 
 elseif (WIN32)


vsg-dev:
What is the log lib required for?

geefr:
AndroidWindow.cpp LOG macro, forwards to android_log_print (the device console)

vsg-dev:
On Sat, 18 Jun 2022 at 09:48, Gareth Francis ***@***.***>
wrote:

> AndroidWindow.cpp LOG macro, forwards to android_log_print (the device
> console)
> Message ID: ***@***.***>

This like a replacement for std::cout/std::cerr, there are other places in
the VSG that still has std::cout/std::cerr.

What happens with the android_log_print vs std::cout/std::cerr?

geefr:
On Android there isn't really a cout/cerr. If you set a flag they can be forwarded to the Android device log, but by default they're passed to /dev/null.

The android log itself is meant for development/debugging, usually you connect to the phone and monitor via `adb logcat`. Without the adb dev tool or other android apps normal users wouldn't be viewing the log. Showing messages to the user is done through the UI of an app, or could be done via the short 'toast' messages on screen.

For now, linking to the Android log library is needed - If vsg doesn't the application would anyway (but vsg calls android_log, so does need to link).

In the future it may be good to have an abstracted logging system that forwards to things like Android log, spdlog, allows an application level callback to be used, etc.

robertosfield:
Thanks for the insights @geefr, this is really helpful.

My current feeling is the right thing to do is to do an experimental with a vsg::log API that use std::cout/cerr by default can be customized by applications, such the Android example in vsgExamples.  I will have a think about this today/tomorrow.

In the meantime I'll just leave this PR pending as if the experiment just looks it won't easily converge to a release ready API then we can either merge this PR or just make the LOG macro map to null itself.

robertosfield:
I have now completed the backwards compatibility work and have merged it with VSG master: #461

I will now look at the possibility of creating a logging capability to address the Android log message issue.  I don't know how it'll turn out but will post a link to the log dev branch if/when it takes shapes.

robertosfield:
As a first step I have created a logging branch of vsgExamples:

   [https://github.com/vsg-dev/vsgExamples/tree/logging](https://github.com/vsg-dev/vsgExamples/tree/logging)

With an vsglog as a testbed for trying out different API approaches, I'm starting super simple with a pure virtual Logger base class, but this is what I've got so far:

   [https://github.com/vsg-dev/vsgExamples/blob/logging/examples/core/vsglog/vsglog.cpp](https://github.com/vsg-dev/vsgExamples/blob/logging/examples/core/vsglog/vsglog.cpp)

geefr:
Yeah looks like a fairly simple/sensible approach.
For Android specifically I think there's a few cases for applications:
* No logging from vsg (default, with cout/cerr > /dev/null)
* A native (C++) application, which can provide a custom logger
* A java application - Regardless of any future java wrappers, would have some C++ layer, which again can provide a custom logger

Not sure what difference it would make, but in the default logger the mutex could be avoided, if messages are composed before printing.

(There's also std::format, becoming more popular. But would require either C++20 or a fallback to libfmt)

vsg-dev:
I think the mutex will be required even with std::cout being thread safe as it's only the characters that are atomically handled, so two whole strings written to different threads would ended up visually interleaved.  While it wouldn't crash it wouldn't be readable, and readability is the reason for logging.  

This was an issue that we never resolved in the OSG project, so if we can I'd rather solve it this time, we potentially could port solutions we come up with  to the OSG.

W.r.t std::format from C++20, I don't want to add any further constraints than C++17 if we can avoid it.  This also applies to adding external dependencies, which is the whole point of my wandering down the vsg::log() possibility.

robertosfield:
I have converted the Android_Window.cpp to use the new vsg::debug(..) API from the logging branch.  The commit with these changes is: f9ebe3d4eb9250cb0a65e2b97d11dbac39ca31db

Could you test the [logging](https://github.com/vsg-dev/VulkanSceneGraph/tree/logging) branch out and let me know if it's compiling under Android without the need for the Android logging library?

geefr:
Okay looks like the logging changes are good, ignoring ongoing android issues.

Built for android api 32 (latest): Compiled and linked as shared libs, no issues.
api 24 (oldest with vulkan): Link errors on the recent vulkan compat changes - undefined reference to vkGetPhysicalDeviceProperties2.

I might have time over the weekend to dig into that, but either way yes the logging changes look good to me, should allow an android logger to be hooked up well.

robertosfield:
Thanks for the testing.

What is the Vulkan version in the Android API 24?

geefr:
Based on what CMake gets and vulkan_core.h - 1.2.158
Each android api version does have its own libvulkan.so however, and it looks like both api 24 and api 32 share the same set of vulkan headers.

vulkan_core.h has the declaration for vkGetPhysicalDeviceProperties2, but in the api 24 version of libvulkan.so it's missing.
That function is present from api version 28 onwards..

As the vulkan headers are the same I assume there's no way to automatically detect that, but the android api version could be used in cmake to override the vulkan version selection? Alternatively the minimum android api could be set to 28 or so, that's Android 9, which would cover 80% of devices based on google's docs. Keeping the minimum at api 24 (Android 7) would cover 95% of devices.

robertosfield:
What error are you seeing w.r.t vkGetPhysicalDeviceProperties2?

Is it compile, link error or runtime error?

geefr:
Just a link error, as it's not present in libvulkan.so when building for api 24
```
ld: error: undefined symbol: vkGetPhysicalDeviceProperties2
>>> referenced by PhysicalDevice.h:62 (/home/gareth/dev/VulkanSceneGraph/include/vsg/vk/PhysicalDevice.h:62)
>>>               CMakeFiles/vsg.dir/raytracing/RayTracingPipeline.cpp.o:(VkPhysicalDeviceRayTracingPipelinePropertiesKHR vsg::PhysicalDevice::getProperties<VkPhysicalDeviceRayTracingPipelinePropertiesKHR, (VkStructureType)1000347001>() const)
>>> did you mean: vkGetPhysicalDeviceProperties
>>> defined in: /home/gareth/Android/Sdk/ndk/24.0.8215888/toolchains/llvm/prebuilt/linux-x86_64/sysroot/usr/lib/arm-linux-androideabi/24/libvulkan.so
clang++: error: linker command failed with exit code 1 (use -v to see invocation)
```

vsg-dev:
On Thu, 23 Jun 2022 at 18:33, Gareth Francis ***@***.***>
wrote:

> Just a link error, as it's not present in libvulkan.so when building for
> api 24
>
> ld: error: undefined symbol: vkGetPhysicalDeviceProperties2
> >>> referenced by PhysicalDevice.h:62 (/home/gareth/dev/VulkanSceneGraph/include/vsg/vk/PhysicalDevice.h:62)
> >>>               CMakeFiles/vsg.dir/raytracing/RayTracingPipeline.cpp.o:(VkPhysicalDeviceRayTracingPipelinePropertiesKHR vsg::PhysicalDevice::getProperties<VkPhysicalDeviceRayTracingPipelinePropertiesKHR, (VkStructureType)1000347001>() const)
> >>> did you mean: vkGetPhysicalDeviceProperties
> >>> defined in: /home/gareth/Android/Sdk/ndk/24.0.8215888/toolchains/llvm/prebuilt/linux-x86_64/sysroot/usr/lib/arm-linux-androideabi/24/libvulkan.so
>
>
It looks like it's a 1.1 feature so I'm surprised it's not available.  To
address the problem we'll need to put it into vsg::Extensions.


https://www.khronos.org/registry/vulkan/specs/1.3-extensions/man/html/vkGetPhysicalDeviceProperties2.html

I'll look into this tomorrow.Message ID:
***@***.***>

>

geefr:
Ah right, I guess android api 28 was when vulkan 1.1 started to be supported then..The Android docs are a little vague on those points (Or I missed the obvious bits).

Here's a basic CI workflow for compiling vsg, may help streamline the compilation checks a bit:
https://github.com/geefr/VulkanSceneGraph/blob/android-ci/.github/workflows/ci-android.yml

We'd need to decide on which architectures / api versions should be checked here, as there's a lot of combinations there.
There's also the possibility of different NDK/toolchain versions, which over time may have their own version-specific behaviours.

robertosfield:
To resolve the issue with vkGetPhysicalDeviceProperties2 I've changed the PhysicalDevice class to setup function pointers in the constructor that are then used late in the PhysicalDevice::getFeatures/getPropoerties(..) methods.  I have just merged these changes with VSG master from the [getProcAddr](https://github.com/vsg-dev/VulkanSceneGraph/pulls?q=is%3Apr+is%3Aclosed) branch that I created today.  This branch also adds Instance::getProcAddr(...) and Device::getProcAddr(..) convenience functions to streamline extension setup.

I have also merged these changes with the [logging](https://github.com/vsg-dev/VulkanSceneGraph/tree/logging) branch.

robertosfield:
I have now merged the vsg::Logger work with master #462.  This along with the change to vsg::PhysicalDevice should address the build issues.  

@geefr. The ci-android,yml looks to be a useful addition to our automated build testing, could you generated a PR for it?

robertosfield:
I will now close this PR as the various changes have been addressed.

PR #468, opened 2022-06-26
Android CI

geefr:
Basic workflow for Android - Just compilation and linking.
Build selection is currently the 4 main architectures, api 24, and latest.
(24 being the oldest that has vulkan support)

### DIFF
--- a/.github/workflows/ci-android.yml
+++ b/.github/workflows/ci-android.yml
@@ -11,7 +11,7 @@ jobs:
     strategy:
       matrix:
         os: [ubuntu-latest]
-        android-platform: [24, 32]
+        android-platform: [24, latest]
         android-abi: [armeabi-v7a, arm64-v8a, x86, x86_64]
         build-shared: [ON]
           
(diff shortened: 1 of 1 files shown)


robertosfield:
I've been wondering whether it's really need to have 4 build combinations for Android. I'd prefer to avoid to high CPU load/energy use associated with our automated builds.  My inclination would be to test arm and different NDK versions, and x86* probably isn't required.

Thoughts?

geefr:
Sure that's a sensible approach.

Almost all Android devices are Arm now, with a few x86/64 devices.
Arguably the x86 Android compilation is almost equivalent to desktop Linux, often it's only used for the device simulator, which is much faster then running the Arm code.
I think most physical x86 devices also include an emulation layer, but it varies (I haven't checked in a while).

The only possibly concern for vsg is that the native toolchains can have version and architecture specific bugs. Those could be handled if/when they happen though, as anyone building an actual application would need to handle such things regardless.

robertosfield:
Thanks for the feedback.   I've dropped the x86 and x86_64 from the Android
build.  Potentially we can add builds that could be run manually like with
have for Cygwin.

PR #482, opened 2022-07-07
Fix Windows build with UNICODE defined

WSSDude420:
## Description

This PR fixes build when UNICODE or _UNICODE is defined during compilation on Windows platform. 
It also alleviates issue of using macros which is generally discouraged nowadays
> New applications should always call the Unicode versions.
Source: https://docs.microsoft.com/en-us/windows/win32/learnwin32/working-with-strings#unicode-and-ansi-functions

## Type of change

### DIFF
--- a/include/vsg/platform/win32/Win32_Window.h
+++ b/include/vsg/platform/win32/Win32_Window.h
@@ -39,7 +39,7 @@ namespace vsgWin32
             uint16_t modifierMask = 0;
 
             //bool rightSide = (lParam & 0x01000000) != 0;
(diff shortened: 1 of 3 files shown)


robertosfield:
What happens when you don't build with UNICODE defined with these changes?

robertosfield:
Could you please separate the function name changes from the size changes into two separate commits - conflating two types of changes in one PR makes it hard to manage, track changes and revert if required.

WSSDude420:
Ok np, will rework PR with TCHAR instead. There was no need to close this I believe, I could have just change it and commit it to the branch but I'll make another one then.

PR #483, opened 2022-07-07
Fix inclusion in another CMake project

WSSDude:
## Description

When including VSG directly as another CMake project via using add_subdirectory, project fails to configure as paths are all wrong. Also, there was forgotten check for Clang under MSVC environment which resulted in tons of warnings from both, compiler (due to unrecognized flags) and from compilation itself (warnings that are of no use, like C++98compat).

It also renames uninstall target to uninstall-vsg to not clash with SDL2 uninstall target (otherwise, it won't configure due to already defined target)

## Type of change


WSSDude420:
Something messed up when I tried to clean it up a bit I guess, lemme fix it real quick.

robertosfield:
I'm not clear on what the motivation is for the latest changes to your original PR.  As a general guide I'd rather keep things consistent with normal CMake conventions and only diverge when there is a specific reason to do so.

WSSDude420:
I reverted that change to variable. Build should be fixed now.

robertosfield:
Your initial commit was from CMAKE_SOURCE_DIR to CMAKE_CURRENT_SOURCE_DIR, now you've introduced a VSG specific var for the same purpose.  I don't know why you've done this.  I think it would be better to have two separate PR's.  What I may be best to do is just start from scratch with your MyFirstVsgApplication variant and create a VSG, vsgXchange etc. branches to see what is needed.

The change to the install name is also a bit of concern, I understand that you came across a conflict, bit if this was a conflict then what happens with all the other 3rd party projects that just happen to define their own uninstall, are you going to try and get all 3rd party projects to make changes?  The vsgMacros also are used in other VSG proejcts beyond the VSG itself so these targets could end up colliding themselves.  As this another whole potential can of worms this change really should be kept separate.

WSSDude420:
Ok, can do the split. Will leave only the path changes here.
In regards to that MSVC Clang, should that be left in as well or should that also be separated?

WSSDude420:
Did the extra change you requested for now, gotta leave for a bit. If you also want to get rid of that MSVC Clang, just let me know.
I also modified the sample that can be found [here](https://github.com/WSSDude420/vsg-test-submodule) (which I posted also in #450) to do the renaming there.

OpenSiege:
@robertosfield Can you please approve the workflows for this.

sbrkopac:
@robertosfield Can you please approve the workflows for this.

robertosfield:
Thanks, I just did a quick review, that's enough for now.  I will probably merge as a branch and adapt/experiment with the changes in this branch and decide what is appropriate thing to do.

The MSVC Clang issue is separate so I won't merge it with the other changes, though as a branch this is probably fine.  I don't know the background info in the MSCV Clang so will need to understand that and what the best solution is likely to be before merging, it's certainly not something I'd merge with the other changes though, even if I agree with the change.  Changes like this may well fix the build for one person only to break it for another so it crucial that they can be traced and backed out if necessary without causing lots of collateral damage.

This type of incremental approach is what I'd recommend you use for your own projects.  Modern version control systems like git give us the ability to conveniently manage things at a pretty fine grained way, collapsing unrelated changes into single code drops is bad for project management long term.

WSSDude420:
Yeah, I just didnt want to create PR with one line change, felt kinda wrong. But sure, did not leave yet, can quickly get rid of it from here, gimme sec.

WSSDude420:
For future PRs here, I'll keep in mind what you wrote :)

WSSDude420:
I'll try to work on the Windows issue and this Clang one tomorrow, will try to make separate PRs and describe them more thoroughly so we can maybe make more proper fixes for each separate issue that way.

In short though as you mentioned it, problem is that in MSVC environment, Clang is used via clang-cl driver which emulates MSVC native cl compiler. It accepts exactly same flags as it.

"GNU" flags used with regular Clang compiler do not work with it.

robertosfield:
Thanks.

robertosfield:
To help myself and others test out this PR I've merged it as is as the branch [WSSDude420-Fix_inclusion_in_other_CMake_project](https://github.com/vsg-dev/VulkanSceneGraph/tree/WSSDude420-Fix_inclusion_in_other_CMake_project).

I'll close this PR, and we can generate a PR from the now local branch.

rhabacker:
As a general rule for including a cmake project in another cmake project, I remembered a hint at the [dbus](https://gitlab.freedesktop.org/dbus/dbus/-/commit/fac146f44d235a1c8c810de09c9582a0dd28b406) project to use `PROJECT_SOURCE_DIR` instead of `CMAKE_SOURCE_DIR`, since the latter points to the top level cmake project , while PROJECT_SOURCE_DIR points to the directory where the current `PROJECT()` is used.

robertosfield:
Over the weekend I'll do some more work on the submodule example - adding
other modules like vsgXchange and vsgImGui. vsgImGui itself pulls in ImGui
as a submodule.

We can see if we can get others in the community to test it out and report
back their experience to see what patterns emerge.

PR #485, opened 2022-07-09
Fix Windows build with UNICODE defined

WSSDude420:
## Description

Fixes compilation on WIndows when UNICODE is defined. In this case, macros resolve to functions accepting wchar_t.
Changes made should be hopefully least invasive way to fix this issue.

## Type of change

## How Has This Been Tested?

Defined additional compile definition `UNICODE` on Windows platform.
Master does not compile while PR does.


WSSDude420:
Reopening this, hopefully this should be better now.

sbrkopac:
This looks much better.

PR #498, opened 2022-08-11
[ADDED] overridePipelineStates to vsg::View

siystar:
# Pull Request Template

## Description

Added overridePipelineStates to vsg::View for configuring alternate rendering pathways like early-Z pass, shadow mapping, deferred shading etc.

## Type of change

## How Has This Been Tested?

Shadow mapping class in our application

## Checklist:

### DIFF
--- a/include/vsg/viewer/View.h
+++ b/include/vsg/viewer/View.h
@@ -63,6 +63,9 @@ namespace vsg
         /// view dependent state used for positional state like lighting, texgen and clipping
         ref_ptr<ViewDependentState> viewDependentState;
 
+        /// override states for customization of graphics pipelines for this view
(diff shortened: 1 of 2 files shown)


robertosfield:
On a first review, I wonder if a dedicated vsg::Group subclass that has the default and override settings for PipelineStates.  This would mean you need to add an extra node for your usage case, but would allow us to use it more generally.  Though the problem would be multi-buffering of the pipeline as it currently utilizes viewID.... so perhaps assigning it to View would be best...

I'll ponder more on this.  If you can point me at code that illustrates the usage case that needs this then this will help me when pondering the pros and cons of different approaches.  Thanks.

PR #552, opened 2022-11-01
spellcheck pass [no ci]

sbrkopac:
## Description

Spellcheck pass

@robertosfield I have left a couple of TODO items as I wasn't sure if they needed to be fixed up


include/vsg/commands/DrawIndexed.h
```diff
@@ -19,7 +19,7 @@ THE SOFTWARE IS PROVIDED "AS IS", WITHOUT WARRANTY OF ANY KIND, EXPRESS OR IMPLI
 namespace vsg
 {
 
-    /// DrawIndexed command encapsulates vkCmdDrawIndexed call and associatesd settings
+    /// DrawIndexed command encapsulates vkCmdDrawIndexed call and associates settings
```
robertosfield, line 22: ? No, associated is correct spelling in this context.

robertosfield:
With a few updates now looks good.

The bizarre thing is, I've been consistently used codespell to try and catch typos, and it's caught quite a few before I checked in, but none of these ones??!  Will need to look at updating codespell.  What tool are you using?

sbrkopac:
I use a [tool from Visual Studio marketplace](https://marketplace.visualstudio.com/items?itemName=EWoodruff.VisualStudioSpellCheckerVS2022andLater). Seems to work pretty well now that I have it setup with the proper added words and run through this process a few times.

If you're using codespell it might be worthwhile to setup an ignore file and a few other things right into the repo for future use. I can give it a try on a previous commit and see what I get.

robertosfield:
I will have a look at codespell setup tomorrow to see if I can glean why it's not working well as I need it to.

PR #563, opened 2022-11-08
Restructured the handling the vsgMacros.cmake and vsg_glslangConfig.cmake so they are both installed in lib/cmake/vsg and included by vsgConfig.cmake.

robertosfield:


robertosfield:
@rhabacker I have gone ahead and merged the vsgMacros,cmake changes to VSG, vsgXchange, vsgImGui, vsgQt and vsgExample masters.  Now it's time to cross fingers and toes that everything works fine.

PR #585, opened 2022-11-15
Package workaround for egor-tensin/setup-cygwin#9

### LINKED ISSUE #534: Github workflow Cygwin build failing
I have just manually run the windows-mingw-cygwin action and MinGW succeeds, but Cywin fails:

[https://github.com/vsg-dev/VulkanSceneGraph/actions/runs/3337513716](https://github.com/vsg-dev/VulkanSceneGraph/actions/runs/3337513716)

@sbrkopac It looks like it's failing due to the setup of Cywin rather than the build of the VSG itself.  As I don't have any expertise in this area I don't know where to start with resolving the problem so would appreciate help.

sbrkopac:
Fixes #534


robertosfield:
Wahooo.  Thanks for the effort resolving this.

PR #620, opened 2022-11-30
Fixed defines matching

asafran:
## Description

Changed the algorithm for matching defines and ArrayState.

Direct defines sets comparison leads to errors when using additional defines. For example.. when using the texture, instancing and displacement map at the same time, the set of defines will be: { "VSG_DIFFUSE_MAP", "VSG_DISPLACEMENT_MAP", "VSG_INSTANCE_POSITIONS" }, and the search for the desired ArrayState will fail. In the StateGroups generated using vsg::Builder, the ArrayState is not assigned when using an image and an offset map at the same time.

Fixes # (issue)

Fixed null state of the prototypeArrayState array when using an displacement map and a diffuse map (texture image) at the same stategroup.

The definesArrayStates needs to be sorted from more defines to fewer.

## Type of change

## How Has This Been Tested?

Create mesh using vsgbuilder from the vsgExamples, specify image and displacement map.
Test ArrayState handling using vsgintersection example with generated mesh.

### DIFF
--- a/src/vsg/utils/ShaderSet.cpp
+++ b/src/vsg/utils/ShaderSet.cpp
@@ -150,7 +150,8 @@ ref_ptr<ArrayState> ShaderSet::getSuitableArrayState(const std::set<std::string>
 {
     for (auto& definesArrayState : definesArrayStates)
     {
-        if (definesArrayState.defines == defines) return definesArrayState.arrayState;
+        bool includes = std::includes(defines.begin(), defines.end(), definesArrayState.defines.begin(), definesArrayState.defines.end());
+        if (includes) return definesArrayState.arrayState;
     }
 
     return {};


robertosfield:
I would like to investigate the bug.  Is there an easy way to create it, such as with existing VSG examples?

robertosfield:
@asafran I'm closing this PR as it will even more problematic than the original code and will match inappropriately.

From my review of your description and the original code, and the PR what it looks to me is that there are defines that affect the vertex shader and require ArrayState definition and others that have no affect, such as the ones that affect the fragment shader.  The fix will require filtering out of the defines that don't affect the ArrayState selection.  I haven't figured out how to do this yet, but will look at it today.

Could you help with refining test I can use to make sure changes address the problem you came across.

I have created a Issue #622 for this problem so this will be the place to add any further details for recreating the bug.  Thanks.

Issue #632, opened 2022-12-07
Crash (EXC_BAD_ACCESS) in vsg::CompileManager::compile() while running vsgpagedlod


artemkulyk:
**Describe the bug**
Sometimes VulkanSceneGraph crashes in vsg::CompileManager::compile() while running "vsgpagedlod --osm". Possibly, related to some race condition.

**To Reproduce**
1. Compile and install VulkanSceneGraph, vsgXchange, vsgImGui with standard flags.
2. Compile vsgExamples with standard flags.
3. Run "vsgpagedlod --osm". I don't sure that flag --osm is required, but I reproduced the bug with it several times.
4. Press "1", "2", "3" ... "9" while crash is reproduced.
5. If crash is not reproduced, restart the app and try again.

**Expected behavior**
No crashing

**Screenshots**
<img width="1440" alt="Screenshot 2022-12-07 at 07 54 18" src="https://user-images.githubusercontent.com/88086187/206109711-41337e0f-9c99-4c67-b361-8b99f6a4f6a3.png">

**Desktop (please complete the following information):**
 - OS: Mac OS Ventura
 - HW: Apple Silicon (M1)
 - Compiler: Apple clang version 14.0.0 (clang-1400.0.29.202)
 - Target: arm64-apple-darwin22.1.0
 - MoltenVK Vulkan SDK 1.3.231
 - VulkanSceneGraph master
 - vsgExamples master

**Additional context**
Core dumps are attached below.
[corefilemodified_memory.zip](https://github.com/vsg-dev/VulkanSceneGraph/files/10173497/corefilemodified_memory.zip)
[corefilestack.zip](https://github.com/vsg-dev/VulkanSceneGraph/files/10173498/corefilestack.zip)

rolandhill:
Hi Artemkulyk & Robert,

I get a similar occasional crash in my own application that pages terrain
tiles using DatabasePager. Refer to this report:
https://github.com/vsg-dev/VulkanSceneGraph/issues/490

Regards,
Roland

artemkulyk:
Hi @rolandhill,

Yes, it seems the same issue.

Best regards,
Artem Kulyk

rolandhill:
Hi Artem,

I just posted draft PR #633 https://github.com/vsg-dev/VulkanSceneGraph/pull/633.

I've had this for a while without being able to reliably recreate the crash. Maybe give that PR a try and see if it helps. There's more of an explanation at that link.

Regards,
[...]

Issue #490, opened 2022-07-25
Crash in CompileManager due to null context


rolandhill:
Hi Robert,

I got a crash in CompileManager due to a NULL context. CompileManager was being called from DatabasePager. Please refer to the annotated code below.

During compile(), "ref_ptr<View> view = context->view;" is called with context = 0x00. Inspecting compileTraversal shows that there is only 1 context and it's not NULL. The only thing I can think of is that another thread is manipulating the contexts in compileTraversal at the same time - is this a possibility? It all worked perfectly paging tiles for several minutes before the crash.

Sorry, I can't put together example code to demonstrate this. I'm using the stock DatabasePager implemented via a merge operation as per your examples.

Regards,
Roland

CompileResult CompileManager::compile(ref_ptr<Object> object, ContextSelectionFunction contextSelection)
{
    CollectResourceRequirements collectRequirements;
    object->accept(collectRequirements);

    auto& requirements = collectRequirements.requirements;
    auto& binStack = requirements.binStack;

    CompileResult result;
    result.maxSlot = requirements.maxSlot;
    result.containsPagedLOD = requirements.containsPagedLOD;
    result.views = requirements.views;

    auto compileTraversal = compileTraversals->take_when_available();

    // if no CompileTraversals are available abort compile
    if (!compileTraversal) return result;

    std::list<ref_ptr<Context>> contexts;
    if (contextSelection)
    {
        for (auto& context : compileTraversal->contexts)
        {
            if (contextSelection(*context)) contexts.push_back(context);
        }
    }
    else
    {
        contexts = compileTraversal->contexts;
    }

    compileTraversal->contexts.swap(contexts);

    for (auto& context : compileTraversal->contexts) **<------ at time of crash compileTraversal contains 1 context, which is not NULL**
    {
        ref_ptr<View> view = context->view;                    **<------------ Context 0x00**
        if (view && !binStack.empty())
        {
            if (auto itr = result.views.find(view.get()); itr == result.views.end())
            {
                result.views[view] = binStack.top();
            }
        }

        context->reserve(requirements);
    }

    object->accept(*compileTraversal);

    //debug("Finished compile traversal ", object);

    compileTraversal->record(); // records and submits to queue
    compileTraversal->waitForCompletion();

    debug("Finished waiting for compile ", object);

    compileTraversals->add(compileTraversal);

    compileTraversal->contexts.swap(contexts);

    result.result = VK_SUCCESS;
    return result;
}

from

[...]

PR #633, opened 2022-12-07
Request discard

rolandhill:
Regarding issue https://github.com/vsg-dev/VulkanSceneGraph/issues/490 and https://github.com/vsg-dev/VulkanSceneGraph/issues/632.

I'm wondering if a tile/node is being discarded while DatabasePager is trying to compile it?? Several crashes pointed in this direction, but I haven't got the traces to show it.

I've made the changes in this PR and don't think I've seen the same crash since, however it is occasional and I found it difficult to reliably repeat. One change is to reinstate a commented out lock. The other is to add a similar lock.

If you have a reliable way to recreate the crash then it's worth a try - just a suggestion.

Roland

### DIFF
--- a/src/vsg/app/CompileManager.cpp
+++ b/src/vsg/app/CompileManager.cpp
@@ -133,13 +133,13 @@ CompileResult CompileManager::compile(ref_ptr<Object> object, ContextSelectionFu
         {
             if (contextSelection(*context)) contexts.push_back(context);
         }
+
+        compileTraversal->contexts.swap(contexts);
     }
     else
     {
         contexts = compileTraversal->contexts;
     }
-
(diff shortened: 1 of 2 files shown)


robertosfield:
Thanks.  I'll do a proper review of the topic tomorrow.

robertosfield:
I am thinking that this PR is no longer required as a fix is now in VSG master.  If it does turn out that the fix in VSG master is insufficient then we can look at reproducing any remaining issues and then coming up with a fix, potentially cherry-picking for this PR.

rolandhill:
No problem, Robert. We'll see if the problem is fixed over time.

Roland

PR #642, opened 2022-12-10
Fix spontaneous crash during starting app in fullscreen mode on macOS.

### LINKED ISSUE #640: App starts in windowed mode even if WindowTraits::fullscreen was set to true on macOS
**Describe the bug**
App starts in windowed mode even if WindowTraits::fullscreen was set to true on macOS

**To Reproduce**
Try to create fullscreen app on macOS. App starts in windowed mode.

**Expected behavior**
App starts in fullscreen mode if WindowTraits::fullscreen set to true.

- OS: Mac OS Ventura
- HW: Apple Silicon (M1)
- Compiler: Apple clang version 14.0.0 (clang-1400.0.29.202)
- Target: arm64-apple-darwin22.1.0
- MoltenVK Vulkan SDK 1.3.231
- VulkanSceneGraph master

artemkulyk:
## Description
Switch to fullscreen mode was done before view is initialised. As a result, app can spontaneously crash on start in method handleFrameSizeChange.

Fixes #640

## Type of change

## How Has This Been Tested?

Tested on local machine.

**Test Configuration**:
* OS: Mac OS Ventura
* HW: Apple Silicon (M1)
* Compiler: Apple clang version 14.0.0 (clang-1400.0.29.202)
* Target: arm64-apple-darwin22.1.0
* MoltenVK Vulkan SDK 1.3.231
* VulkanSceneGraph master

## Checklist:

### DIFF
--- a/src/vsg/platform/macos/MacOS_Window.mm
+++ b/src/vsg/platform/macos/MacOS_Window.mm
@@ -796,10 +796,6 @@ MacOS_Window::MacOS_Window(vsg::ref_ptr<vsg::WindowTraits> traits) :
     [_window setRestorable:YES];
     [_window setOpaque:YES];
     [_window setBackgroundColor:[NSColor whiteColor]];
-    if (traits->fullscreen)
-    {
-        [_window toggleFullScreen:nil];
-    }
 
     
     // create view
@@ -812,6 +808,14 @@ MacOS_Window::MacOS_Window(vsg::ref_ptr<vsg::WindowTraits> traits) :
     [_window setContentView:_view];
     _window.initialFirstResponder = _view;
     [_window makeFirstResponder:_view];
+
+    if (traits->fullscreen)
+    {
+        NSRect screenFrame = [[NSScreen mainScreen] frame];
+        [_window setBackgroundColor: NSColor.whiteColor];
+        [_window setCollectionBehavior:NSWindowCollectionBehaviorFullScreenPrimary];
+        [_window toggleFullScreen:NSApp.delegate];
+    }
 
     auto devicePixelScale = _traits->hdpi ? [_window backingScaleFactor] : 1.0f;
     [_metalLayer setContentsScale:devicePixelScale];
--- a/src/vsg/platform/macos/MacOS_Window.mm
+++ b/src/vsg/platform/macos/MacOS_Window.mm
@@ -812,7 +812,6 @@ MacOS_Window::MacOS_Window(vsg::ref_ptr<vsg::WindowTraits> traits) :
     if (traits->fullscreen)
     {
         NSRect screenFrame = [[NSScreen mainScreen] frame];
-        [_window setBackgroundColor: NSColor.whiteColor];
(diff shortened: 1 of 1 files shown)


robertosfield:
Is there a code example, or some docs that is guiding your changes that I can learn more about the topic?

artemkulyk:
> Is there a code example, or some docs that is guiding your changes that I can learn more about the topic?

https://developer.apple.com/documentation/appkit/nswindowcollectionbehavior?language=objc
https://developer.apple.com/documentation/appkit/nswindow/1419527-togglefullscreen?language=objc

Discussion of this topic:
https://stackoverflow.com/questions/24145269/how-to-begin-nswindow-in-fullscreen


With my previous change the app starts crashing on
https://github.com/vsg-dev/VulkanSceneGraph/blob/master/src/vsg/platform/macos/MacOS_Window.mm#L137.

I really don't know why previous changes worked at the beginning without crashing.

src/vsg/platform/macos/MacOS_Window.mm
```diff
@@ -813,6 +809,14 @@ void createApplicationMenus(void)
     _window.initialFirstResponder = _view;
     [_window makeFirstResponder:_view];
 
+    if (traits->fullscreen)
+    {
+        NSRect screenFrame = [[NSScreen mainScreen] frame];
+        [_window setBackgroundColor: NSColor.whiteColor];
```
robertosfield, line 815: Is this line required as there is a similar but not quite identical code on line 798 above.

artemkulyk: Lines 814-815 are redundant and should be removed. It was blindly copy-pasted from the example.

Screenshot 2022-12-11 at 14 31 24

robertosfield: Thanks for the clarification, I've removed the redundant line and will now merge.

What is the story behind the screenshot?

artemkulyk: It's just vsgpagedlod example with static mapbox tiles and custom map style. By the way, vsg readers extract a file type from the file extension in the path from ```tileReader->imageLayer```. So it is not possible just to change the value of ```tileReader->imageLayer``` and experiment with 3rd party tile providers, like mapbox or HERE. mapbox and HERE tiles urls contain no file name with extension. It's confusing a little bit for newcomers. Possibly, it makes sense to extract file type from the server response instead of the file path.

robertosfield: The vsg::tile ReaderWriter isn't intended to support all ways that tiles could be held on a server, it's a lightweight class that can handle some tile representations.  If a 3rd party source requires a query of supported formats from a server then I presume you'd need to formulate the query and way of mapping this to the final filename and extensions to sue then it'd need a new ReaderWriter that uses curl and as such would belong in vsgXchange.

robertosfield: This type of discussion is best done within Discussions rather piggybacking off a PR.

PR #663, opened 2022-12-15
Create and install shared library with version on MinGW

rhabacker:
## Description
In package-oriented distributions like openSUSE, shared libraries are usually installed with a version to better handle dependencies between packages.

The main rationale is to allow for two or more (incompatible) versions of a shared library, such as libfoo.so.1 and libfoo.so.2 to be independently selectively installable (and trackable by rpm) at the same time without causing package conflicts.

To achieve this also for MinGW based packages installed on a distribution, the shared library 'vsg' for this compiler is also created and installed with version. Since the name of the generated import library is not affected, there are no changes for dependent packages.

## Type of change

## How Has This Been Tested?

- [x] Did a build of vsg and depending packages with this commit applied at https://build.opensuse.org/package/show/home:rhabacker:branches:games:mingw32/mingw32-libvsg 

**Test Configuration**:
* Host: opensuse cloud build system
* Toolchain: mingw gcc 12
* SDK: mingw-header/runtime 10

## Checklist:

### DIFF
--- a/src/vsg/CMakeLists.txt
+++ b/src/vsg/CMakeLists.txt
@@ -432,6 +432,9 @@ set_property(TARGET vsg PROPERTY VERSION ${VSG_VERSION_MAJOR}.${VSG_VERSION_MINO
 set_property(TARGET vsg PROPERTY SOVERSION ${VSG_SOVERSION})
 set_property(TARGET vsg PROPERTY POSITION_INDEPENDENT_CODE ON)
 set_property(TARGET vsg PROPERTY CXX_STANDARD 17)
+if(MINGW)
+    set_property(TARGET vsg PROPERTY RUNTIME_OUTPUT_NAME vsg-${VSG_SOVERSION})
+endif()
 
 target_compile_definitions(vsg PRIVATE ${EXTRA_DEFINES})
 target_include_directories(vsg PUBLIC $<BUILD_INTERFACE:${VSG_SOURCE_DIR}/include> $<BUILD_INTERFACE:${VSG_BINARY_DIR}/include>)


robertosfield:
I'm reluctant to add platform specific special cases, I would like to explore exactly what the VSG installs right now on MingW without these changes, and what it does with them.

For reference on my Kubuntu system when I compile as shared library and install I see the libs:

$ ls ~/Install/lib/libvsg*.so* -1
/home/robert/Install/lib/libvsg.so
/home/robert/Install/lib/libvsg.so.1.0.3
/home/robert/Install/lib/libvsg.so.11

From the sound of it you are trying to replicate this in MinGW, but I'm not clear what parts you've replicated, what parts you haven't.    Having things work different on different platforms is not idea, but fitting in with local contents is also needs to be born in mind.

I presume other CMake users have this same issue under MinGW/Windows so perhaps CMake itself could provide the behavior rather than having to replicate it in our local CMake scripts.

rhabacker:
An installation of the mentioned files on an openSUSE 64bit distribution would lead to the following results 
 
> usr/lib64/libvsg.so
> /usr/lib64/libvsg.so.1.0.3
> /usr/lib64/libvsg.so.11

These files would be put into two packages:
 
- Development package (named libvsg-devel) 
    
    /usr/lib64/libvsg.so

- associated runtime package (libvsg11)

    /usr/lib64/libvsg.so.1.0.3
    /usr/lib64/libvsg.so.11

On the same distribution, an installation of mingw32-libvsg creates

    /usr/i686-w64-mingw32/sys-root/mingw/lib/libvsg.dll.a 
    /usr/i686-w64-mingw32/sys-root/mingw/bin/libvsg.dll

These files would also be placed in two packages:

- Development package (named mingw32-libvsg-devel). 

    /usr/i686-w64-mingw32/sys-root/mingw/lib/libvsg.dll.a
 
- Shared library in the runtime package (mingw32-libvsg)

    /usr/i686-w64-mingw32/sys-root/mingw/bin/libvsg.dll

Only the shared library should be changed, which should get a suffix `-<SOVERSION>`, as for example happened with the existing curl library.

    /usr/i686-w64-mingw32/sys-root/mingw/bin/libcurl-4.dll

For vsg this should be done by changing from 

    /usr/i686-w64-mingw32/sys-root/mingw/bin/libvsg.dll

to
   
    /usr/i686-w64-mingw32/sys-root/mingw/bin/libvsg-11.dll

using the current SOVERSION.
    
> I assume other CMake users have the same problem on MinGW/Windows, so maybe CMake itself could provide the behavior instead of replicating it in our local CMake scripts.

In the current cmake version 2.5, according to current knowledge, such support is not included, so it has always been done in the local CMake scripts (see osg for example).

I will open a corresponding request and submit a matching patch. The patch, if accepted, would probably not be included before cmake version 2.26, i.e. until this version is widely distributed, one would have to make do with the current patch.

If it is known with which version the patch will be released, one could change to the use the new cmake build-in support with this  

    if(MINGW AND CMAKE_VERSION VERSION_GREATER_EQUAL 2.6.0)
       set(CMAKE_MINGW_VERSIONED_SONAME 1)
    elseif(MINGW)
        set_property(TARGET vsg PROPERTY RUNTIME_OUTPUT_NAME vsg-${VSG_SOVERSION})
    endif()

robertosfield:
Thanks for the explanation.

Is there any reason why we shouldn't apply the -${SO_VERSION} to all Windows shared library builds?

My preference would be for Windows have libvsg-${SO_VERSION}.dll and libvsg.a that references it, regardless of the build tool you use.

The .dll.a seems a bit awkward but I can see that it would allow one to build a static and shader library version of the VSG and install it on one system, so can see value in using that.  Are other projects using the .dll.a convention?

rhabacker:
> Thanks for the explanation.
> 
> Is there any reason why we should not apply the -${SO_VERSION} to all Windows shared library builds?

For the local patch, it can be easily extended by using `WIN32` instead of `MINGW`.
Currently I only have a working patch for mingw.  To have appropriate support in cmake, this would need to be investigated, implemented and tested. May be it is easy too, but I haven't looked yet.

> I would prefer Windows to have libvsg-${SO_VERSION}.dll and libvsg.a pointing to it, regardless of the build tool used.

Unfortunately, the compilers and build systems see it differently.

> The .dll.a seems a bit cumbersome. 

There are fixed conventions built into the buildsystems and compilers/linkers. e.g. for the filename suffix [cmake:set(MAKE_SHARED_LIBRARY_SUFFIX](https://gitlab.kitware.com/search?search=set%28CMAKE_SHARED_LIBRARY_SUFFIX&nav_source=navbar&project_id=541&group_id=415&search_code=true&repository_ref=master) and  prefix: msvc uses none, MinGW 'lib', cygwin 'cyg' etc. [cmake:set(MAKE_SHARED_LIBRARY_PREFIX)](https://gitlab.kitware.com/search?search=set%28CMAKE_SHARED_LIBRARY_PREFIX&nav_source=navbar&project_id=541&group_id=415&search_code=true&repository_ref=master)

> but I can see that it would allow one to build a static and a shared library version of the VSG and install them on a system, so I can see the value in using that. 

Yes, that's the advantage, you install both if you want to and decide later which variant you want to use.  With msvc you don't have this possibility to distinguish, so often a workaround is used and `_static` is appended to the filename.

Do other projects use the .dll.a convention?

According to cmake git repo, cygwin gcc, clang and mingw gcc understand this extension [.dll.a]([https://gitlab.kitware.com/search?group_id=415&nav_source=navbar&page=2&project_id=541&repository_ref=master&search=.dll.a&search_code=true) and so do other common build systems like meson, qmake.

robertosfield:
Thanks for the links.  I've had a read through and understand the topic a little better now.  My current thought is that the Visual Studio Windows .dll should have the SOVERSION as well, so using WIN32 seems appropriate.

I will merge as is then we can create a branch to test using this for all of WIN32.

robertosfield:
I have done the merge and then changed the CMake check to use WIN32 rather than MINGW, the branch is:

https://github.com/vsg-dev/VulkanSceneGraph/tree/Windows_SOVERSION_dll

I'll watch the automated builds now to see is things still compile OK, if it does we'll need  Windows users using VS, MinGW and Cygwin to test it out before we merge with VSG master.

I'll make a 1.0.3 release for this and other changes before Christmas.

rhabacker:
> I have done the merge and then changed the CMake check to use WIN32 rather than MINGW, the branch is:
> https://github.com/vsg-dev/VulkanSceneGraph/tree/Windows_SOVERSION_dll

I see 

> if it does we'll need Windows users using VS

The [VC build](https://github.com/vsg-dev/VulkanSceneGraph/actions/runs/3705873941/jobs/6280356983) shows that the installed file names are as expected:
    -- Installing: C:/Program Files (x86)/vsg/lib/vsg.lib
    -- Installing: C:/Program Files (x86)/vsg/bin/vsg-11.dll

>  MinGW

The vsg-dev related [CI build jobs for mingw](https://github.com/vsg-dev/VulkanSceneGraph/actions/workflows/win-cygwin-mingw.yml) was not triggered by this branch and only generates static libraries
On my [personal mingw32 builds](https://build.opensuse.org/package/show/home:rhabacker:branches:games:mingw32/mingw32-libvsg) it works as expected 
[  426s] -- Installing: /home/abuild/rpmbuild/BUILDROOT/mingw32-libvsg-1.0.0+20221215+git.1ea9c86-lp153.439.1.x86_64/usr/i686-w64-mingw32/sys-root/mingw/lib/libvsg.dll.a
[  426s] -- Installing: /home/abuild/rpmbuild/BUILDROOT/mingw32-libvsg-1.0.0+20221215+git.1ea9c86-lp153.439.1.x86_64/usr/i686-w64-mingw32/sys-root/mingw/bin/libvsg-11.dll

> Cygwin
The vsg-dev related [CI build jobs for mingw](https://github.com/vsg-dev/VulkanSceneGraph/actions/workflows/win-cygwin-mingw.yml) was not triggered by this branch and only generates static libraries

is there a CI job for building on cygwin ?

robertosfield:
I think it's safe enough to merge with master so I've gone ahead and merged and have started the windows-mingw-cygwin action:

https://github.com/vsg-dev/VulkanSceneGraph/actions/runs/3712142913

rhabacker:
> I think it's safe enough to merge with master so I've gone ahead and merged and have started the windows-mingw-cygwin action:
> 
> https://github.com/vsg-dev/VulkanSceneGraph/actions/runs/3712142913

Please note that these jobs are not building a shared library.

rhabacker:
> > I think it's safe enough to merge with master so I've gone ahead and merged and have started the windows-mingw-cygwin action:
> > https://github.com/vsg-dev/VulkanSceneGraph/actions/runs/3712142913
> 
> Please note that these jobs are not building a shared library.

It is required to add -DBUILD_SHARED_LIBS=1 to https://github.com/vsg-dev/VulkanSceneGraph/blob/master/.github/workflows/win-cygwin-mingw.yml#L27 and https://github.com/vsg-dev/VulkanSceneGraph/blob/master/.github/workflows/win-cygwin-mingw.yml#L65 to see this change.

PR #666, opened 2022-12-16
Add creating a line to class Builder

rhabacker:
## Description
The commit from this merge request add a method to the class vsg::Builder to draw a line similar to the other available primitives.

Fixes vsg-dev/osg2vsg#63 (partial)

## Type of change

## How Has This Been Tested?

- [x] Compiled vsgbuilder from the vsgExample project after added drawing line support

**Test Configuration**:
* Hardware: GTX-1070
* Toolchain: g++ 7
* SDK: Vulkan 1.3.231

## Checklist:

- [ ] Any dependent changes have been merged and published in downstream modules (pending, see https://github.com/vsg-dev/vsgExamples/pull/197)

### DIFF
--- a/include/vsg/utils/Builder.h
+++ b/include/vsg/utils/Builder.h
@@ -90,7 +90,7 @@ namespace vsg
     VSG_type_name(vsg::GeometryInfo);
 
     /// Builder class that creates subgraphs that can render primitive geometries.
-    /// Supported shapes are Box, Capsule, Cone, Cylinder, Disk, Quad, Sphere and HeightField.
+    /// Supported shapes are Box, Capsule, Cone, Cylinder, Disk, Line, Quad, Sphere and HeightField.
     /// Uses GeometryInfo and StateInfo to guide the geometry position/size and rendering state.
     class VSG_DECLSPEC Builder : public Inherit<Object, Builder>
     {
@@ -105,6 +105,7 @@ namespace vsg
         ref_ptr<Node> createCone(const GeometryInfo& info = {}, const StateInfo& stateInfo = {});
         ref_ptr<Node> createCylinder(const GeometryInfo& info = {}, const StateInfo& stateInfo = {});
         ref_ptr<Node> createDisk(const GeometryInfo& info = {}, const StateInfo& stateInfo = {});
+        ref_ptr<Node> createLine(const GeometryInfo& info, const StateInfo& stateInfo = {});
         ref_ptr<Node> createQuad(const GeometryInfo& info = {}, const StateInfo& stateInfo = {});
         ref_ptr<Node> createSphere(const GeometryInfo& info = {}, const StateInfo& stateInfo = {});
         ref_ptr<Node> createHeightField(const GeometryInfo& info = {}, const StateInfo& stateInfo = {});
@@ -126,6 +127,7 @@ namespace vsg
         GeometryMap _capsules;
         GeometryMap _cones;
         GeometryMap _cylinders;
+        GeometryMap _lines;
         GeometryMap _quads;
         GeometryMap _spheres;
         GeometryMap _heightfields;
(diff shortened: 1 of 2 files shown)


robertosfield:
I'm rather lost why a building a line segment is of value from vsg::Builder.  It's really heavy way to implement the drawing of single line segment, and really not what I'd want any VSG user to be doing as it's just so inefficient.

What is the purpose here?

robertosfield:
I have reflected on the need for generating lines and I think the right approach is to passing in a complete vsg::vec3Array or possibly a list of vsg::vec3Array to generate lines for.  Creating primitives mesh might also be something you'd want to do in this way.  Perhaps it could all be just done together.

I am going to close this PR as I just don't feel it's useful for real world applications.  If we want an API to make it lines and mesh geometries more easily then vsg::Builder might be a starting place, but to brain storm what this might be I think a Discussion thread would be the right place to do this.

rhabacker:
> I have reflected on the need for generating lines and I think the right approach is to passing in a complete vsg::vec3Array or possibly a list of vsg::vec3Array to generate lines for.

This means that when drawing lines it is necessary that 
1. geometryInfo gets an additional member for setting vertices, which contains the points for the lines
2. since there is currently no explicit way in the StateInfo class to set the topology of the GraphicPipeline, it is necessary to enable the wireframe in the vsg::StateInfo instance to display lines.

Or is there more to do ?

rhabacker:
> Or is there more to do ?

What about normals and texture coordinates?  When you create a line, they are calculated in createLine(), but if you use more than one line, I assume they have to be specified as well. Or can they be calculated completely in createLines() too ?

robertosfield:
All good questions, that I don't yet have quick answers to, I will need to sit down and ponder on the issue.

vsg::Bulder was written specifically for primitives shapes like cubes, sphere etc, rather than free form meshes/points/lights so the StateInfo/GeometryInfo start creaking at the seams when we start trying to generalize them to this more general purpose.

It may be that we need to have two builder classes - one for shapes, and one for free forms geometry setup.  I need time to think about the topic though.

rhabacker:
> All good questions, that I don't yet have quick answers to, I will need to sit down and ponder on the issue.
> 
> vsg::Bulder was written specifically for primitives shapes like cubes, sphere etc, rather than free form meshes/points/lights so the StateInfo/GeometryInfo start creaking at the seams when we start trying to generalize them to this more general purpose.
> 
> It may be that we need to have two builder classes - one for shapes, and one for free forms geometry setup. I need time to think about the topic though.

Vulkan specifies several [primitive types] (https://www.khronos.org/registry/vulkan/specs/1.2-extensions/man/html/VkPrimitiveTopology.html). Perhaps an additional "PrimitiveBuilder" class would fit.

rhabacker:
> Vulkan specifies several [primitive types](https://www.khronos.org/registry/vulkan/specs/1.2-extensions/man/html/VkPrimitiveTopology.html). Perhaps an additional `PrimitiveBuilder` class would fit.

>>    geometryInfo gets an additional member for setting vertices, which contains the points for the lines

May be PrimitiveGeometryInfo or GeometryInfoPrimitive ?

>>    since there is currently no explicit way in the StateInfo class to set the topology of the GraphicPipeline

PrimitiveStateInfo or StateInfoPrimitive ?

rhabacker:
> > It may be that we need to have two builder classes - one for shapes, and one for free forms geometry setup. I need time to think about the topic though.
> 
> Vulkan specifies several [primitive types](https://www.khronos.org/registry/vulkan/specs/1.2-extensions/man/html/VkPrimitiveTopology.html). Perhaps an additional `PrimitiveBuilder` or `BuilderPrimitive` class would fit.

I see a current use for such a class in osg2vsg to avoid multiple implementations when converting osg geometry to vsg, since the necessary steps are basically the same.

PR #747, opened 2023-03-10
Add option for source-level shader debugging in RenderDoc

timoore:
This allows a user who is creating their own shader sets to do source-level debugging of the shaders. This doesn't work for the builtin VSG shader sets, and can't unless they too are compiled with debugging enabled.

This is useful and pretty cool as well. See https://youtu.be/Fja4lT508cA for a demo of source-level shader debugging.

### DIFF
--- a/include/vsg/io/Options.h
+++ b/include/vsg/io/Options.h
@@ -89,6 +89,7 @@ namespace vsg
         ///     "text" will substitute for vsg::createTextShaderSet()
         std::map<std::string, ref_ptr<ShaderSet>> shaderSets;
 
+        bool shaderSourceDebugging = false;
     protected:
         virtual ~Options();
     };
--- a/include/vsg/state/ShaderModule.h
+++ b/include/vsg/state/ShaderModule.h
@@ -50,6 +50,7 @@ namespace vsg
         SpirvTarget target = SPIRV_1_0;
         bool forwardCompatible = false;
         std::set<std::string> defines;
+        bool sourceDebugging = false;
 
         int compare(const Object& rhs_object) const override;
 
--- a/include/vsg/utils/ShaderSet.h
+++ b/include/vsg/utils/ShaderSet.h
@@ -76,7 +76,7 @@ namespace vsg
     {
     public:
         ShaderSet();
-        explicit ShaderSet(const ShaderStages& in_stages);
+        explicit ShaderSet(const ShaderStages& in_stages, ref_ptr<ShaderCompileSettings> in_hints = {});
 
         /// base ShaderStages that other variants as based on.
         ShaderStages stages;
@@ -88,6 +88,7 @@ namespace vsg
         std::set<std::string> optionalDefines;
         GraphicsPipelineStates defaultGraphicsPipelineStates;
 
+        ref_ptr<ShaderCompileSettings> defaultShaderHints;
         /// variants of the rootShaderModule compiled for different combinations of ShaderCompileSettings
         std::map<ref_ptr<ShaderCompileSettings>, ShaderStages, DereferenceLess> variants;
 
--- a/src/vsg/io/Options.cpp
+++ b/src/vsg/io/Options.cpp
@@ -44,7 +44,8 @@ Options::Options(const Options& options) :
     mapRGBtoRGBAHint(options.mapRGBtoRGBAHint),
     sceneCoordinateConvention(options.sceneCoordinateConvention),
     formatCoordinateConventions(options.formatCoordinateConventions),
-    shaderSets(options.shaderSets)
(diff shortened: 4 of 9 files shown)


robertosfield:
I am back looking at the this PR again and reviewing the changes to ShaderCompiler I'm now curious about more fully supporting the glslang::SpvOptions structure:

struct SpvOptions {
    bool generateDebugInfo {false};
    bool stripDebugInfo {false};
    bool disableOptimizer {true};
    bool optimizeSize {false};
    bool disassemble {false};
    bool validate {false};
    bool emitNonSemanticShaderDebugInfo {false};
    bool emitNonSemanticShaderDebugSource{ false };
};

In the case of a few of the functions they rely upon SPIRV-Tools so we'd need to pull that in and compile it like I've done with glslang.

Would could possibly cherry-pick the easy to map options like generateDebugInfo and put it into ShaderCompileSettings, some of the options won't be appropriate.

With the built-in shaders I think if one assigns a ShaderSet to vsg::Options::shaderSet[] maps and enable the generateDebugInfo on it then it should automatically recompile the shader source to SPRIV if not ShaderSet::variants match that.  If this does work then we wouldn't need to put any shader options directly into vsg::Options.

The addition of ShaderSet::defaultShaderHints is probably appropriate.

What I'm inclined to do is do cherry-pick elements of this PR that look uncontroversial, and perhaps tweak the naming, then we can make an assessment of what more might be appropriate to merge.

robertosfield:
Just spotted the changes to the serialization isn't yet protected by a version check, I'll add this.

I have merged the changes as a timoore-shader-source-debugging branch, and will cherry-pick from this.

robertosfield:
I have created a SpvOptions_generateDebugInfo branch, only two commits so far that add the ShaderCompileSetttings and ShaderCompiler changes.:

https://github.com/vsg-dev/VulkanSceneGraph/compare/SpvOptions_generateDebugInfo

timoore:
> I am back looking at the this PR again and reviewing the changes to ShaderCompiler I'm now curious about more fully supporting the glslang::SpvOptions structure:
There is probably useful stuff there, although all of the three debugging options seem to be necessary to generate the symbols for source-level debugging (and the option to the parser is needed as well).
> 
> struct SpvOptions { bool generateDebugInfo {false}; bool stripDebugInfo {false}; bool disableOptimizer {true}; bool optimizeSize {false}; bool disassemble {false}; bool validate {false}; bool emitNonSemanticShaderDebugInfo {false}; bool emitNonSemanticShaderDebugSource{ false }; };
> 
> In the case of a few of the functions they rely upon SPIRV-Tools so we'd need to pull that in and compile it like I've done with glslang.
> 
I obviously didn't look at that at all. The introspection functions of the SPIRV-Tools are supposed to be useful.

> With the built-in shaders I think if one assigns a ShaderSet to vsg::Options::shaderSet[] maps and enable the generateDebugInfo on it then it should automatically recompile the shader source to SPRIV if not ShaderSet::variants match that. If this does work then we wouldn't need to put any shader options directly into vsg::Options.
Ok, but the source code to the shaders needs to be available. I admit that I'm not clear if that's the case with the built-in shaders :smile:

robertosfield:
I have finished the changes that I feel are appropriate and have created a PR #748.

The difference are:

* The ShaderCompileSet5ettings::generateDebugInfo naming 
* Update of version to 1.0.4 and use of this in the serialization changes
* ShaderSettings::defaultShaderHints is by default null.
* No changes to Options as I think assign a custom ShaderSet is better and handle further additions cleanly.

I haven't tested this in yet, I'll leave this PR open for discussion.  I'll leave it overnight before I merge with master.

robertosfield:
The VulkanPBRT project used SPIRV-Tools for the introspection capabilities, but their work predated vsg::ShaderSet which provides a richer set of introspection capabilities so I don't think you'd use that now unless you had no option but to work with pre-existing .spv files.

I think the glslang HLSL support that I believe utilizes SPIRV-Tools and the possibility of optimization might be a bigger reason  to pull in SPIRV-Tools.

However, I'm reluctant to add back in a dependency that caused hassles for various users.  I made the change to compiling glslang from source in the VSG to work around these hassles.  If we get to a point where we really need functionality in SPIRV-Tools then we can look at it then.  For now I'd like to expose the functionality that gives us useful features without extra complexity for end users.

robertosfield:
I have merged the #748 branch with VSG master and also merged the change to vsgviewer to vsgExamples master so will close this PR.

Could you write up an introduction to using RenderDoc on the vsg-users Discussion thread, I haven't used this enough to do it justice myself but I think the community might find it really helpful to know what it is and how it can be used, and the new generateDebugInfo support.  Thanks.

PR #748, opened 2023-03-11
Added SpvOptions::generateDebugInfo support to ShaderCompileSettings.

robertosfield:


timoore:
It's too bad that both our solutions require changes to / cooperation from the client application. Dare I suggest an environment variable that would enable shader source code debugging?

robertosfield:
On Sat, 11 Mar 2023 at 22:46, Tim Moore ***@***.***> wrote:

> It's too bad that both our solutions require changes to / cooperation from
> the client application. Dare I suggest an environment variable that would
> enable shader source code debugging?
>
> The issue of 3rd parties snooping in a company's application shaders is an
issue we'd need to resolve if we did something like this.  Perhaps compile
in the option for enabling /disabilingit via cmake.

I have been wondering whether linking it into the WindowTraints::debugLayer
might be appropriate, though that's currently quite disconnected from
building scene graphs etc.

First thing first though, we just nee dto merge the feature as is, then we
can ponder on further refinements.

Robert.

robertosfield:
Other possibilities:

* Set the generateDebugInfo on the ShaderCompiler and have it do a or between the shader modules settings and it's local one,
* Enable it for debug builds by default, though this still has to handle ShaderSet selection.

robertosfield:
@timoore I will merge this morning after I've done some more thinking about enabling it cleanly.  

Any more ideas?

I am currently wondering if a utility function might be useful for enabling it on the built-in ShaderSet i.e. set up vsg::Options with custom ShaderSet with the defaultShaderHints set to generateDebugInfo.  Don't have any ideas that have popped into my head that are the obvious way to do it.

robertosfield:
I've added a enableGenerateDebugInfo(..) function to vsgviewer and if I use it and use the Vulkan debug layer I get:

        [0] 0x56404e4375e0, type: 3, name: NULL
UNASSIGNED-CoreValidation-Shader-FeatureNotEnabled(ERROR / SPEC): msgNum: -1940616195 - Validation Error: [ UNASSIGNED-CoreValidation-Shader-FeatureNotEnabled ] Object 0: handle = 0x56404e4375e0, type = VK_OBJECT_TYPE_DEVICE; | MessageID = 0x8c548bfd | Shader requires extension VK_KHR_shader_non_semantic_info but is not enabled on the device
    Objects: 1

Have you seen this? 

I haven't checked in my changes to vsgviewer, first I'll look into the above debug error.

timoore:
> I've added a enableGenerateDebugInfo(..) function to vsgviewer and if I use it and use the Vulkan debug layer I get:
> 
> ```
>     [0] 0x56404e4375e0, type: 3, name: NULL
> ```
> 
> UNASSIGNED-CoreValidation-Shader-FeatureNotEnabled(ERROR / SPEC): msgNum: -1940616195 - Validation Error: [ UNASSIGNED-CoreValidation-Shader-FeatureNotEnabled ] Object 0: handle = 0x56404e4375e0, type = VK_OBJECT_TYPE_DEVICE; | MessageID = 0x8c548bfd | Shader requires extension VK_KHR_shader_non_semantic_info but is not enabled on the device Objects: 1
> 
> Have you seen this?
> 
> I haven't checked in my changes to vsgviewer, first I'll look into the above debug error.

I do see this when running vsgCs' worldviewer standalone i.e., not under RenderDoc. I think the Vulkan layer installed by RenderDoc must provide the extension. Do you think it's worthwhile to check for the extension before enabling the feature?

timoore:
> @timoore I will merge this morning after I've done some more thinking about enabling it cleanly.
> 
> Any more ideas?
> 
> I am currently wondering if a utility function might be useful for enabling it on the built-in ShaderSet i.e. set up vsg::Options with custom ShaderSet with the defaultShaderHints set to generateDebugInfo. Don't have any ideas that have popped into my head that are the obvious way to do it.

I like the idea of a global setting in the ShaderCompiler that is enabled by a VSG option. Turning it on in debug builds would work too. I don't see much interest in enabling / disabling the feature on a per-ShaderSet basis. I'm sympathetic to your concerns about third party IP and see why an environment variable might not be a good idea, but I don't want to bend over backwards to make the shader source hard to get at, to the detriment of debuggability. 

Until now I've been dependent on RenderDoc shader debugging with its disassembly, which is not too different from debugging C++ in ReleaseWithDebugInfo mode. This source level debugging is a nice improvement, not vital, but also not revealing a lot more shader implementation than you can already see with RenderDoc.

robertosfield:
I have added fallback for the VK_KHR_SHADER_NON_SEMANTIC_INFO_EXTENSION_NAME for older Vulkan verrsions: 38f669f77e34c555728d3e7e4ae80fe46b300cd7.

I have also checked in the enabling of the generateDebugInfo for the built in ShaderSet in vsgviewer:

https://github.com/vsg-dev/vsgExamples/commit/f2a5c1bc8108ba6f01b089f8f4e969473850a1c7

Enabling the generateDebugInfo is still limited to explicitly enabling it, but it's hard to just do it enable it automatically for all usage cases as precompile SPRIV shaders can't be redone.  

I think it's worth just merging this branch and the change to vsgviewer to get the ball rolling, then later we can add further support when we come up with clean way of doing it.

robertosfield:
@timoore  I have just installed renderdoc from the ubuntu repo but it crashes when I start the application.... So I'm now pulling down the source code and will see how I get on.

Which renderdoc version are you using?

timoore:
> @timoore I have just installed renderdoc from the ubuntu repo but it crashes when I start the application.... So I'm now pulling down the source code and will see how I get on.
> 
> Which renderdoc version are you using?

I downloaded RenderDoc 1.25 from https://renderdoc.org/ . I believe that 1.25 is required for shader source debugging.

PR #767, opened 2023-04-06
Fix bugs in XCB pointer / button motion

timoore:
Button motion events, i.e. pointer motion with buttons pressed, must be requested when the window is created.

The modifier key state bits were getting inserted into the vsg::ButtonMask value.


robertosfield:
How should I reproduce the problems you are seeing that this PR fixes?

timoore:
> How should I reproduce the problems you are seeing that this PR fixes?

If you have the numlock key enabled and print out a move event mask value e.g. by using a debugger and the track ball, you should see that the modifier bit makes its way into the mask.

I'm not entirely sure how the Trackball is working now without the XCB_EVENT_MASK_BUTTON_MOTION request, but while debugging another manipulator I was definitely not seeing motion events with buttons pressed without it.

robertosfield:
Curious that Trackball is working fine as is.  Is the event handler you have that reproduces the issue publicly available?

timoore:
> Curious that Trackball is working fine as is. Is the event handler you have that reproduces the issue publicly available?

My work in progress is in the [map-manipulator](https://github.com/timoore/vsgCs/tree/map-manipulator) branch of vsgCs. It's a port of the osgEarth Earth manipulator by Pelican.

robertosfield:
I have been investigating a multi-threading dynamic data issue most of this afternoon so haven't been able to do the necessary setup to start testing this PR. I will need to do that tomorrow now.

timoore:
So I got confused during my debugging... After looking at the limited documentation of XCB, it's clear that XCB_EVENT_MASK_BUTTON_MOTION is redundant, and my code works without that event mask. On the other hand, the problem of modifier bits leaking into the button mask is real. vsg::Trackball isn't affected by it because it explicitly tests the button bits in which it is interested.

robertosfield:
Thanks for the refinement. Now merging.

PR #768, opened 2023-04-06
Add RefPtr to Inherit class

asherikov:
## Description

What do you think about adding some syntactic sugar to `Inherit` class to allow using `vsg::CLASS::RefPtr` instead of `vsg::ref_ptr<vsg::CLASS>`?

## Type of change

### DIFF
--- a/include/vsg/core/Inherit.h
+++ b/include/vsg/core/Inherit.h
@@ -27,6 +27,9 @@ namespace vsg
     class Inherit : public ParentClass
     {
     public:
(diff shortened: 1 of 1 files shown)


robertosfield:
I'm not a fan of obscuring functionality and introducing any extra naming conventions.  If we do add stuff like this it really needs to add a capability and be coherent with the rest of the library.

What value does this addition have? What problem does it solve?

asherikov:
Ok, the change is pretty much cosmetic.

PR #769, opened 2023-04-07
[ADDED] VSG_SUPPORTS_Windowing option to reduce build dependencies

siystar:
# Pull Request Template

## Description

Added optional VSG_SUPPORTS_Windowing switch to reduce build dependencies when using the VSG with external windowing such as SDL, QT or headless

Fixes # (issue)

## Type of change

## How Has This Been Tested?

Successfully built the VSG without native windowing headers installed

**Test Configuration**:
* Firmware version:
* Hardware:
* Toolchain:
* SDK:

## Checklist:


robertosfield:
This is a feature I've been to provide, but done a bit differently.

I have been wanting to provide the ability to have zero or more windowing solutions built, rather than just one "native" windowing supported by the the VSG itself.  For instance on Linux we have the option of Xcb, Xlib or Wayland, one might want to select within none, one, or a combination of.

The WindowTraits would need to include a window preference parameter to help select which windowing to use to create a window, and the vsg::Window::create() mechanism would need to check from a list of available options.  Possibly would could put WindowCreation prototypes into the vsg::ObjectFactory.

With these schemes we'll need coherent CMake checks and options to say which windowing implementation to check for and compile against.  My thought select none would be an option. 

Potentially we could introduce a CMake var that does what you want but does it in a way that will be compatible with the future support mentioned above.

robertosfield:
@siystar I have created a Windowing branch for do experiments with changes to the build of the windowing support:

https://github.com/vsg-dev/VulkanSceneGraph/tree/Windowing

The only change I've checked in so far has been to rename include/vsg/platform/unix to include/vsg/xcb in prep for future Wayland and Xlib support.

While thinking about the linking to the different Window libs, especially in the multiple Windowing support case then perhaps it might just be easier to put both the head and implementation into the headers for the include/vsg/platform/* code.

The VSG would then not build any Windowing as part of the lib, and applications would then need to link and use the appropriate headers.  Making this work seamlessly for current VSG applications/examples might require a bit of work but this might be possible.

I do kinda like the idea of the libvsg.so being agnostic of the Windowing system that users want to use.

robertosfield:
I have decided not to tackle a major change to Windowing for now and have used your PR as a starting place.  The following branch is what I am about to merge with VSG master:

https://github.com/vsg-dev/VulkanSceneGraph/tree/siystar-windowing

Additions to your PR are:

1. include/vsg/core/Version.h now contains the setting of the VSG_SUPPORTS_Windowing var
2. the installed cmake/vsg/vsgConfig.cmake now contains the appropriate dependencies for that build, so if you've disabled the Windowing support no additional dependencies will be listed.
3. include/vsg/platform/unix is now named include/vsg/platform/xcb
4. a default implementation of vsg::Window::create(ref_ptr<WindowTraits>) is now provided by Window.cpp which returns a null ref_ptr<>.  The examples all catch this correctly and fail to run rather than report a linking error as they did with this PR in it's original form.

I have tested under Linux with VSG_SUPPORTS_Windowing set to 0 and 1 and the examples either fail gracefully or work as before.  vsgQtWindow works with either setting.

robertosfield:
I have now merge the siystar-windowing branch with VSG master.

PR #782, opened 2023-04-13
Add CompileManager::compileImmediate()

timoore:
Add an additional CompileTraversal object to the compile manager which can be used without waiting for a traversal to become available.

This is intended for the somewhat dubious usage of compiling a small part of a scene graph, e.g., a BindDescriptorSet command, on the fly. A call to the existing CompileManager::compile() function can stall waiting for a long time if a compile traversal is also happening in a paging thread.

### DIFF
--- a/include/vsg/app/CompileManager.h
+++ b/include/vsg/app/CompileManager.h
@@ -62,12 +62,19 @@ namespace vsg
         /// compile object
         CompileResult compile(ref_ptr<Object> object, ContextSelectionFunction contextSelection = {});
 
+        //// compile object without waiting for a CompileTraversal to become available. This should
+        //// only be called from one thread (probably the main thread). It's most useful when the
+        //// compilation time will be short and objects will not have to be uploaded to the GPU.
+        CompileResult compileImmediate(ref_ptr<Object> object, ContextSelectionFunction contextSelection = {});
     protected:
         using CompileTraversals = ThreadSafeQueue<ref_ptr<CompileTraversal>>;
         size_t numCompileTraversals = 0;
         ref_ptr<CompileTraversals> compileTraversals;
+        ref_ptr<CompileTraversal> immediateTraversal;
 
         CompileTraversals::container_type takeCompileTraversals(size_t count);
+        CompileResult compileAux(ref_ptr<Object> object, ContextSelectionFunction contextSelection,
+                                 ref_ptr<CompileTraversal> compileTraversal);
     };
     VSG_type_name(vsg::CompileManager);
 
--- a/src/vsg/app/CompileManager.cpp
+++ b/src/vsg/app/CompileManager.cpp
@@ -76,6 +76,7 @@ CompileManager::CompileManager(Viewer& viewer, ref_ptr<ResourceHints> hints)
 #else
     numCompileTraversals = 1;
 #endif
+    immediateTraversal = CompileTraversal::create(*ct);
 }
 
 CompileManager::CompileTraversals::container_type CompileManager::takeCompileTraversals(size_t count)
@@ -102,6 +103,7 @@ void CompileManager::add(ref_ptr<Device> device, const ResourceRequirements& res
 
         compileTraversals->add(ct);
     }
+    immediateTraversal->add(device, resourceRequirements);
 }
 
 void CompileManager::add(Window& window, ref_ptr<ViewportState> viewport, const ResourceRequirements& resourceRequirements)
@@ -113,6 +115,7 @@ void CompileManager::add(Window& window, ref_ptr<ViewportState> viewport, const 
 
         compileTraversals->add(ct);
     }
+    immediateTraversal->add(window, viewport, resourceRequirements);
 }
 
 void CompileManager::add(Window& window, ref_ptr<View> view, const ResourceRequirements& resourceRequirements)
@@ -124,6 +127,7 @@ void CompileManager::add(Window& window, ref_ptr<View> view, const ResourceRequi
 
         compileTraversals->add(ct);
     }
+    immediateTraversal->add(window, view, resourceRequirements);
 }
 
 void CompileManager::add(Framebuffer& framebuffer, ref_ptr<View> view, const ResourceRequirements& resourceRequirements)
@@ -135,6 +139,7 @@ void CompileManager::add(Framebuffer& framebuffer, ref_ptr<View> view, const Res
 
         compileTraversals->add(ct);
     }
+    immediateTraversal->add(framebuffer, view, resourceRequirements);
 }
 
 void CompileManager::add(const Viewer& viewer, const ResourceRequirements& resourceRequirements)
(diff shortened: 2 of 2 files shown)


timoore:
I could see reworking this if it's thought preferable for a client to have their own CompileTraversal to call when they please. In that case, the old compile / new compileAux should be part of the CompileTraversal interface.

robertosfield:
This just looks like trouble waiting to happen. Adding complexity as a workaround to a performance issue isn't the right way to go, especially when it would be so fragile and open to misuse.

There CompileManager carefully manages the CompileTraversal objects to avoid race conditions so just bypassing it won't be safe.  It may work most of the time you test it but it's likely to trip you up at some point in the future, in possibly difficult to diagnose ways.

I think a better step would be to add easier support for assigning multiple CompileTraversal to the CompileManager, and if we find race conditions then we can look at them in a general sense.  Adding a size_t numCompileTraversal = 1 to the CompileManager would be a first step.  And adding a method to assign new CompileTraversal after construction another, though we'd need to take care what happens with handling the windows etc. which we need to compile for.

robertosfield:
CompileManager exists to ease the burden on developers because it's a able to handle users adding new views and windows into the mix in a central way.  If users start creating and managing their own CompileTraversal then they'll have to manage what happens when new views/windows get added and will need to handle any risks of race conditions.

The right thing to do is refine the CompileManager to perform better and robustly whilst keep the public API simple. We also need to to avoid adding implicit rules of when and how certain methods can be called.  The current API is distilled down to what is safe and easy to use.  

I'm going to close this PR as it goes in the wrong direction adding complexity and could easily be misused and cause crashes.

timoore:
I just want to create a descriptor set and update it, so I guess I'll write code to do that.

timoore:
> To create the Vulkan objects for the descriptor set you'll need to create a vsg::Context to pass to the descriptorSet->compile(context) method. For this really limited usage you might not need to set up too much on the vsg::Context. Normally it needs lots of info about viewer/views/viewports so that it can handle all the different types of compilation of Vulkan objects and data transfer, but in your case you might be able to get away with something quite minimal.

Right, I think it will be easy to do that.
> 
> However, I still think having multiple CompileTraversal and the vsg::Context they manage would be the more general purpose way of tackling it.

I totally agree. It's a big task. Should the "user" ensure that common subgraphs are never compiled at the same time, or do guards need to be built into the compile traversal and scene graph objects? Is shader compilation special? ...
> 
> Are there multiple database threads all contending for the CompileManager's CompileTraversal? Or it it just one database pager thread and the main thread?
> 
> Perhaps having masks for each CompileTraversal in the CompileManager so you can reserve subsets of the the available CompileTraversal for particular purposes, or perhaps do it be priority. A background thread has low priority while something in the main thread has high priority.
In my usage there are several threads that can turn 3D Tiles into VSG scengraph objects and compile them. Perhaps it could be tuned differently, but it never occurred to me that there would be contention for compile().

robertosfield:
Shader Compilation is kinda orthogonal to creating vulkan objects and transferring data.  However, one can't create pipeline objects with they require Shader Compilation for them.  One could possible run ShaderCompile with it's own set of threads and request compilation, set aside the pipeline creation till it's complete and get one with creation of all the other vulkan objects and transfer of data etc.

Protection from compiling the same subgraphs/objects at the same time as others is ideally you'd do at the top level without needing to mutex local objects.  We don't want to mutex low level stuff like StateCommands and Commands as this would spill over into the record traversal and destroy performance.

We have to be really selective about what we permit to be run multi-threaded and how.

However, they are may be simple things we can do to give us performance without putting in lots of layer of complexity and overhead.  Having categories/priorities of CompileTraversal might be one way of achieving what you need short term without adding significant complexity to CompileManager.

robertosfield:
FYI, the OSG's DatabasePager separated http requests from local file system requests so that the all the threads wouldn't be consumed by the slow http requests. I haven't yet implemented this in the VSG but I'll do it at some stage.  This fast vs slow request approach is similar to what you are wanting on the compile side - we have low latency ops which shouldn't be held up by high latency ones.

robertosfield:
@timoore I done some more thinking about this topic and think the priority approach is right way to go.  I have a rough idea of how to implement it, but can't tackle it right away as I have other work to get on with.  Once I have some time I'll implement this, but this might take a few days/week.  In the mean time I would suggest just sticking with your current patched VSG and once I've refactored CompileManager you can switch across.

timoore:
> @timoore I done some more thinking about this topic and think the priority approach is right way to go. I have a rough idea of how to implement it, but can't tackle it right away as I have other work to get on with. Once I have some time I'll implement this, but this might take a few days/week. In the mean time I would suggest just sticking with your current patched VSG and once I've refactored CompileManager you can switch across.

I've backed out compileImmediate() and copy-pasted it, more or less, into vsgCs. It turned out that I needed an actual compile traversal to compile my state command. I don't know if it's safe, but I get the same benefit as I did from compileImmediate().

I don't think there's really a fast / slow issue -- my intuition says that even the slowest parts of compilation should be plenty fast enough compared to loading the source data via http. Nevertheless, it's taking 100ms to compile a tile in vsgCs, and that's without textures and probably without compiling shaders or pipelines either (steady state time). The problem has got to be waiting for the results of the compile queue submit and contention with actual graphics operations on the graphics queue. I'm going to look at returning a fence or a timeline semaphore as part of CompileResult and have the Viewer test that; maybe it won't wait but will defer the update to the next frame if it isn't ready.

robertosfield:
If there is copying of data during the compile traversal then the copy work will need to submitted to the Vulkan queue and then have to wait on a fence for the completion of the transfer.  

This is a bit different to TransferTask that just submits the command buffer that does the transfer to the GPU then immediately returns to the calling thread.  It's not quite fire and forget though as the semaphore it uses to signal the completion of the copy is assigned to the main rendering submission as a semaphore to wait upon. 

Your suggestion of passing back a semaphore or fence would do something similar to TransferTask in giving your the ability to query the status of the synchronization primitive or use it in the main thread recording.  One would have to tell the CompileManager::compile() that you didn't want it to wait.  If one has multiple logical devices to copy the data to then you'd need a semaphore/fence per logical device.

timoore:
After a deep dive into "off-CPU profiling,", I have a better understanding about latency in the CompileManager, in vsgCs at least. The basic problem is that there is only one compile traversal available, for reasons described above, and vsgCs loader threads go to sleep waiting for it to become available. But the compile traversal itself is often asleep on a fence, waiting for its queued up Vulkan commands to finish. I think I have a nice solution for that, including freeing up the resources used by the compile operation, but it involves timeline semaphores. @robertosfield, would you be willing to require Vulkan 1.2 or at least the VK_KHR_timeline_semaphore, which seems to be well supported?

robertosfield:
On Sun, 16 Apr 2023 at 23:32, Tim Moore ***@***.***> wrote:

> After a deep dive into "off-CPU profiling,", I have a better understanding
> about latency in the CompileManager, in vsgCs at least. The basic problem
> is that there is only one compile traversal available, for reasons
> described above, and vsgCs loader threads go to sleep waiting for it to
> become available. But the compile traversal itself is often asleep on a
> fence, waiting for its queued up Vulkan commands to finish. I think I have
> a nice solution for that, including freeing up the resources used by the
> compile operation, but it involves timeline semaphores. @robertosfield
> <https://github.com/robertosfield>, would you be willing to require
> Vulkan 1.2 or at least the VK_KHR_timeline_semaphore, which seems to be
> well supported?
>
I don't want to add extra constraints above the current Vulkan-1.1
constraint.  We can already have extension support for things like mesh
shaders and ray tracing shaders and would suggest following this approach
for support beyond this.

Issue #783, opened 2023-04-17
CompileManager::compile() is single-threaded and sleeps too much


timoore:
The bug is really "data compilation is really slow in my application." in vsgCs it takes on average 100ms to "compile" a 256x479 (ish) texture. This is a continuation of #782 with an eye to writing down the various performance problems with CompileManager::compile() and inching towards a solution.

In VSG usage, "compile" means "create Vulkan objects and upload to the GPU everything needed to render a scenegraph." The main entry point to compilation is through the CompileManager::compile() function which is called when files are loaded at the beginning of an application and when new data is added via the DatabasePager. The problems described here were encountered in vsgCs which doesn't use the DatabasePager, but the issues are the same.

The data uploaded by compilation, which are usually textures and geometry for rendering, must often be further processed by commands that are dispatched to Vulkan, e.g. to copy textures to fast GPU memory or to create mipmaps. A complication is that this compilation often must be repeated for different usages of the graphics data, due to the way in which Vulkan manages graphical state through "pipeline objects." In VSG a Context object exists for each of these usages; broadly speaking, a context is created for each View and rendering method. At the lowest levels of the VSG scenegraph, different Vulkan pipeline objects are stored for each context. New contexts may be added as a program progresses. compile() does not (intentionally!) touch the Viewer's scene graph; the database pager or application does that in the main thread, using results of compile().

This compilation should be very parallelizable, thanks to the design of Vulkan. The slowest parts -- compilation of shaders from source and pipeline creation -- can be run in parallel with each other and rendering. It's a bit tricky to submit commands to a single Vulkan queue from several threads, but VSG already deals with that. The problems come with synchronization among the several actors:

* The Vulkan spec is fairly non-commital about whether operations on a queue can block or not. VSG queue operations might end up sleeping inside the mutex that protects the main graphics queue, blocking any other queue operations.
* Vulkan temporary memory used by compile() needs to be cleaned up after it is used. This is currently done by waiting -- on the CPU -- on a fence after each compile operation and then doing the cleanup. In the current source code, this prevents any other compilation operation from starting.
* Vulkan rendering commands can't use uploaded data until they are ready. The waiting CPU-side on a fence does take care of this, but using a semaphore during command submission would allow work to proceed on the CPU.
* The VSG low-level data structures can't be changed while in use.  There are several ways this might happen, but the current serial compile basically protects against this.

[...]

PR #786, opened 2023-04-19
Nonblocking compile

timoore:
This makes the compile action non-blocking, relying instead on a new FenceAction to clean up after a compilation and on the semaphore waiting and signalling on the main graphics queue for synchronization of Vulkan objects. This makes a huge improvement to the problems reported in #783: the 100ms compile times mentioned for vsgCs are now .5ms (200x!), without making any effort to run compile actions in parallel.

I'm not sure this is quite ready for merge. There is a possible issue with headless rendering and synchronizing the compilation. I also haven't tested examples that use the database pager. However, I wanted to get the code out there for review and comment.

### DIFF
--- a/include/vsg/vk/Context.h
+++ b/include/vsg/vk/Context.h
@@ -135,8 +135,21 @@ namespace vsg
         void copy(ref_ptr<BufferInfo> src, ref_ptr<BufferInfo> dest);
 
         /// return true if there are commands that have been submitted
+        ///
+        /// record() needs to submit commands e.g., to upload textures into GPU memory, so the
+        /// scene graph compiled here cannot be recorded and rendered until those submissions are
+        /// complete. The sequence of the DatabasePager and Viewer operations mostly take care of
+        /// this: a sub scene graph isn't inserted into the main graph until compilation is
+        /// complete. We are not waiting (stalling) for completion, but the commands necessary to do
+        /// the rendering will have already been submitted before any commands from the main scene
+        /// graph. There does need to be synchronization between the compile commands and the main
+        /// rendering commands; that is normally provided by wait semaphores in the record
+        /// traversal.
+        ///
+        /// XXX If there are no semaphores there, then record() should signal one in its submission.
         bool record();
 
+        //// Returns immediately.
         void waitForCompletion();
 
         ref_ptr<MemoryBufferPools> deviceMemoryBufferPools;
--- a/include/vsg/vk/Fence.h
+++ b/include/vsg/vk/Fence.h
@@ -27,13 +27,17 @@ namespace vsg
         operator VkFence() const { return _vkFence; }
         VkFence vk() const { return _vkFence; }
 
-        VkResult wait(uint64_t timeout) const;
+        VkResult wait(uint64_t timeout);
 
         VkResult reset() const;
 
         VkResult status() const { return vkGetFenceStatus(*_device, _vkFence); }
-
-        bool hasDependencies() const { return (_dependentSemaphores.size() + _dependentCommandBuffers.size()) > 0; }
+        bool hasDependencies() const
+        {
+            return !(_dependentSemaphores.empty()
+                     && _dependentCommandBuffers.empty()
+                     && _actions.empty());
+        }
 
         void resetFenceAndDependencies();
 
@@ -42,7 +46,10 @@ namespace vsg
 
         Device* getDevice() { return _device; }
         const Device* getDevice() const { return _device; }
-
+        using action_container_type = std::list<FenceAction>;
+        void putActions(uint64_t in_submission, action_container_type& actions);
+        uint64_t submission;
+        action_container_type& actions() { return _actions; }
     protected:
         virtual ~Fence();
 
@@ -51,6 +58,7 @@ namespace vsg
         CommandBuffers _dependentCommandBuffers;
 
         ref_ptr<Device> _device;
+        action_container_type _actions;
     };
     VSG_type_name(vsg::Fence);
 
--- a/include/vsg/vk/Queue.h
+++ b/include/vsg/vk/Queue.h
@@ -15,7 +15,9 @@ THE SOFTWARE IS PROVIDED "AS IS", WITHOUT WARRANTY OF ANY KIND, EXPRESS OR IMPLI
 #include <vsg/core/Object.h>
 #include <vsg/vk/vulkan.h>
 
+#include <functional>
 #include <mutex>
+#include <list>
 #include <vector>
 
 namespace vsg
@@ -23,6 +25,9 @@ namespace vsg
     // forward declare
     class Fence;
 
+    //// An action that will be performed once a Fence is waited on successfully.
+    using FenceAction = std::pair<uint64_t, std::function<void()>>;
+    
     /// Queue encapsulates a single vkQueue, used to submit vulkan commands for processing.
     class VSG_DECLSPEC Queue : public Inherit<Object, Queue>
     {
@@ -36,6 +41,14 @@ namespace vsg
         VkResult submit(const std::vector<VkSubmitInfo>& submitInfos, Fence* fence = nullptr);
 
         VkResult submit(const VkSubmitInfo& submitInfo, Fence* fence = nullptr);
+
+        //// Queue submit with a functor that will be called when the commands in this submission
+        //// have completed. This happens when a fence passed in a submission -- either this or a
+        //// later one -- is waited upon. Vulkan guarantees that the commands will have finished at
+        //// that point.
+        VkResult submit(const std::vector<VkSubmitInfo>& submitInfos, std::function<void()> func,
+                        Fence* fence = nullptr);
+        VkResult submit(const VkSubmitInfo& submitInfo, std::function<void()> func, Fence* fence = nullptr);
 
         VkResult present(const VkPresentInfoKHR& info);
 
@@ -56,6 +69,8 @@ namespace vsg
         uint32_t _queueFamilyIndex;
         uint32_t _queueIndex;
         std::mutex _mutex;
+        std::list<FenceAction> _fenceActions;
+        uint64_t _submissionNo;
     };
     VSG_type_name(vsg::Queue);
 
--- a/src/vsg/app/RecordAndSubmitTask.cpp
+++ b/src/vsg/app/RecordAndSubmitTask.cpp
@@ -126,6 +126,7 @@ VkResult RecordAndSubmitTask::finish(CommandBuffers& recordedCommandBuffers)
     if (recordedCommandBuffers.empty())
     {
         // nothing to do so return early
+        // XXX This is probably not the place to guarantee the frame rate.
         std::this_thread::sleep_for(std::chrono::milliseconds(16)); // sleep for 1/60th of a second
         return VK_SUCCESS;
     }
--- a/src/vsg/vk/Context.cpp
+++ b/src/vsg/vk/Context.cpp
@@ -275,17 +275,41 @@ void Context::copy(ref_ptr<BufferInfo> src, ref_ptr<BufferInfo> dest)
     copyBufferCmd->add(src, dest);
 }
 
+namespace
+{
+    // XXX What about BuildAccelerationStructureCommands? These were not being cleared, but that is
+    // probably a bug.
+    struct Cleanup
+    {
+        ref_ptr<CopyAndReleaseImage> copyImageCmd;
+        ref_ptr<CopyAndReleaseBuffer> copyBufferCmd;
+        ref_ptr<CommandBuffer> commandBuffer;
+
+        // destructive constructor
+        Cleanup(ref_ptr<CopyAndReleaseImage>& in_copyImage, ref_ptr<CopyAndReleaseBuffer>& in_copyBuffer,
+                ref_ptr<CommandBuffer>& in_commandBuffer)
+        {
+            copyImageCmd.swap(in_copyImage);
+            copyBufferCmd.swap(in_copyBuffer);
+            commandBuffer.swap(in_commandBuffer);
+        }
+
+        void operator()()
+        {
+            copyImageCmd = nullptr;
+            copyBufferCmd = nullptr;
+            commandBuffer = nullptr;
+        }
+    };
+}
+
 bool Context::record()
 {
     if (commands.empty() && buildAccelerationStructureCommands.empty()) return false;
 
     //auto before_compile = std::chrono::steady_clock::now();
 
-    if (!fence)
-    {
-        fence = vsg::Fence::create(device);
-    }
-    else
+    if (fence)
     {
         fence->reset();
     }
@@ -325,6 +349,10 @@ bool Context::record()
     submitInfo.sType = VK_STRUCTURE_TYPE_SUBMIT_INFO;
     submitInfo.commandBufferCount = 1;
     submitInfo.pCommandBuffers = commandBuffer->data();
+    // A semaphore here synchronizes the submits in this batch, and all the ones before it, up to
+    // the wait stage. This semaphore is not being used now; the wait semaphores in the submit batch for
+    // the scene graph synchronize this batch instead. If there are no semaphores there
+    // (headless?), then this semaphore should be used.
     if (semaphore)
     {
         submitInfo.signalSemaphoreCount = 1;
@@ -337,38 +365,15 @@ bool Context::record()
         submitInfo.pSignalSemaphores = nullptr;
     }
 
-    graphicsQueue->submit(submitInfo, fence);
+    graphicsQueue->submit(submitInfo, Cleanup(copyImageCmd, copyBufferCmd, commandBuffer), fence);
+    commands.clear();
 
     return true;
 }
 
 void Context::waitForCompletion()
 {
-    if (!commandBuffer || !fence)
-    {
-        return;
-    }
-
-    if (commands.empty() && buildAccelerationStructureCommands.empty())
-    {
-        return;
-    }
-
-    // we must wait for the queue to empty before we can safely clean up the commandBuffer
-    uint64_t timeout = 1000000000;
-
-    VkResult result;
-    while ((result = fence->wait(timeout)) == VK_TIMEOUT)
-    {
-        info("Context::waitForCompletion() ", this, " fence->wait() timed out, trying again.");
-    }
-
-    if (result != VK_SUCCESS)
-    {
-        info("Context::waitForCompletion()  ", this, " fence->wait() failed with error. VkResult = ", result);
-    }
-
-    commands.clear();
-    copyImageCmd = nullptr;
-    copyBufferCmd = nullptr;
+    // Nothing to wait for! But, if the caller really needs to wait for completion, then a fence
+    // should be passed in. Somehow.
+    return;
 }
--- a/src/vsg/vk/Fence.cpp
+++ b/src/vsg/vk/Fence.cpp
@@ -11,6 +11,7 @@ THE SOFTWARE IS PROVIDED "AS IS", WITHOUT WARRANTY OF ANY KIND, EXPRESS OR IMPLI
 </editor-fold> */
 
 #include <vsg/core/Exception.h>
+#include <vsg/io/Logger.h>
 #include <vsg/io/Options.h>
 #include <vsg/vk/Fence.h>
 
@@ -52,16 +53,48 @@ void Fence::resetFenceAndDependencies()
 
     _dependentSemaphores.clear();
     _dependentCommandBuffers.clear();
+    if (!_actions.empty())
+    {
+        vsg::warn("fence has actions on reset.");
+    }
+    _actions.clear();
 
     reset();
 }
 
-VkResult Fence::wait(uint64_t timeout) const
+VkResult Fence::wait(uint64_t timeout)
 {
-    return vkWaitForFences(*_device, 1, &_vkFence, VK_TRUE, timeout);
+    VkResult result = vkWaitForFences(*_device, 1, &_vkFence, VK_TRUE, timeout);
+    if (result == VK_SUCCESS)
+    {
+        // I believe the Vulkan spec permits multiple threads to wait on the same fence. However, a
+        // fence cannot be waited on before it is submitted, so it would be hard to do in
+        // practice. If VSG ever does that, then these actions will need to be protected by a
+        // mutex.
+        for (auto& a : _actions)
+        {
+            if (a.first > submission)
+            {
+                vsg::fatal("Fence is performing action for a future submit!");
+            }
+            a.second();
+        }
+        _actions.clear();
+    }
+    return result;
 }
 
 VkResult Fence::reset() const
 {
+    if (!_actions.empty())
+    {
+        vsg::warn("Fence with actions was reset.");
+    }
     return vkResetFences(*_device, 1, &_vkFence);
+}
+
+void Fence::putActions(uint64_t in_submission, action_container_type& actions)
+{
+    submission = in_submission;
+    _actions.splice(_actions.end(), actions);
 }
--- a/src/vsg/vk/Queue.cpp
+++ b/src/vsg/vk/Queue.cpp
@@ -19,7 +19,8 @@ using namespace vsg;
 Queue::Queue(VkQueue queue, uint32_t queueFamilyIndex, uint32_t queueIndex) :
     _vkQueue(queue),
     _queueFamilyIndex(queueFamilyIndex),
-    _queueIndex(queueIndex)
+    _queueIndex(queueIndex),
+    _submissionNo(0)
 {
 }
 
@@ -30,12 +31,22 @@ Queue::~Queue()
 VkResult Queue::submit(const std::vector<VkSubmitInfo>& submitInfos, Fence* fence)
 {
     std::scoped_lock<std::mutex> guard(_mutex);
+    if (fence)
+    {
+        fence->putActions(_submissionNo, _fenceActions);
+    }
+    _submissionNo++;
     return vkQueueSubmit(_vkQueue, static_cast<uint32_t>(submitInfos.size()), submitInfos.data(), fence ? fence->vk() : VK_NULL_HANDLE);
 }
 
 VkResult Queue::submit(const VkSubmitInfo& submitInfo, Fence* fence)
 {
     std::scoped_lock<std::mutex> guard(_mutex);
+    if (fence)
+    {
+        fence->putActions(_submissionNo, _fenceActions);
+    }
+    _submissionNo++;
     return vkQueueSubmit(_vkQueue, 1, &submitInfo, fence ? fence->vk() : VK_NULL_HANDLE);
 }
 
@@ -49,4 +60,28 @@ VkResult Queue::waitIdle()
 {
     std::scoped_lock<std::mutex> guard(_mutex);
     return vkQueueWaitIdle(_vkQueue);
+}
+
+VkResult Queue::submit(const std::vector<VkSubmitInfo>& submitInfos, std::function<void()> func, Fence* fence)
+{
+    std::scoped_lock<std::mutex> guard(_mutex);
+    _fenceActions.push_back({_submissionNo, std::move(func)});
+    if (fence)
+    {
+        fence->putActions(_submissionNo, _fenceActions);
+    }
+    _submissionNo++;
+    return vkQueueSubmit(_vkQueue, static_cast<uint32_t>(submitInfos.size()), submitInfos.data(), fence ? fence->vk() : VK_NULL_HANDLE);
+}
+
+VkResult Queue::submit(const VkSubmitInfo& submitInfo, std::function<void()>  func, Fence* fence)
+{
+    std::scoped_lock<std::mutex> guard(_mutex);
+    _fenceActions.push_back({ _submissionNo, std::move(func)});
+    if (fence)
+    {
+        fence->putActions(_submissionNo, _fenceActions);
+    }
+    _submissionNo++;
+    return vkQueueSubmit(_vkQueue, 1, &submitInfo, fence ? fence->vk() : VK_NULL_HANDLE);
 }


robertosfield:
I have just done a first pass code review and it's not in a state that's ready to merge and I think heads off in the wrong direction for how all these classes should interact. My gut reaction is there has a be a cleaner way of doing what you require.  

I have been working flat out on other work this week so haven't had a chance to look into the topic. Unfortunately I'll likely be busy for the next two days as well, but if I have some time I'll look into the topic.  How easy is it to reproduce the performance bottleneck you are seeing with vsgCs?

timoore:
> I have just done a first pass code review and it's not in a state that's ready to merge and I think heads off in the wrong direction for how all these classes should interact. My gut reaction is there has a be a cleaner way of doing what you require.

I agree about the readiness to merge because there are some things in the code I touched that I don't fully understand, such as what's going on with acceleration structures and the dependent semaphore stuff that's currently in Fence. I'm not quite with you on the relative cleanliness of this approach. This PR delivers powerful semantics in only 100 lines of code: a strong happens-after relation between commands in a batch and code on the CPU, without having to do more waiting on the CPU than is already necessary to present an image to the window system. I'm a little uncomfortable that the Fence object is not tied more closely to the Queue in this approach.

I'm also happy to have shown that the compile operation is much faster than I had ever hoped. The Cesium terrain tiles and images are not enormous, but 500 microseconds to get them ready for the GPU? Yes please. I think it would be very limiting to keep a scheme that often results in a multi-frame delay in preparing paged data when the underlying operation is so fast.

> 
> I have been working flat out on other work this week so haven't had a chance to look into the topic. Unfortunately I'll likely be busy for the next two days as well, but if I have some time I'll look into the topic. How easy is it to reproduce the performance bottleneck you are seeing with vsgCs?

It's easy enough to observe if you are tracing vsgCs :sunglasses: I'm not sure how to observe it otherwise. The program feels snappier to me with this change, but it's easy to overlook the problem when loading data over a network. I think the before / after difference would be more noticeable when reading data from disk. There's a storage cache option in vsgCs, but it tends to blow up after a few minutes of use; I haven't had a chance to diagnose the problem.

robertosfield:
At this point I don't want to think too much about compile side as it's complex topic once you introduce threading and the need for synchronization.  It may be that changing vsg::Fence and vsg::Semaphore  + vsg::Queue may be appropriate as well changes to vsg::CompileTraversal/vsg::Context/vsg::ShaderCompiler/vsg::TransferTask.  To get it right requires a bit of Vulkan Mild Meld ;-)

My plan is clear my inbox of other VSG support + VSG client tasks in the first half of this week then for the second half of this week just dive completely into the compile topic.  This way I can concentrate fully without having several background processes bubbling away.

timoore:
> @timoore I am pleased to say that I've finally clear my task list and am now diving into the topic of refactoring compile to make it more efficient.
> 
> One of the key aspects to the work will be collecting a set of use cases that I can use as benchmarks for profiling the existing codebase and any modifications to it. I will instrument the key bits of compilation and the database pager so I have an objective measure of various costs.

If you don't already have a favorite profiling tool in your quiver, I do recommend [Tracy](https://github.com/wolfpld/tracy). I've integrated it into the vsgCs build as an option and have instrumented interesting parts of the code.
> 

> For my use cases for testing it would be helpful to have one, or more, test cases using vsgCs.. Does it support caching of tile yet? Caching of tiles would lead to more tiles being loaded rapidly and stressing the non IO side of the paging/compile. So could you suggest how best to test with vsgCs.

vsgCs does support using an sqlite cache that's supplied by Cesium native, but it was blowing up the last time I tried it. I'm traveling today and can't look at the cache code right now, but I've made recent improvements in the http/curl side which makes the http transport a lot faster. Just loading up one of the world files e.g. osm-buildings.json, produces "bursty" loading behavior where several terrain tiles are all loaded and compiled at the same time.

robertosfield:
I have just updated vsgCs and tested with:

`worldviewer --ion-token-file ~/.cesiumiontoken --world-file tests/osm-buildings.json
`

Paging in is still really slow for me.  Does the compile changes in this PR improve upon this?

The new camera manipulator is a bit jolty when zooming in.  Is this still in dev?

timoore:
> I have just updated vsgCs and tested with:
> 
> `worldviewer --ion-token-file ~/.cesiumiontoken --world-file tests/osm-buildings.json `
> 
> Paging in is still really slow for me. Does the compile changes in this PR improve upon this?
> 

Yes, subjectively. Objectively the time from the beginning of a tile's compilation until it is ready has been reduced by 100s of milliseconds in some cases.

> The new camera manipulator is a bit jolty when zooming in. Is this still in dev?

Yup.

robertosfield:
I have now merged this PR as a timoore-nonblocking-compile branch:

https://github.com/vsg-dev/VulkanSceneGraph/tree/timoore-nonblocking-compile

I am currently testing with it and will use as a reference for the various tests I'll do this week.

robertosfield:
When running with this branch and enabling debug layer I'm getting the Vulkan debug layer errors.

Test:
`worldviewer --ion-token-file ~/.cesiumiontoken --world-file tests/osm-buildings.json -d`


Errors:
```
UNASSIGNED-Threading-MultipleThreads(ERROR / SPEC): msgNum: 337425955 - Validation Error: [ UNASSIGNED-Threading-MultipleThreads ] Object 0: handle = 0x1d000000001d, type = VK_OBJECT_TYPE_COMMAND_POOL; | MessageID = 0x141cb623 | THREADING ERROR : vkFreeCommandBuffers(): object of type VkCommandPool is simultaneously used in thread 0x7effcd7fb640 and thread 0x7effee810180
    Objects: 1
        [0] 0x1d000000001d, type: 25, name: NULL
UNASSIGNED-Threading-MultipleThreads(ERROR / SPEC): msgNum: 337425955 - Validation Error: [ UNASSIGNED-Threading-MultipleThreads ] Object 0: handle = 0x1d000000001d, type = VK_OBJECT_TYPE_COMMAND_POOL; | MessageID = 0x141cb623 | THREADING ERROR : vkCmdBlitImage(): object of type VkCommandPool is simultaneously used in thread 0x7effee810180 and thread 0x7effcd7fb640
    Objects: 1
        [0] 0x1d000000001d, type: 25, name: NULL
UNASSIGNED-Threading-MultipleThreads(ERROR / SPEC): msgNum: 337425955 - Validation Error: [ UNASSIGNED-Threading-MultipleThreads ] Object 0: handle = 0x1d000000001d, type = VK_OBJECT_TYPE_COMMAND_POOL; | MessageID = 0x141cb623 | THREADING ERROR : vkFreeCommandBuffers(): object of type VkCommandPool is simultaneously used in thread 0x7effcd7fb640 and thread 0x7effee810180
    Objects: 1
        [0] 0x1d000000001d, type: 25, name: NULL
UNASSIGNED-Threading-MultipleThreads(ERROR / SPEC): msgNum: 337425955 - Validation Error: [ UNASSIGNED-Threading-MultipleThreads ] Object 0: handle = 0x1d000000001d, type = VK_OBJECT_TYPE_COMMAND_POOL; | MessageID = 0x141cb623 | THREADING ERROR : vkCmdPipelineBarrier(): object of type VkCommandPool is simultaneously used in thread 0x7effee810180 and thread 0x7effcdffc640
    Objects: 1

```

timoore:
Yeah, I just noticed those too. Seems to be coming from my "minicompile" in the main thread. I will take a closer look this evening.

I recently discovered that you can use the vkconfig program to force a breakpoint trap when the validation layer logs something. That's been useful for trying to suss this one out.

robertosfield:
Reverting to VSG master the debug errors go away.

--

I will now add some timing stats into the various compile traversal parts so I can get a handle on the relative costs of database paging/compile.

robertosfield:
@timoore I have added a very basic timer class into the CompileManager.cpp and then used this in the CompileManager::compile(..) method and got some interesting results.  

If I run:

`vsgtiledatabase --bing-maps`

And press '1' to take the camera to Buckingham Palace it outputs the console a whole stream of output like:

```
info: CompileManager::compile()
info:    object->accept(collectRequirements) 0.006
info:    compileTraversal = compileTraversals->take_when_available() 97.2242
info:    context->reserve(requirements); 0.02564
info:    object->accept(*compileTraversal) 0.199289
info:    compileTraversal->record() 0.09983
info:    compileTraversal->waitForCompletion() 49.6378
info:    after compileTraversals->add(compileTraversal); 0.00395
info: CompileManager::compile()
info:    object->accept(collectRequirements) 0.00968
info:    compileTraversal = compileTraversals->take_when_available() 93.6192
info:    context->reserve(requirements); 0.02776
info:    object->accept(*compileTraversal) 2.32825
info:    compileTraversal->record() 0.09285
info:    compileTraversal->waitForCompletion() 47.51
info:    after compileTraversals->add(compileTraversal); 0.00833
info: CompileManager::compile()
info:    object->accept(collectRequirements) 0.00479
info:    compileTraversal = compileTraversals->take_when_available() 1514.13
info:    context->reserve(requirements); 0.0258
info:    object->accept(*compileTraversal) 1.15698
info:    compileTraversal->record() 0.06611
info:    compileTraversal->waitForCompletion() 48.8313
info:    after compileTraversals->add(compileTraversal); 0.00112
```

The big ones are the compileTraversal = compileTraversals->take_when_available()  and compileTraversal->waitForCompletion().  

There are multiple database paging threads contending on the CompileManager::compile() call which explains why the take_when_available() is so high, and the context is also down to waitForCompletion() taking 3 frames.  Which is rather suspicious, so I switched off vsync with:

`vsgtiledatabase --bing-maps --IMMEDIATE
`

Then on pressing '1' I get:

```
info: CompileManager::compile()
info:    object->accept(collectRequirements) 0.00265
info:    compileTraversal = compileTraversals->take_when_available() 0.0003
info:    context->reserve(requirements); 0.00667
info:    object->accept(*compileTraversal) 0.19846
info:    compileTraversal->record() 0.060259
info:    compileTraversal->waitForCompletion() 0.74041
info:    after compileTraversals->add(compileTraversal); 0.00022
info: CompileManager::compile()
info:    object->accept(collectRequirements) 0.00512
info:    compileTraversal = compileTraversals->take_when_available() 0.00047
info:    context->reserve(requirements); 0.0198
info:    object->accept(*compileTraversal) 0.12463
info:    compileTraversal->record() 0.06706
info:    compileTraversal->waitForCompletion() 1.19014
info:    after compileTraversals->add(compileTraversal); 0.00028
info: CompileManager::compile()
info:    object->accept(collectRequirements) 0.0043
info:    compileTraversal = compileTraversals->take_when_available() 0.00039
info:    context->reserve(requirements); 0.01858
info:    object->accept(*compileTraversal) 0.11886
info:    compileTraversal->record() 0.165039
info:    compileTraversal->waitForCompletion() 0.79521
info:    after compileTraversals->add(compileTraversal); 0.00059
```

Which is quite a bit better... and points to the fence in Context::waitForCompletion() being stalled by my NVIdia driver GPU tying the transfer data to the frames, which is really not what we'd want.  

If I run vsgCs's worldviewer with --IMMEDIATE it also exhibits the same behavior - dropping the waitForCompletion() to less than a ms from several frames worth.

So.... I need to look into why this is happens, perhaps the driver is serializing all the submissions in some way so that fences only get signaled once all the previously submissions have completed.  Ideally we'd want to decouple this so that the compilation is not held by rendering/vsync.

robertosfield:
I have checked in my hacky timer code, I'll rewrite something more general purpose later, for now I'm just throwing stuff together with the aim of learning how it works in practice and letting that inform what it should look in a final solution.

https://github.com/vsg-dev/VulkanSceneGraph/tree/CompileTimingStats

Even from this crude stats it's clear than something odd is going on with the compile fence being held up by the main rendering.  Resolving this will be my main focus now as it'll likely give us the most bang for our buck w.r.t making compile more efficient.

timoore:
From a rest stop on the autoroute....

On Tue, May 2, 2023, 4:28 PM Robert Osfield ***@***.***>
wrote:

> I have checked in my hacky timer code, I'll rewrite something more general
> purpose later, for now I'm just throwing stuff together with the aim of
> learning how it works in practice and letting that inform what it should
> look in a final solution.
>
> https://github.com/vsg-dev/VulkanSceneGraph/tree/CompileTimingStats
>
> Even from this crude stats it's clear than something odd is going on with
> the compile fence being held up by the main rendering. Resolving this will
> be my main focus now as it'll likely give us the most bang for our buck
> w.r.t making compile more efficient.
>
> —
>
That's exactly what I was seeing. Most of vsgCs' main loop time is spent
waiting on a fence because the rendering is not challenging. The compile's
wait has to get in line behind that fence wait.

robertosfield:
I have dived into the topic of queues and found that it's all rather tangled with vkPhysicalDevice/vkDevice setup.  The current incarnation in the VSG master is using the same queue for the graphics and transfer operations, and for the NVidia driver at least it looks like fences are signaled in submission order, which in the end means the vsync is holding everything up.

I have attempted to setup the VSG so that we have a queue for graphics and queue for transfer, to do this I've had to hack several places in the core VSG.  Now I think I have two queues but it's failing in compile step as we're using vkCmdBlitImage in the transfer and setup of the texture data.

I feel like I've fallen down the rabbit hole.  My brain is now mush so I'll quick for the day and hope that it all makes more sense after a sleep.

robertosfield:
It may be that we just need to use a semaphore to signal the data is ready to use and only use a fence to say that the resources can be released/reused.

A semaphore doesn't require a round trip between the CPU and GPU, but would mean that we'd have to merge data with the scene graph without knowing whether the semaphore might cause a frame break.  As the rendering of the scene graph happens after the subgraph is merged one would hope that the data is already transferred and semaphore is signaled 

The tricky part of going with this approach is allowing another compile traversal while the resources may still be actively used.  This brings us back to the multiple compile traversals or having multiple sets of resources.

timoore:
> When running with this branch and enabling debug layer I'm getting the Vulkan debug layer errors.
> 
> Test: `worldviewer --ion-token-file ~/.cesiumiontoken --world-file tests/osm-buildings.json -d`
> 
> Errors:
> 
> ```
> UNASSIGNED-Threading-MultipleThreads(ERROR / SPEC): msgNum: 337425955 - Validation Error: [ UNASSIGNED-Threading-MultipleThreads ] Object 0: handle = 0x1d000000001d, type = VK_OBJECT_TYPE_COMMAND_POOL; | MessageID = 0x141cb623 | THREADING ERROR : vkFreeCommandBuffers(): object of type VkCommandPool is simultaneously used in thread 0x7effcd7fb640 and thread 0x7effee810180
>     Objects: 1
>         [0] 0x1d000000001d, type: 25, name: NULL
> UNASSIGNED-Threading-MultipleThreads(ERROR / SPEC): msgNum: 337425955 - Validation Error: [ UNASSIGNED-Threading-MultipleThreads ] Object 0: handle = 0x1d000000001d, type = VK_OBJECT_TYPE_COMMAND_POOL; | MessageID = 0x141cb623 | THREADING ERROR : vkCmdBlitImage(): object of type VkCommandPool is simultaneously used in thread 0x7effee810180 and thread 0x7effcd7fb640
>     Objects: 1
>         [0] 0x1d000000001d, type: 25, name: NULL
> UNASSIGNED-Threading-MultipleThreads(ERROR / SPEC): msgNum: 337425955 - Validation Error: [ UNASSIGNED-Threading-MultipleThreads ] Object 0: handle = 0x1d000000001d, type = VK_OBJECT_TYPE_COMMAND_POOL; | MessageID = 0x141cb623 | THREADING ERROR : vkFreeCommandBuffers(): object of type VkCommandPool is simultaneously used in thread 0x7effcd7fb640 and thread 0x7effee810180
>     Objects: 1
>         [0] 0x1d000000001d, type: 25, name: NULL
> UNASSIGNED-Threading-MultipleThreads(ERROR / SPEC): msgNum: 337425955 - Validation Error: [ UNASSIGNED-Threading-MultipleThreads ] Object 0: handle = 0x1d000000001d, type = VK_OBJECT_TYPE_COMMAND_POOL; | MessageID = 0x141cb623 | THREADING ERROR : vkCmdPipelineBarrier(): object of type VkCommandPool is simultaneously used in thread 0x7effee810180 and thread 0x7effcdffc640
>     Objects: 1
> ```

I'm about to turn this into another issue, but for my information, what is VkSurfaceCapabilitiesKHR::minImageCount on your test machine? I think this is a VSG problem with respect to recycling that I try to work around in vsgCs.

robertosfield:
I have figured out how to create two graphics queues, one for rendering and another for compile and the difference in performance is ASTONISHING!!!!  I can't stress this enough.

This is the stats I'm now seeing for the same test case above, it's 10X and more faster.  Paged databases on the local disk now pull is so rapidly.

info: CompileManager::compile()
info:    object->accept(collectRequirements) 0.00516
info:    compileTraversal = compileTraversals->take_when_available() 0.00061
info:    context->reserve(requirements); 0.01718
info:    object->accept(*compileTraversal) 0.936918
info:    compileTraversal->record() 0.07146
info:    compileTraversal->waitForCompletion() 1.36142
info:    after compileTraversals->add(compileTraversal); 0.00103
info: CompileManager::compile()
info:    object->accept(collectRequirements) 0.00391
info:    compileTraversal = compileTraversals->take_when_available() 0.00039
info:    context->reserve(requirements); 0.01139
info:    object->accept(*compileTraversal) 0.1397
info:    compileTraversal->record() 0.08964
info:    compileTraversal->waitForCompletion() 2.41809
info:    after compileTraversals->add(compileTraversal); 0.00027
info: CompileManager::compile()
info:    object->accept(collectRequirements) 0.00379
info:    compileTraversal = compileTraversals->take_when_available() 0.00074
info:    context->reserve(requirements); 0.01161
info:    object->accept(*compileTraversal) 0.11735
info:    compileTraversal->record() 0.06547
info:    compileTraversal->waitForCompletion() 5.11096
info:    after compileTraversals->add(compileTraversal); 0.00069

robertosfield:
> > When running with this branch and enabling debug layer I'm getting the Vulkan debug layer errors.
> > Test: `worldviewer --ion-token-file ~/.cesiumiontoken --world-file tests/osm-buildings.json -d`
> > Errors:
> > ```
> > UNASSIGNED-Threading-MultipleThreads(ERROR / SPEC): msgNum: 337425955 - Validation Error: [ UNASSIGNED-Threading-MultipleThreads ] Object 0: handle = 0x1d000000001d, type = VK_OBJECT_TYPE_COMMAND_POOL; | MessageID = 0x141cb623 | THREADING ERROR : vkFreeCommandBuffers(): object of type VkCommandPool is simultaneously used in thread 0x7effcd7fb640 and thread 0x7effee810180
> >     Objects: 1
> >         [0] 0x1d000000001d, type: 25, name: NULL
> > UNASSIGNED-Threading-MultipleThreads(ERROR / SPEC): msgNum: 337425955 - Validation Error: [ UNASSIGNED-Threading-MultipleThreads ] Object 0: handle = 0x1d000000001d, type = VK_OBJECT_TYPE_COMMAND_POOL; | MessageID = 0x141cb623 | THREADING ERROR : vkCmdBlitImage(): object of type VkCommandPool is simultaneously used in thread 0x7effee810180 and thread 0x7effcd7fb640
> >     Objects: 1
> >         [0] 0x1d000000001d, type: 25, name: NULL
> > UNASSIGNED-Threading-MultipleThreads(ERROR / SPEC): msgNum: 337425955 - Validation Error: [ UNASSIGNED-Threading-MultipleThreads ] Object 0: handle = 0x1d000000001d, type = VK_OBJECT_TYPE_COMMAND_POOL; | MessageID = 0x141cb623 | THREADING ERROR : vkFreeCommandBuffers(): object of type VkCommandPool is simultaneously used in thread 0x7effcd7fb640 and thread 0x7effee810180
> >     Objects: 1
> >         [0] 0x1d000000001d, type: 25, name: NULL
> > UNASSIGNED-Threading-MultipleThreads(ERROR / SPEC): msgNum: 337425955 - Validation Error: [ UNASSIGNED-Threading-MultipleThreads ] Object 0: handle = 0x1d000000001d, type = VK_OBJECT_TYPE_COMMAND_POOL; | MessageID = 0x141cb623 | THREADING ERROR : vkCmdPipelineBarrier(): object of type VkCommandPool is simultaneously used in thread 0x7effee810180 and thread 0x7effcdffc640
> >     Objects: 1
> > ```
> 
> I'm about to turn this into another issue, but for my information, what is VkSurfaceCapabilitiesKHR::minImageCount on your test machine? I think this is a VSG problem with respect to recycling that I try to work around in vsgCs.

Could this be an issue with deleting/recycling resources from the scene graph while they are still queue up or concurrently rendering on the GPU simply because double/triple buffering leads to several frames being racked up.

If so then the solution would be to schedule deletion/reuse of resources for as many frames as one has in the swap chain.  This would require some mechanism for queue resources for deletion/reuse to hold things back from immediate deletion/reuse.

robertosfield:
I have hacked the support for creating a second graphics queue to test the concept and as it works I'll now create another branch and try to come up with a coherent interface for it.

timoore:
> > > When running with this branch and enabling debug layer I'm getting the Vulkan debug layer errors.
> > > UNASSIGNED-Threading-MultipleThreads(ERROR / SPEC): msgNum: 337425955 - Validation Error: [ UNASSIGNED-Threading-MultipleThreads ] Object 0: handle = 0x1d000000001d, type = VK_OBJECT_TYPE_COMMAND_POOL; | MessageID = 0x141cb623 | THREADING ERROR : vkFreeCommandBuffers(): object of type VkCommandPool is simultaneously used in thread 0x7effcd7fb640 and thread 0x7effee810180
> > >     Objects: 1
> > >         [0] 0x1d000000001d, type: 25, name: NULL
> > > UNASSIGNED-Threading-MultipleThreads(ERROR / SPEC): msgNum: 337425955 - Validation Error: [ UNASSIGNED-Threading-MultipleThreads ] Object 0: handle = 0x1d000000001d, type = VK_OBJECT_TYPE_COMMAND_POOL; | MessageID = 0x141cb623 | THREADING ERROR : vkCmdPipelineBarrier(): object of type VkCommandPool is simultaneously used in thread 0x7effee810180 and thread 0x7effcdffc640
> > >     Objects: 1
> > > ```
> > 
> > 
> > I'm about to turn this into another issue, but for my information, what is VkSurfaceCapabilitiesKHR::minImageCount on your test machine? I think this is a VSG problem with respect to recycling that I try to work around in vsgCs.
> 
> Could this be an issue with deleting/recycling resources from the scene graph while they are still queue up or concurrently rendering on the GPU simply because double/triple buffering leads to several frames being racked up.
> 

Yes, that's exactly it.
> If so then the solution would be to schedule deletion/reuse of resources for as many frames as one has in the swap chain. This would require some mechanism for queue resources for deletion/reuse to hold things back from immediate deletion/reuse.

From the beginning of the vsgCs project I've done that, with a DeletionQueue that keeps freed-up tiles and images around for two frames before deleting. For some reason it's not always working when the compiler doesn't wait around for 100s of ms. I don't see the errors on my desktop Nvidia machine where minImageCount=2, but I do see it with AMD integrated graphics where minImageCount=3. Hence my question.

timoore:
> I have hacked the support for creating a second graphics queue to test the concept and as it works I'll now create another branch and try to come up with a coherent interface for it.

Do you have a plan for when there's only 1 graphics queue available (e.g., on AMD RADV RENOIR)?

robertosfield:
I have just checked into VSG master my changes to use a separate vk/vsg::Queue for the CompileTraversal:

https://github.com/vsg-dev/VulkanSceneGraph/commit/06cdbdaf33c969495d2d8f85d27b58db6839d078

This also requires an update to vsgExamples so that vsgcompute example sets up the vsg::CompileTraversal in a way that is consistent with this new approach.

I will be making a VulkanSceneGraph-1.0.6 release this week to include these changes.

I believe this improvement addresses the performance issue that this PR attempted to resolve so will now close this PR.

timoore:
Two problems with this:
It fails when there is only one graphics queue available, so VSG master is broken on one of my main development platforms.

This approach is 30-40% slower than submitting to a single queue and not blocking, as measured by Tracy.

robertosfield:
What hardware drivers one support graphics queue?

What happens when it isn't available?

Comparisons against a something that was generating errors isn't valid.

robertosfield:
I have just tested with my Intel laptop with Ubuntu 22.04 and it works fine.  I have did all the original testing on my Kubuntu 22.04 desktop that presently has a Geforce RTX 2060.

timoore:
> What hardware drivers one support graphics queue?
> 

AMD RADV RENOIR. This is the integrated graphics in my Dell Inspiron laptop. I can't find the number of graphics queues in the big Vulkan database, but I think this is a common scenario for integrated AMD graphics.

> What happens when it isn't available?
> 

To start with:
    Objects: 1
        [0] 0x120aed0, type: 1, name: NULL
VUID-VkDeviceQueueCreateInfo-queueCount-00382(ERROR / SPEC): msgNum: 1787754462 - Validation Error: [ VUID-VkDeviceQueueCreateInfo-queueCount-00382 ] Object 0: handle = 0x1c6d110, type = VK_OBJECT_TYPE_PHYSICAL_DEVICE; | MessageID = 0x6a8ef7de | vkCreateDevice: pCreateInfo->pQueueCreateInfos[0].queueCount (=2) is not less than or equal to available queue count for this pCreateInfo->pQueueCreateInfos[0].queueFamilyIndex} (=0) obtained previously from vkGetPhysicalDeviceQueueFamilyProperties (i.e. is not less than or equal to 1). The Vulkan spec states: queueCount must be less than or equal to the queueCount member of the VkQueueFamilyProperties structure, as returned by vkGetPhysicalDeviceQueueFamilyProperties in the pQueueFamilyProperties[queueFamilyIndex] (https://vulkan.lunarg.com/doc/view/1.3.239.0/linux/1.3-extensions/vkspec.html#VUID-VkDeviceQueueCreateInfo-queueCount-00382)

Then:
amdgpu: The CS has been rejected, see dmesg for more information (-22).
(Never seen that before!)

And then a spew of other errors, no VSG compiling.

> Comparisons against a something that was generating errors isn't valid.

Are you referring to the errors caused by VSG recycling? I wasn't seeing those on the Nvidia 1070 where I did the comparison. I can't do any comparison on my AMD machine at the moment without reverting.

timoore:
> I have just tested with my Intel laptop with Ubuntu 22.04 and it works fine. I have did all the original testing on my Kubuntu 22.04 desktop that presently has a Geforce RTX 2060.

What is the queue count for queue family 0?

robertosfield:
On Wed, 3 May 2023 at 15:12, Tim Moore ***@***.***> wrote:

> What hardware drivers one support graphics queue?
>
> AMD RADV RENOIR.Message ID: <vsg-dev/VulkanSceneGraph/pull/786/c1533102790
> @github.com>
>
I will pull out my 2060 cards and test with the AMD5700G integrated GPU to
see what happens. It may be that we have to check for support then fallback
to using a single queue on those systems.

robertosfield:
On Wed, 3 May 2023 at 15:15, Tim Moore ***@***.***> wrote:

> I have just tested with my Intel laptop with Ubuntu 22.04 and it works
> fine. I have did all the original testing on my Kubuntu 22.04 desktop that
> presently has a Geforce RTX 2060.
>
> What is the queue count for queue family 0?
>
There isn't a check in the VSG for this yet, it just sets up the desired
queue settings in vsg::Device constructor.

I will look up what Vulkan has for reporting the number of queue supported
by the hardware.

robertosfield:
Testing the QueueCountCheck branch on my Intel laptop I get the following:

$ vsgdraw -d
info: vsg::Device::Device()
info:    queueFamilyProperties: queueFlags = 7, queueCount = 1

And when enabling Vulkan debug layer I also get:

VUID-VkDeviceQueueCreateInfo-queueCount-00382(ERROR / SPEC): msgNum: 1787754462 - Validation Error: [ VUID-VkDeviceQueueCreateInfo-queueCount-00382 ] Object 0: handle = 0x557d5643a4a0, type = VK_OBJECT_TYPE_PHYSICAL_DEVICE; | MessageID = 0x6a8ef7de | vkCreateDevice: pCreateInfo->pQueueCreateInfos[0].queueCount (=2) is not less than or equal to available queue count for this pCreateInfo->pQueueCreateInfos[0].queueFamilyIndex} (=0) obtained previously from vkGetPhysicalDeviceQueueFamilyProperties (i.e. is not less than or equal to 1). The Vulkan spec states: queueCount must be less than or equal to the queueCount member of the VkQueueFamilyProperties structure, as returned by vkGetPhysicalDeviceQueueFamilyProperties in the pQueueFamilyProperties[queueFamilyIndex] (https://vulkan.lunarg.com/doc/view/1.2.182.0/linux/1.2-extensions/vkspec.html#VUID-VkDeviceQueueCreateInfo-queueCount-00382)
    Objects: 1
        [0] 0x557d5643a4a0, type: 2, name: NULL

I will need to add some runtime checks with fallback to using just one queue.  Ideally we want to be using separate queues where possible as this should enable better multi-threading of work.

robertosfield:
I have now pulled out my Geforce 2060 so that I'm now using my AMD5700G integrated graphics, when I run vsgdraw -d I see:

$ vsgdraw -d
info: vsg::Device::Device()
info:    queueFamilyProperties: queueFlags = 15, queueCount = 1
info:    queueFamilyProperties: queueFlags = 14, queueCount = 4
VUID-VkDeviceQueueCreateInfo-queueCount-00382(ERROR / SPEC): msgNum: 1787754462 - Validation Error: [ VUID-VkDeviceQueueCreateInfo-queueCount-00382 ] Object 0: handle = 0x562258cbf800, type = VK_OBJECT_TYPE_PHYSICAL_DEVICE; | MessageID = 0x6a8ef7de | vkCreateDevice: pCreateInfo->pQueueCreateInfos[0].queueCount (=2) is not less than or equal to available queue count for this pCreateInfo->pQueueCreateInfos[0].queueFamilyIndex} (=0) obtained previously from vkGetPhysicalDeviceQueueFamilyProperties (i.e. is not less than or equal to 1). The Vulkan spec states: queueCount must be less than or equal to the queueCount member of the VkQueueFamilyProperties structure, as returned by vkGetPhysicalDeviceQueueFamilyProperties in the pQueueFamilyProperties[queueFamilyIndex] (https://www.khronos.org/registry/vulkan/specs/1.1-extensions/html/vkspec.html#VUID-VkDeviceQueueCreateInfo-queueCount-00382)
    Objects: 1
        [0] 0x562258cbf800, type: 2, name: NULL

So it looks like we have multiple queues available, but only one graphics queue.

This makes me wonder if we should be slitting up the compile traversal into:

1. Compile Shaders to SPRIR-V
2. Create Vulkan objects
3. Transfer data to GPU
4. Run any image blits

1&2 don't require any submissions
3 requires at least a transfer queue
4 requires a graphics queue

If a compile traversal doesn't require 3 or 4 then is nothing in them that requires a submission to any queue so could return immediately.

Could we differ type 4. command to the main traversal?   Could this be part of TransferTask?

Could we move all of 3 & 4 to the TransferTask?

--

For now I need to refactor the queue set up to not allocate and use queues that aren't supported so we don't get the errors.

Later we can resolve how to handle the different types of tasks carried about by the CompileTraversal/CompileManager in a more coherent way.

timoore:
> Testing the QueueCountCheck branch on my Intel laptop I get the following:
> 
> $ vsgdraw -d info: vsg::Device::Device() info: queueFamilyProperties: queueFlags = 7, queueCount = 1
> 
> And when enabling Vulkan debug layer I also get:
> 
> VUID-VkDeviceQueueCreateInfo-queueCount-00382(ERROR / SPEC): msgNum: 1787754462 - Validation Error: [ VUID-V
...
> 
> I will need to add some runtime checks with fallback to using just one queue. Ideally we want to be using separate queues where possible as this should enable better multi-threading of work.

The best performance, by a significant amount, comes from using a single graphics queue for compilation and not waiting on the submits on the CPU. It's my understanding that little, if any, hardware supports multiple command processors per GPU, so that the multiple graphics queues are a driver construct.

Transfer queues are usually backed by real hardware. I'm all for using a transfer queue where possible, but that's a different problem. It will be a shame if the fall-back does the same waiting that is there now, so that single graphics queue systems will have totally unnecessary compilation times.

I've checked in a fix in vsgCs for the various use / deletion validation errors; I'd be curious if you are still seeing that problem.

timoore:
> 
> So it looks like we have multiple queues available, but only one graphics queue.
> 
> This makes me wonder if we should be slitting up the compile traversal into:
> 
> 1. Compile Shaders to SPRIR-V
> 2. Create Vulkan objects
> 3. Transfer data to GPU
> 4. Run any image blits
> 
> 1&2 don't require any submissions 3 requires at least a transfer queue 4 requires a graphics queue
> 
> If a compile traversal doesn't require 3 or 4 then is nothing in them that requires a submission to any queue so could return immediately.
> 
> Could we differ type 4. command to the main traversal? Could this be part of TransferTask?
> 
> Could we move all of 3 & 4 to the TransferTask?
> 

I think you're on to something here. Another part of the puzzle is doing something with the Vulkan pipeline cache.

As I mentioned earlier, the expense of creating pipelines and GPU objects is a huge problem in the industry right now. Having a fast, multithreaded solution would be a very cool capability.

robertosfield:
I have refactored the Device::Device() constructor so that it automatically detects a request for more queues than their are available and a change to Device::getQueue(..) so that it returns the closest queue that matches what queue has been asked for.  I have also move the queue allocation into the Device constructor as I think it makes sense to just do this together as all the settings are available at his point.  

I still have debug code in place to report the counts and the fallback, the output looks like:

$ vsgtiledatabase --osm
info: vsg::Device::Device()
info:    queueFamilyProperties: queueFlags = 15, queueCount = 1
info:    queueFamilyProperties: queueFlags = 14, queueCount = 4
info:    available count 1, queueSetting.queuePiorities.size() = 2
info: Device::Device() creation failed to create request queueCount.
info: Device::getQueue(0, 1) failled back to next closest.
info: Device::getQueue(0, 1) failled back to next closest.

I will add the queue count output to the vsgdeviceselection example and removing it from the Device constructor as it was only there for debug purposes.

With this change we now can use the extra graphics queue and get the significantly better paging performance where the hardware supports it.  I consider this just stepping stone to making compile more efficient, obviously lots more needs to happen to optimize different parts of the process and take advantage of different hardware capabilities better.

robertosfield:
I have merged my fixes so that vsg::Device::Device() now clamps the requested number of queues to the available resources, and uses a fallback in Device::getQueue(..) so that everything continues to work even when one asks for the impossible :-)

The commit is: 523b28a61565d90c7c2bef752a6b4a49e5ed36a8

I have tested this out on my Intel laptop and my AMD5700G desktop using the integrated GPU in both cases and both now work without any Vulkan debug errors.  I have also updated the CompileTiming Stats branch to use this new code and run it on these two systems and found that neither block like my desktop blockis on the Context::waitForCompletion() like my NVidia hardware/driver does.

So... it seems that not having the second graphics queue for compile is less critical to AMD and Intel hardware than it is on NVidia, and it's the second graphics queue that evens the playing field between them.  There is still more work to do, but it's nice to see the NVidia performing a bit closer to where it should always have been when paging.

timoore:
> I have tested this out on my Intel laptop and my AMD5700G desktop using the integrated GPU in both cases and both now work without any Vulkan debug errors. I have also updated the CompileTiming Stats branch to use this new code and run it on these two systems and found that neither block like my desktop blockis on the Context::waitForCompletion() like my NVidia hardware/driver does.
> 

I confirm this on AMD Renoir. It's interesting: the AMD driver must block in vkAcquireNextImageKHR() while waiting for vsync or whatever, while on Nvidia we block in the vkQueueSubmit() call that waits on the semaphore returned by vkAcquireNextImageKHR(). So, the timing of the main loop is quite different in the two systems. Maybe that doesn't matter, maybe it does...

> So... it seems that not having the second graphics queue for compile is less critical to AMD and Intel hardware than it is on NVidia, and it's the second graphics queue that evens the playing field between them. There is still more work to do, but it's nice to see the NVidia performing a bit closer to where it should always have been when paging.

I'm happy for now :sunglasses: I think there is still performance gains to be had by understanding the sync guarantees provided by  semaphores, fences, and the new timeline semaphores, and knowing when one needs to block, or not.

PR #788, opened 2023-04-22
[ADDED] scope guard for taken compileTraversals

siystar:
# Pull Request Template

## Description

Added scope guard for compileTraversals taken by CompileManager to return them to the available queue on exception

Fixes # (issue)

## Type of change

## How Has This Been Tested?

Tested that shader compile errors encountered in compile thread no longer freeze the application

## Checklist:

### DIFF
--- a/src/vsg/app/CompileManager.cpp
+++ b/src/vsg/app/CompileManager.cpp
@@ -163,7 +163,22 @@ CompileResult CompileManager::compile(ref_ptr<Object> object, ContextSelectionFu
     result.earlyDynamicData = requirements.earlyDynamicData;
     result.lateDynamicData = requirements.lateDynamicData;
 
-    auto compileTraversal = compileTraversals->take_when_available();
+    struct LockCompileTraversal
+    {
+        CompileTraversals& compileTraversals;
+        ref_ptr<CompileTraversal> compileTraversal;
+        LockCompileTraversal(CompileTraversals& in_compileTraversals) : compileTraversals(in_compileTraversals)
+        {
+            compileTraversal = compileTraversals.take_when_available();
+        }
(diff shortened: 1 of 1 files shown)


timoore:
I like [gsl::finally](https://www.cppstories.com/2017/04/finalact/) for this sort of thing, instead of writing a new struct every time you need a new guard.

robertosfield:
Thanks for the fix. I have reviewed the changes and understand their intent but it adds complexity and looses clarity of what is happening in the code so I would like to look for a cleaner/more readable way of doing it  I will leave this PR up until I come up with an alternative.

robertosfield:
We should probably have try/catch around the compile traversal.

I am planning a proper review of vsg::CompileManager/CompileTraversal this week so can take this PR and the one from @timoore as reference for what issues need to encompassed in any changes I make.

robertosfield:
@siystar  I am currently looking at this PR.  

The if (contextSelection) branch also has a paired compileTraversal->contexts.swap(contexts); will also be sensitive to an exception being thrown.  So I think this should probably be tackled at the same time.

I will ponder what the cleanest way to do this will be.  A try catch is one route.  Using something equivalent to gsl::finally is another, though it'll have to be something standard rather than require another external dependency.

I should have something settle by the end of this afternoon.

robertosfield:
I have implement a try/catch in the CompileManger::compile(..) method:

https://github.com/vsg-dev/VulkanSceneGraph/tree/CompileCatchException

I have also added a CompileResult::message string so we can pass back a message when the result value.  This is just a first step with refactors of CompielManager that I think is needed, but I think it should be sufficient for addressing the issue that is the topic of this PR so I'm thinking this PR is now moot.

PR #800, opened 2023-05-05
Use Mask instead of uint32_t

timoore:
A couple of uses were missed in the change from uint32_t to vsg::Mask (uint64_t) in Switch nodes, etc.


robertosfield:
Well spotted. Surprised the compilers didn't shout about these.

PR #811, opened 2023-05-13
[ADDED] renderPass to RenderGraph

siystar:
# Pull Request Template

## Description

Added renderPass member to RenderGraph to support different renderPasses operating on the same framebuffer or window.

## Type of change

## How Has This Been Tested?

Tested by running 3D scene render pass, storing the rendered image for screen-space raytracing purposes, then running GUI render pass on top of the 3D scene

## Checklist:

### DIFF
--- a/include/vsg/app/RenderGraph.h
+++ b/include/vsg/app/RenderGraph.h
@@ -41,7 +41,7 @@ namespace vsg
         ref_ptr<Framebuffer> framebuffer;
         ref_ptr<Window> window;
 
-        /// RenderPass to use passed to the vkCmdBeginRenderPass, either obtained from which of the framebuffer or window are active
+        /// RenderPass to use passed to the vkCmdBeginRenderPass, if renderPass is set it takes precedence, if not then either obtained from which of the framebuffer or window are active
         RenderPass* getRenderPass();
 
         /// Get the Exten2D of the attached Framebuffer or Window.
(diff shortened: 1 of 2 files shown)


robertosfield:
I have done a code review and it makes worse something I've been unhappy about since the addition of support for using the Framebuffer as an alternate source for the renderPass - rather than having two possible sources of RenderPass setting this PR means we'll have three.  This duplication is fine for an expert user who has reviewed the code and figured out in which circumstances which object is used as a the source, but new users they'll be even more confused - they might attempt to attach window, a framebuffer and a renderPass as that's what the public interface suggests is supported.

One way I considered for resolving this was to have a RenderGraph::renderPassSource member i.e.

   // renderPassSource; provides the source of the RenderPass to use, it can be a window, framebuffer or a renderPass
   ref_ptr<Object> renderPassSource;

The in the RenderGraph::getRenderPass() method it could do a cast to RenderPasss, Window or Framebuffer to see which one is appropriate to use.

robertosfield:
I have started to experiment with renderPassSource but hit upon places where the window/framebuffer resources are required.

So the duality lies with the source of the framebuffer and the render pass.  Urgghjh...

OK, so I'm now wondering if there should be a framebufferSource which can be either a Window or Framebuffer, and the separate renderPass member.

To avoid the potential confusing with who provides the active renderPass  I'm now inclining towards having a setFramebufferSource(..) method that takes either a Window or a Framebuffer that sets the renderPass for users, or we require users to explicitly set the renderPass member.

robertosfield:
I haven't updated vsgExamples to work with these changes yet, and haven't tested other than to get them to compile, I'll finish up tomorrow.  I'm sharing now so you know the alternative I'm considering:

https://github.com/vsg-dev/VulkanSceneGraph/tree/RenderGraphRefactor

robertosfield:
I have tried a couple of different approaches but couldn't maintain backwards compatibility so I've decided to merge this PR, but applied a small revision to avoid setting renderPassInfo.renderPass multiple times. With commit 634a10cdba8478e64383858f702054952b890196.

I'll close this PR now as it's merged with master via the branch I merged earlier.

PR #835, opened 2023-05-27
Filter repeated keypress events when holding keys down on windows

jamierobertson1:
# Pull Request Template

## Description

Holding a key down on windows generates repeated keydown events after a second or so. This change filters out key press events by checking if their last state was also down (using the "previous key state", bit 30) 

[https://learn.microsoft.com/en-us/windows/win32/inputdev/wm-keydown]

Note that the "repeat count" field isn't much help for this as is not cumulative when multiple messages are sent when a key is held down. 

Fixes # (issue)
Keyboard functionality in trackball manipulator on windows not working properly when holding down alpha numeric keys.

## Type of change

- [x] Bug fix (non-breaking change which fixes an issues

## How Has This Been Tested?

- Tested trackball manipulator on vsgviewer example.

**Test Configuration**:

Windows 10

### DIFF
--- a/src/vsg/platform/win32/Win32_Window.cpp
+++ b/src/vsg/platform/win32/Win32_Window.cpp
@@ -616,8 +616,8 @@ LRESULT Win32_Window::handleWin32Messages(UINT msg, WPARAM wParam, LPARAM lParam
         if (_keyboard->getKeySymbol(wParam, lParam, keySymbol, modifiedKeySymbol, keyModifier))
         {
             int32_t repeatCount = (lParam & 0xffff);
-            bool previousKeyDown = (lParam & 0x40000000);
-            if (!previousKeyDown)
+            bool previouslyKeyDown = (lParam & 0x40000000);
+            if (!previouslyKeyDown)
                 bufferedEvents.emplace_back(vsg::KeyPressEvent::create(this, event_time, keySymbol, modifiedKeySymbol, keyModifier, repeatCount));
         }
         break;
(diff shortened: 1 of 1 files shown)


vsg-dev:
Ideally we want all platforms to produce the same behavior rather then make them diverge even further, this PR would make things diverge further.

Under Xcb key repeats are presented as an initial KeyPress,, then a KeyRelease followed by a KeyPress which have the same time signature .  With the new vsg::Keyboard class that enables tracking which keys are pressed and how long they are pressed is written to match these KeyRelease/KeyPress pairs that signify a key repeat.

A key repeat is useful to know about even if the Xcb/X11 approach is a bit clunky, and under Windows the code doesn't emulate this but does at least is intended to report repeated key presses.  I didn't write the Windows event handling code and I don't have a Windows system to test so have to rely upon the community to test/refine the Windows implementation.  I don't know the current status of what happens under Windows, reading the code it looks like existing code come closer to allow apps to track key repeats rather than this PR that explicitly treats to eat these events, thus prevent apps from even doing something about key repeats.

Key repeats are important part of UI's for instance vsgImGui relies upon the VSG catching the windows events and passing them along to ImGui to enable it to respond to keyboard input in a sensible way.  This PR would step input widgets from responding to key repeats like we'd expect from a UI.

It may be best to post to the discussion forum to explaining what behavior you are seeing under Windows and what you are trying to achieve.  Of relevance may be my recent work on adding keyboard controls to vsg::Trackall which I explain in this forum thread:

https://github.com/vsg-dev/VulkanSceneGraph/discussions/801

jamierobertson1:
Fair enough, I never really thought that the implications if that one through. Will post to the forum. It was an attempt to get vsg::trackball key's to work in windows. It seems the difference in windows is there are no KeyRelease events when holding a key down.

PR #869, opened 2023-07-06
Descriptor set layouts based on shader set definitions

timoore:
This moves GraphicsPipelineConfigurator away from managing a single descriptor set plus another for view dependent state. The layout is generated using the shader set and defines. These changes are ported from vsgCs, with an extra effort for compatibility with exisiting uses of GraphicsPipelineConfigurator.

There are two objectives here:
* support multiple, arbitrary descriptor sets i.e., don't assume model data on 0 and lighting on 1. I use this in vsgCs.
* Support additional descriptors in vsgXchange by defining them in a custom shader set.

I haven't changed the assimp loadger in vsgXchange yet, pending feedback on these changes.

### DIFF
--- a/include/vsg/utils/ShaderSet.h
+++ b/include/vsg/utils/ShaderSet.h
@@ -135,6 +135,18 @@ namespace vsg
     };
     VSG_type_name(vsg::ShaderSet);
 
+    /**
+     * @brief Create a pipeline layout for some sets of a shader set.
+     *
+     * @param shaderSet the shader set
+     * @param defines Set of defines for assembling the bindings in the sets
+     * @param sets Number of sets to use, starting from 0. Defaults to all sets.
+     * @return the vsg::PipelineLayout object.
+     */
+    extern VSG_DECLSPEC ref_ptr<PipelineLayout> makePipelineLayout(ref_ptr<ShaderSet> shaderSet,
+                                                                   const std::set<std::string>& defines = {},
+                                                                   int sets = -1);
+
     /// create a ShaderSet for unlit, flat shaded rendering
     extern VSG_DECLSPEC ref_ptr<ShaderSet> createFlatShadedShaderSet(ref_ptr<const Options> options = {});
 
--- a/include/vsg/utils/GraphicsPipelineConfigurator.h
+++ b/include/vsg/utils/GraphicsPipelineConfigurator.h
@@ -73,6 +73,15 @@ namespace vsg
 
         void init();
 
+        std::set<std::string>& defines()
+        {
+            return shaderHints->defines;
+        }
+        const std::set<std::string>& defines() const
+        {
+            return shaderHints->defines;
+        }
+
         // setup by init()
         ref_ptr<DescriptorSetLayout> descriptorSetLayout;
         ref_ptr<PipelineLayout> layout;
--- a/include/vsg/utils/GraphicsPipelineConfigurator.h
+++ b/include/vsg/utils/GraphicsPipelineConfigurator.h
@@ -28,6 +28,8 @@ THE SOFTWARE IS PROVIDED "AS IS", WITHOUT WARRANTY OF ANY KIND, EXPRESS OR IMPLI
 #include <vsg/state/ViewportState.h>
 #include <vsg/utils/ShaderSet.h>
 
+#include <optional>
+
 namespace vsg
 {
 
@@ -83,14 +85,22 @@ namespace vsg
     class VSG_DECLSPEC DescriptorConfigurator : public vsg::Inherit<Object, DescriptorConfigurator>
     {
     public:
-        DescriptorConfigurator(ref_ptr<ShaderSet> in_shaderSet = {});
-
+        DescriptorConfigurator(ref_ptr<ShaderSet> in_shaderSet = {})
+            : shaderSet(in_shaderSet)
+        {
+        }
+        DescriptorConfigurator(uint32_t in_setNumber, vsg::ref_ptr<vsg::ShaderSet> in_shaderSet = {})
+            : shaderSet(in_shaderSet), setNumber(in_setNumber)
+        {
+        }
         ref_ptr<ShaderSet> shaderSet;
         bool blending = false;
         bool two_sided = false;
 
         bool assignTexture(const std::string& name, ref_ptr<Data> textureData = {}, ref_ptr<Sampler> sampler = {});
+        bool assignTexture(const std::string& name, const vsg::ImageInfoList& imageInfoList, uint32_t arrayElement = 0);
         bool assignUniform(const std::string& name, ref_ptr<Data> data = {});
+        bool assignUniform(const std::string& name, const vsg::BufferInfoList& bufferInfoList, uint32_t arrayElement = 0);
 
         // assign Descriptors to a DescriptorSet
         void init();
@@ -102,6 +112,7 @@ namespace vsg
 
         // filled in by init()
         ref_ptr<DescriptorSet> descriptorSet;
+        std::optional<uint32_t> setNumber;
     };
     VSG_type_name(vsg::DescriptorConfigurator);
 
--- a/src/vsg/utils/GraphicsPipelineConfigurator.cpp
+++ b/src/vsg/utils/GraphicsPipelineConfigurator.cpp
@@ -10,6 +10,7 @@ THE SOFTWARE IS PROVIDED "AS IS", WITHOUT WARRANTY OF ANY KIND, EXPRESS OR IMPLI
 
 </editor-fold> */
 
+#include <vsg/io/Logger.h>
 #include <vsg/io/Options.h>
 #include <vsg/utils/GraphicsPipelineConfigurator.h>
 
@@ -226,15 +227,21 @@ void GraphicsPipelineConfigurator::init()
 //
 // DescriptorConfigurator
 //
-DescriptorConfigurator::DescriptorConfigurator(ref_ptr<ShaderSet> in_shaderSet) :
-    shaderSet(in_shaderSet)
-{
-}
 
 bool DescriptorConfigurator::assignTexture(const std::string& name, ref_ptr<Data> textureData, ref_ptr<Sampler> sampler)
 {
     if (auto& textureBinding = shaderSet->getUniformBinding(name))
     {
+        if (!setNumber)
+        {
+            setNumber = textureBinding.set;
+        }
+        else if (textureBinding.set != *setNumber)
+        {
+            vsg::warn("Attempted binding texture ", name, " with set number ", textureBinding.set, " in set ",
+                  setNumber.value());
+            return false;
+        }
         // set up bindings
         if (!textureBinding.define.empty()) defines.insert(textureBinding.define);
         descriptorBindings.push_back(VkDescriptorSetLayoutBinding{textureBinding.binding, textureBinding.descriptorType, textureBinding.descriptorCount, textureBinding.stageFlags, nullptr});
@@ -253,11 +260,78 @@ bool DescriptorConfigurator::assignUniform(const std::string& name, ref_ptr<Data
 {
     if (auto& uniformBinding = shaderSet->getUniformBinding(name))
     {
+        if (!setNumber)
+        {
+            setNumber = uniformBinding.set;
+        }
+        else if (uniformBinding.set != *setNumber)
+        {
+            vsg::warn("Attempted binding uniform ", name, " with set number ", uniformBinding.set, " in set ",
+                      setNumber.value());
+            return false;
+        }
         // set up bindings
         if (!uniformBinding.define.empty()) defines.insert(uniformBinding.define);
         descriptorBindings.push_back(VkDescriptorSetLayoutBinding{uniformBinding.binding, uniformBinding.descriptorType, uniformBinding.descriptorCount, uniformBinding.stageFlags, nullptr});
 
         auto uniform = DescriptorBuffer::create(data ? data : uniformBinding.data, uniformBinding.binding);
+        descriptors.push_back(uniform);
+
+        return true;
+    }
+    return false;
+}
+
+bool DescriptorConfigurator::assignTexture(const std::string& name, const vsg::ImageInfoList& imageInfoList,
+                                              uint32_t arrayElement)
+{
+    if (auto& textureBinding = shaderSet->getUniformBinding(name))
+    {
+        if (!setNumber)
+        {
+            setNumber = textureBinding.set;
+        }
+        else if (textureBinding.set != *setNumber)
+        {
+            vsg::warn("Attempted binding texture ", name, " with set number ", textureBinding.set, " in set ",
+                  setNumber.value());
+            return false;
+        }
+        // set up bindings
+        if (!textureBinding.define.empty()) defines.insert(textureBinding.define);
+        descriptorBindings.push_back(VkDescriptorSetLayoutBinding{textureBinding.binding, textureBinding.descriptorType,
+                                                                  textureBinding.descriptorCount, textureBinding.stageFlags,
+                                                                  nullptr});
+
+        // create texture image and associated DescriptorSets and binding
+        auto texture = vsg::DescriptorImage::create(imageInfoList, textureBinding.binding, arrayElement,
+                                                    textureBinding.descriptorType);
+        descriptors.push_back(texture);
+        return true;
+    }
+    return false;
+}
+
+bool DescriptorConfigurator::assignUniform(const std::string& name, const vsg::BufferInfoList& bufferInfoList,
+                                              uint32_t arrayElement)
+{
+    if (auto& uniformBinding = shaderSet->getUniformBinding(name))
+    {
+        if (!setNumber)
+        {
+            setNumber = uniformBinding.set;
+        }
+        else if (uniformBinding.set != *setNumber)
+        {
+            vsg::warn("Attempted binding uniform ", name, " with set number ", uniformBinding.set, " in set ",
+                      setNumber.value());
+            return false;
+        }
+        // set up bindings
+        if (!uniformBinding.define.empty()) defines.insert(uniformBinding.define);
+        descriptorBindings.push_back(VkDescriptorSetLayoutBinding{uniformBinding.binding, uniformBinding.descriptorType, uniformBinding.descriptorCount, uniformBinding.stageFlags, nullptr});
+
+        auto uniform = vsg::DescriptorBuffer::create(bufferInfoList, uniformBinding.binding, arrayElement);
         descriptors.push_back(uniform);
 
         return true;
--- a/src/vsg/utils/GraphicsPipelineConfigurator.cpp
+++ b/src/vsg/utils/GraphicsPipelineConfigurator.cpp
@@ -192,21 +192,33 @@ int GraphicsPipelineConfigurator::compare(const Object& rhs_object) const
 
 void GraphicsPipelineConfigurator::init()
 {
-    if (!descriptorSetLayout)
-    {
-        descriptorSetLayout = vsg::DescriptorSetLayout::create(descriptorBindings);
-    }
-
     vsg::PushConstantRanges pushConstantRanges;
     for (auto& pcb : shaderSet->pushConstantRanges)
     {
         if (pcb.define.empty()) pushConstantRanges.push_back(pcb.range);
     }
 
-    vsg::DescriptorSetLayouts desriptorSetLayouts{descriptorSetLayout};
-    if (additionalDescriptorSetLayout) desriptorSetLayouts.push_back(additionalDescriptorSetLayout);
-
-    layout = vsg::PipelineLayout::create(desriptorSetLayouts, pushConstantRanges);
+    if (descriptorSetLayout)
+    {
+        // User has set descriptor set layout, so go with that
+       DescriptorSetLayouts descriptorSetLayouts{descriptorSetLayout};
+       if (additionalDescriptorSetLayout) descriptorSetLayouts.push_back(additionalDescriptorSetLayout);
+       layout = vsg::PipelineLayout::create(descriptorSetLayouts, pushConstantRanges);
+    }
+    else
+    {
+        layout = makePipelineLayout(shaderSet, defines());
+        // Populate descriptorSetLayout and additionalDescriptorSetLayout for code that is expecting
+        // it, though that should be deprecated.
+        if (!layout->setLayouts.empty())
+        {
+            descriptorSetLayout = layout->setLayouts[0];
+            if (layout->setLayouts.size() >= 2)
+            {
+                additionalDescriptorSetLayout = layout->setLayouts[1];
+            }
+        }
+    }
 
     GraphicsPipelineStates pipelineStates;
     if (colorBlendState) pipelineStates.push_back(colorBlendState);
--- a/src/vsg/utils/ShaderSet.cpp
+++ b/src/vsg/utils/ShaderSet.cpp
@@ -352,6 +352,60 @@ void ShaderSet::write(Output& output) const
     }
 }
 
+vsg::ref_ptr<vsg::PipelineLayout> vsg::makePipelineLayout(vsg::ref_ptr<ShaderSet> shaderSet,
+                                                     const std::set<std::string>& defines, int max_set)
+{
+    const auto& uniformBindings = shaderSet->uniformBindings;
+    uint32_t umax_set;
+    if (max_set >= 0)
+    {
+        umax_set = static_cast<uint32_t>(max_set);
+    }
+    else
+    {
+        auto maxItr = std::max_element(uniformBindings.begin(), uniformBindings.end(),
+                                       [](const auto& val1, const auto& val2)
+                                       {
+                                           return val1.set < val2.set;
+                                       });
+        umax_set = maxItr->set;
+    }
+    std::map<uint32_t, DescriptorSetLayoutBindings> sets;
+    for (auto& binding : uniformBindings)
+    {
+        if (umax_set >= binding.set)
+        {
+            if (binding.define.empty() || defines.find(binding.define) != defines.end())
+            {
+                sets[binding.set].push_back(VkDescriptorSetLayoutBinding{binding.binding, binding.descriptorType,
+                                                                         binding.descriptorCount, binding.stageFlags,
+                                                                         nullptr});
+            }
+        }
+    }
+    DescriptorSetLayouts descriptorSetLayouts;
+    for (uint32_t set = 0; set <= umax_set; ++set)
+    {
+        auto dslbItr = sets.find(set);
+        if (dslbItr != sets.end())
+        {
+            descriptorSetLayouts.push_back(vsg::DescriptorSetLayout::create(dslbItr->second));
+        }
+        else
+        {
+            descriptorSetLayouts.push_back(vsg::DescriptorSetLayout::create());
+        }
+    }
+    PushConstantRanges pushConstantRanges;
+    for (auto& pcb : shaderSet->pushConstantRanges)
+    {
+        if (pcb.define.empty()) pushConstantRanges.push_back(pcb.range);
+    }
+
+    auto pipelineLayout = vsg::PipelineLayout::create(descriptorSetLayouts, pushConstantRanges);
+    return pipelineLayout;
+}
+    
 ref_ptr<ShaderSet> vsg::createFlatShadedShaderSet(ref_ptr<const Options> options)
 {
     if (options)


include/vsg/utils/ShaderSet.h
```diff
@@ -135,6 +135,18 @@ namespace vsg
...
 
+    /**
+     * @brief Create a pipeline layout for some sets of a shader set.
+     *
+     * @param shaderSet the shader set
+     * @param defines Set of defines for assembling the bindings in the sets
+     * @param sets Number of sets to use, starting from 0. Defaults to all sets.
+     * @return the vsg::PipelineLayout object.
+     */
+    extern VSG_DECLSPEC ref_ptr<PipelineLayout> makePipelineLayout(ref_ptr<ShaderSet> shaderSet,
```
robertosfield, line 146: Why call it makePipelineLayout() when similar code use createSomething() and even the doxygen docs says create a pipeline layout?

timoore: Fair enough. Actually, I wasn't sure whether or not to make it a non-member function, but I prefer that when the function doesn't depend on any private class members.

robertosfield:
I have just done a first pass review and haven't yet figured out what the changes do/are meant to be used.

It 's end of working day so I'll return to this tomorrow.  Could you point me at the parts of vsgCs that utilize this functionality as this might help me figure out the intent behind the changes.

If it means keeping the code cleaner then I'm prepared to drop backwards compatibility, though as a first pass backwards compatibility is helpful.  At this point I'm really just trying to understand the shortfalls of the present code and what broadly additional functionality is required rather than specifics of the implementation.

FYI. I'm aware of number of shortfalls in the present ShaderSet/GraphicsPipelineConfigurator, I consider what is in VSG master is a first step at providing basic high level tool for describing shaders and the uniforms/textures that are meant to be used with them.  Using the code in real-world applications/libs is what shakes down a novel approach like this.

timoore:
> I have just done a first pass review and haven't yet figured out what the changes do/are meant to be used.
> 

The change to DescriptorConfigurator is just error checking, ensuring that textures and uniforms assigned to the configurator are bound in the same descriptor set in the ShaderSet.

I wrote this code in vsgCs to be able to, using GraphicsPipelineConfigurator and DescriptorConfigurator:
* bind the ViewDependent state on the descriptor set of my choice, i.e. set 0;
* bind the model data somewhere else; in the event, it goes on descriptor set 3;
* support more than one descriptor set

This will be well-suited to the use case of adding custom descriptors to a vsgXchange model: the pipeline layout is derived from the ShaderSet, not from the descriptors that vsgXchange accumulates and binds.

> It 's end of working day so I'll return to this tomorrow. Could you point me at the parts of vsgCs that utilize this functionality as this might help me figure out the intent behind the changes.

I wrote these as derived classes of the VSG classes in vsgCs, so the use is a little wonky in vsgCs master. I've just checked in the branch vsg-desc-set which uses the new VSG classes. They are still referred to by their old names: DescriptorSetConfigurator and MultisetPipelineConfigurator. The pipeline configurator is used in ModelBuilder.cpp.
> 
> If it means keeping the code cleaner then I'm prepared to drop backwards compatibility, though as a first pass backwards compatibility is helpful. At this point I'm really just trying to understand the shortfalls of the present code and what broadly additional functionality is required rather than specifics of the implementation.

The only code that is really for backwards compatibility is in GraphicsPipelineConfigurator::init(), which handles usages of descriptorSetLayout and additionalDescriptorSetLayout.

> 
> FYI. I'm aware of number of shortfalls in the present ShaderSet/GraphicsPipelineConfigurator, I consider what is in VSG master is a first step at providing basic high level tool for describing shaders and the uniforms/textures that are meant to be used with them. Using the code in real-world applications/libs is what shakes down a novel approach like this.

vsgCs is the real world :smile:

robertosfield:
Thanks for the additional details.  What I'm inclined to do is merge the changes as a branch then update vsgCs and vsgXchange with their own matched branches so they all work together.  Then we can review the changes with usage and allow vsgCs to be cleaned up without the workarounds.

robertosfield:
I have merged as the timoore-explicit-desc-sets branch:

https://github.com/vsg-dev/VulkanSceneGraph/tree/timoore-explicit-desc-sets

We can test it from there.

timoore:
I've updated the vsg-desc-set of vsgCs to use timoore-explicit-desc-sets. Actually, I haven't updated its FindContent declaration for VSG, but I doubt that matters for folks looking at this.

I've changed over to using the VSG class names for DescriptorConfigurator and GraphicsPipelineConfigurator. Also, I've moved the binding of the view-dependent descriptor set to the root node (WorldNode), which is something I've meant to do for a while.

robertosfield:
Thanks, just done a review of the changes to vsgCs.  Brain now in "thinking about shadows mode" so not attempt jumping in too deep on the DescriptorConfigurator/GraphicsPipelineConfigurator for now.  Will return to the topic on Monday.

robertosfield:
I have done another review of the changes and they still confuse me.  Adding extra complexity into classes to add extra level of debug capability is not a trade I don't think is well justified in this case, because if it leaves me confused then others will for sure.  Confusing code leads to bugs and hard to maintain code.

I'd like to distill the changes down to just what is functionally required.

I will do a code review of the usage in vsgCs to see if that can give me better insight to the motivation behind the changes, unfortunately as I can't compile vsgCs anymore that's got a bit more complicated as I can't test the runtime so will have to stick to code reviews.

timoore:
> I have done another review of the changes and they still confuse me. Adding extra complexity into classes to add extra level of debug capability is not a trade I don't think is well justified in this case, because if it leaves me confused then others will for sure. Confusing code leads to bugs and hard to maintain code.

It's true that I described the changes to DescriptorConfigurator as "debugging," but that is understating the case. One change insures correctness: it is simply an error to use descriptors in the same set if they have been defined in the ShaderSet as being in different descriptor sets. I also added functions to assign lists of DescriptorImages or DescriptorBuffers to an array of descriptors in one go; I hope that's not too controversial. 

The rest of the PR changes how pipeline layouts are created. Instead of being driven by a list of bindings that are accumulated on the fly, usually for a certain "material," a pipeline layout is created only from the ShaderSet and a list of defines. This supports two use cases which are not currently supported in VSG:

* creating a pipeline layout without a full list of bindings. This is the use in vsgCS for which I wrote the changes. Maybe I need to submit an example to make the utility clear. Among other things, it allows binding a descriptor set without having the full definition of the shader that will actually be used lower down in the tree.
* loading a model with vsgXchange that has custom shaders and additional descriptors / descriptor sets. With this change the pipelines will be set up correctly; a user will only have to visit the loaded graph and add descriptors.

Perhaps an example is needed?

> 
> I'd like to distill the changes down to just what is functionally required.
> 
> I will do a code review of the usage in vsgCs to see if that can give me better insight to the motivation behind the changes, unfortunately as I can't compile vsgCs anymore that's got a bit more complicated as I can't test the runtime so will have to stick to code reviews.

Sorry about that. I do my linux development on Fedora, and the oldest supported Fedora has CMake 3.24. I'm assuming that the version is still causing you problems.

robertosfield:
Thanks for the explanation.  The fact that it isn't obvious what is intended is a red flag for me.  

I've had 2 decades of reviewing open source submissions and trying to figure out the intent behind changes, mostly I can do this quickly, when it isn't obvious right away it's often an indication that the original code itself is part of the problem and the changes are a workaround for this so end up more awkward than they should be, of that the changes just take the code in a direction doesn't quite gel.

If I merge the changes as is I there is a chance I'll be confused when I next try to work on the code as it'll be a code base that is pushing in different direction rather than a coherent   

I understand the need for different usage case and that ShaderSet/*Configurator need to support.  I will review those and once I understand them have a good think about the problems and how to generalize ShaderSet/*Configurator to address these issues.  I am aware of other issues so probably time for me to spend a few days shaking it all down.

robertosfield:
As a heads up, I was just reviews the vsg::GraphicsPipelineConfigurator/DescriptorConfigurator in VSG master and it occurred to me that it definitely makes sense for GraphicsPipelineCofingurator to just have handle multiple DescriptorSets?Layouts internally, but also make make sense for DescriptorConfigurator to directly support multiple DescriptorSet/Layouts as well.

I also want o figure out how we introduce higher level constructs like vsg::ViewDependetState in a coherent/extensible way.  I'm going the direction of ShaderSet taking on more responsibility for specifying additions like ViewDependentState, and having scene graph setup code query what needs to be added rather than having hardwired code paths handle non generic stuff like happens right for handling ViewDependentState.

timoore:
> Thanks for the explanation. The fact that it isn't obvious what is intended is a red flag for me.
> 
> I've had 2 decades of reviewing open source submissions and trying to figure out the intent behind changes, mostly I can do this quickly, when it isn't obvious right away it's often an indication that the original code itself is part of the problem and the changes are a workaround for this so end up more awkward than they should be, of that the changes just take the code in a direction doesn't quite gel.

I take your point. I was trying, as I often do, to make the minimal change needed to get the required functionality and yet still be safe and in good style (thread safe, no funky global state, etc.), partly with the aim of being understandable. I think there's another issue in that I've described the usage several times in words -- even before starting vsgCs at all -- but the major example of usage in the VSG world is in vsgCs itself, which has to handle many other things too. I haven't been willing to write another example until getting some inital buy-off, so around we go...

> 
> If I merge the changes as is I there is a chance I'll be confused when I next try to work on the code as it'll be a code base that is pushing in different direction rather than a coherent

OK, I just don't see the changes as being that significant, but I admit that part of the motivation for making a PR is so it won't be my problem anymore :sunglasses:

timoore:
> As a heads up, I was just reviews the vsg::GraphicsPipelineConfigurator/DescriptorConfigurator in VSG master and it occurred to me that it definitely makes sense for GraphicsPipelineCofingurator to just have handle multiple DescriptorSets?Layouts internally, but also make make sense for DescriptorConfigurator to directly support multiple DescriptorSet/Layouts as well.

I think there should be a real division in responsibilities between `GraphicsPipelineConfigurator` and `DescriptorConfigurator`. The former should not deal with the assembly of descriptor sets at all, only their layouts. I know that the layout is, in a sense, driven by the descriptors that the model-building code can make available (e.g., which textures have been supplied), but I think the "descriptor configuration" should satisfy a given pipeline layout, not necessarily specifiy it.
> 
> I also want o figure out how we introduce higher level constructs like vsg::ViewDependetState in a coherent/extensible way. I'm going the direction of ShaderSet taking on more responsibility for specifying additions like ViewDependentState, and having scene graph setup code query what needs to be added rather than having hardwired code paths handle non generic stuff like happens right for handling ViewDependentState.

robertosfield:
I have created MultiDescriptorSetConfiguration branches of VSG and vsgXchange to experiment with adding multi DescriptorSet support into DescriptorConfiguration etc.  The first set of changes are:

https://github.com/vsg-dev/VulkanSceneGraph/compare/MultiDescriptorSetConfiguration

https://github.com/vsg-dev/vsgXchange/compare/MultiDescriptorSetConfiguration

robertosfield:
I have begun working on a customizable/extensible scheme for controlling how features like ViewDependentState are controlled via ShaderSet.   The commit that is the first step towards this functionality is: 016e265cd8a0491e44554fddb39776c6e071cc58

There is now a ShaderSet::customDescriptorSetBindings vector that is populated by the code setting up the ShaderSet and this is serialized along with the rest of the ShaderSet properties.  I have changed the vsgshaderset example in vsgExamples to assign the new vsg:ViewDependentStateBinding class that does wiring up at runtime,  It's a single line addition to the ShaderSet setup so I hope it'll be easy to work with.  

The changes to vsgExample are checked into the MultiDescriptorSetConfiguration branch:

https://github.com/vsg-dev/vsgExamples/compare/MultiDescriptorSetConfiguration

I haven't wired up vsg::Builder and vsgXchange::assimp and vsg::GraphicsPipelineConfigurator to use this yet, but I've run out of working day so I'll tackle this tomorrow morning.  This should enable me to remove the hard-wired support for ViewDependentState.

Once I have this all working I'll look at the assignment of default uniforms/textures and a then look at the creation of the pipeline layout.





https://github.com/vsg-dev/vsgExamples/compare/MultiDescriptorSetConfiguration

robertosfield:
A short update, I've made the changes to vsgXchange::assimp to use the new ShaderSet::customDescriptorSetBindings for injecting the ViewDependentState support and have it working.  It's still hacky so not ready to check in but at least suggest the approach is workable.   

I'm not happy with the naming but will let that lie for now as the required functionality is more important to figure out.  If there is sensible naming I've overlooked please make suggestions.

robertosfield:
@timoore, After lots of experimentation I have got a refactor of GraphicsPipelineConfigurator/DescriptorConfiguration to make it more extensible, this work isn't complete yet but we're now getting closer.  It may not yet be sufficient for the purposes of vsgCs but should be getting closer, so worth you have a look over the changes.

I have updated the VSG, vsgXchange, vsgPpints and vsgExamples, the later three should provide a bit of guide to how GraphicsPipelineConfigurator can now be used - it's more streamlined than before with GraphcisPipelineConfigurator::copyTo(stateGroup, sharedObjects) method doing more of the work:

https://github.com/vsg-dev/VulkanSceneGraph/tree/MultiDescriptorSetConfiguration
https://github.com/vsg-dev/vsgXchange/tree/MultiDescriptorSetConfiguration
https://github.com/vsg-dev/vsgExamples/tree/MultiDescriptorSetConfiguration
https://github.com/vsg-dev/vsgPoints/tree/MultiDescriptorSetConfiguration

GraphicsPipelineConfiguator now "has a" DescriptorConfigurator to ensure we're using one consistent set of APIs rather than ones that are competing with each other.

DescriptorConfigurator now supports multiple DescriptorSet, and the ViewDependentState support is now done via an extensible customDescriptorSetBinding class.  I'm not happy with the naming of this functionality but needed a placeholder to flesh out the design/implementation.

I haven't yet implemented the fallback Descriptor assignment for when there are non optional uniform entries that haven't been already setup before the GraphicsPipelineConfigurator::init() call, but I'll tackle that next.

I haven't implemented the additional assign methods that you added in this PR, I'll look at that as well.

Hopefully these changes will make sense to you and you'll now how to update vsgCs to work with them.  A MultiDescriptorSetConfiguration branch of vsgCs is probably the way to go.  Potentially I could tackle this but won't have time this week to look at it.

Once I've tackled the above items I write out details about this work on VSG discussion forum to other folks know the direction I'm taking this code.

timoore:
Looking at this while waiting for a trans-Atlantic flight to board... I don't see direct support for my use case in vsgCs. I don't want to bind a pipeline at the root of the scene graph. It's not clear what pipeline if any would be a good default pipeline, and binding pipelines is "expensive" anyway. But that may be OK if I can still use my `makePipelineLayout()` function to create a layout that will be compatible with the descriptor sets that I bind in a given point in the scene graph and above it.

Otherwise, the multiple descriptor set support looks like it's on the right track. I'll take a look at how I can use it in vsgCs.

robertosfield:
There isn't any connection between GraphicsPipelineConfigurator/DescriptorConfigurator and the root node of the scene graph.  There is just a convenience function for copying the state to a StateGroup, this StateGroup can be anywhere in the scene graph.

The GraphicsPipelineConfigurator::init() will setup the PipelineLayout and will handle all the default state.  We could move the create pipeline out into virtual function so it could be called separately.

robertosfield:
I have added BufferInfoList/ImageInfoList variants of DescriptorConfigurator::assignUniform/Texture to provide a bit more of the functionality provided by this PR: e085a9d60373c7cfab6168fedb95ed5b4169ca00

robertosfield:
As far as I'm away the key element missing from my recent work compared to this PR is the vsg::makePipelineLayout() function:

    extern VSG_DECLSPEC ref_ptr<PipelineLayout> makePipelineLayout(ref_ptr<ShaderSet> shaderSet,
                                                                   const std::set<std::string>& defines = {},
                                                                   int sets = -1);

I have reviewed the implementation but still remain confused by the maximum sets parameter.  Why is this needed?

If we are to implement such functionality then I think it would be best done as a virtual StateSet method and called create rather than make.  If we are creating objects then create is the far better naming.

robertosfield:
To help answer my question about the max set limit I have reviewed the code vsgCs that uses it. From src/vsgCs/GraphicsEnvironment.cpp:

    // We only care about the layout of the first three descriptor sets. All the model-specific
    // descriptors are in the fourth set, so we can get the layout for a "generic" shader and use it
    // for lighting and whole-tile parameters.
    overlayPipelineLayout
        = makePipelineLayout(shaderFactory->getShaderSet(VK_PRIMITIVE_TOPOLOGY_TRIANGLE_LIST),
                             shaderDefines,
                             pbr::TILE_DESCRIPTOR_SET);


Which suggests to me that the max set limit is tied to a particular ShaderSet usage approach, rather than a generally useful feature.  In fact I'd go further and suggest it's a bit of hack that could be easily avoided.

The easiest way to avoid this would be to put a define around all the model specific functionality, or use a combination of defines to decide this in the shader, when none of the defines associate with the model specific functionality are required the extra DescriptorSetLayout should never get enabled so automatically won't form part of the PipelineLayout.

The existing ShaderSet/GraphicsPipelineConfiguration/DescriptorConfigurator functionality should properly support this use of defines in this manner.  If it doesn't then this is a bug and we need to address it.

I don't believe the max set limit approach is helpful and should be abandoned as it makes the code confusing and only serves the purposes of a very specific usage approach of ShaderSet, it fails elsewhere and will lead to users trying to figure why it's in there and trying to work backwards to help to use it in their code, when they shouldn't be going there at all - they should just be using defines support as it was always meant to be used.

timoore:
> As far as I'm away the key element missing from my recent work compared to this PR is the vsg::makePipelineLayout() function:
> 
> ```
> extern VSG_DECLSPEC ref_ptr<PipelineLayout> makePipelineLayout(ref_ptr<ShaderSet> shaderSet,
>                                                                const std::set<std::string>& defines = {},
>                                                                int sets = -1);
> ```
> 
> I have reviewed the implementation but still remain confused by the maximum sets parameter. Why is this needed?
> 
See below.
> If we are to implement such functionality then I think it would be best done as a virtual StateSet method and called create rather than make. If we are creating objects then create is the far better naming.

I do like the convention that "create" means "return a ref_ptr." `createPipelineLayout` and `pipelineLayoutCreate` are both fine names. I prefer to keep functions that only use the public interface of a class as non-member functions, but this just a personal preference. I acknowledge that that approach may lose in the presence of polymorphism. 



> To help answer my question about the max set limit I have reviewed the code vsgCs that uses it. From src/vsgCs/GraphicsEnvironment.cpp:
> 
> ```
> // We only care about the layout of the first three descriptor sets. All the model-specific
> // descriptors are in the fourth set, so we can get the layout for a "generic" shader and use it
> // for lighting and whole-tile parameters.
> overlayPipelineLayout
>     = makePipelineLayout(shaderFactory->getShaderSet(VK_PRIMITIVE_TOPOLOGY_TRIANGLE_LIST),
>                          shaderDefines,
>                          pbr::TILE_DESCRIPTOR_SET);
> ```
> 
> Which suggests to me that the max set limit is tied to a particular ShaderSet usage approach, rather than a generally useful feature. In fact I'd go further and suggest it's a bit of hack that could be easily avoided.

While I would be quick to characterize most of my work as "a hack," this isn't. I didn't make this usage up myself; it's described in [Writing an Efficient Vulkan Renderer](https://zeux.io/2020/02/27/writing-an-efficient-vulkan-renderer/) under "Frequency-based descriptor sets." I think it would be good for VSG to adopt it throughout, but we seem far apart on that point :smiley_cat: 

The heart of the issue is that in Vulkan, "two pipeline layouts are defined to be 'compatible for set N' if they were created with identically defined descriptor set layouts for sets zero through N, and if they were created with identical push constant ranges." We can leave out the push constant usage as that is static in VSG. 

The Vulkan spec has a lot more to say about pipeline layout compatibility, but a summary is that the bindings of compatible descriptor sets remain valid across bindings and pipeline changes. As an example base case, if all pipelines use the same layout for descriptor set 0, and that descriptor set's data is constant for the whole scene graph, then it only needs to bound once, somewhere near the root of the scene graph. But, when binding that set, what layouts should be specified for descriptor sets beyond set 0? None make sense; they could (will) be all different. That's why I wrote the maxSet functionality. I don't know if there's any particular problem with including layouts for sets greater than the one that is being bound, but it seems very messy to me. 

> 
> The easiest way to avoid this would be to put a define around all the model specific functionality, or use a combination of defines to decide this in the shader, when none of the defines associate with the model specific functionality are required the extra DescriptorSetLayout should never get enabled so automatically won't form part of the PipelineLayout.
> 

Talk about hacks... I don't see why every shader should need this extra verbiage, especially if they could never work without the descriptors in the "fast-changing" descriptor set.

The real issue is that the pipeline layout for the global and slow-changing values are not just determined by a shader set, they are a part of a protocol between the application and the shaders. Perhaps the hack is that I used a ShaderSet in vsgCs to generate those layouts and didn't create them programatically in my own code. Or, it would be useful to be able to define a pipeline layout outside of the ShaderSet and ensure that a generated shader program conforms to it.

> The existing ShaderSet/GraphicsPipelineConfiguration/DescriptorConfigurator functionality should properly support this use of defines in this manner. If it doesn't then this is a bug and we need to address it.

Does the "defines" mechanism in the shader bindings support the kind of nested usage proposed?

> 
> I don't believe the max set limit approach is helpful and should be abandoned as it makes the code confusing and only serves the purposes of a very specific usage approach of ShaderSet, it fails elsewhere and will lead to users trying to figure why it's in there and trying to work backwards to help to use it in their code, when they shouldn't be going there at all - they should just be using defines support as it was always meant to be used.

Fails where "elsewhere?"

I think it is fortuitous that VSG supported this usage with minimal code on my part, and I will keep using it in vsgCs. A concrete benefit is that I can isolate the descriptor bindings for the global lighting, and those for a whole tile, from the bindings used in the individual models in a tile.

robertosfield:
The desire to reuse the descriptor set / layout should be perfectly viable
using defines - they are there to enable optional code paths in shaders,
they can do this for individual elements of DescriptorSet through to the
whole DescriptorSet.  It's an explicit mechanism that I've carried forward
from my early work on #pragma(tic) shader composition in the OSG.

The max set scheme is an implicit scheme that relies upon shaders,
descriptor sets to be all set up to work with this scheme.  It's not at all
obvious that the scheme is supported, let alone intended.  I read the code
many times and was clueless that this is what you intended.

The usage case you describe should be supported by the VSG's
MultiDescriptorSetConfiguration branch but may require some additional
defines if the extra functionality in high value DescriptorSet is left
implicitly on otherwise.

timoore:
> The desire to reuse the descriptor set / layout should be perfectly viable using defines - they are there to enable optional code paths in shaders, they can do this for individual elements of DescriptorSet through to the whole DescriptorSet. It's an explicit mechanism that I've carried forward from my early work on #pragma(tic) shader composition in the OSG. The max set scheme is an implicit scheme that relies upon shaders, descriptor sets to be all set up to work with this scheme. It's not at all obvious that the scheme is supported, let alone intended. I read the code 

Supported by whom? Vulkan obviously supports it. It wasn't supported by VSG's GraphicsPipelineConfigurator, or I wouldn't have had to go my own route. I assume that binding individual descriptor sets high in the scene graph does work, even in the presence of transparency. It could possibly be better documented in vsgCs, but this is fairly new code.

>many times and was clueless that this is what you intended. The usage case you describe should be supported by the VSG's MultiDescriptorSetConfiguration branch but may require some additional defines if the extra functionality in high value DescriptorSet is left implicitly on otherwise.

I still think it is important to be able to construct partial layouts, as done by makePipelineLayout. If you don't see its value, then it doesn't need to be in core VSG.

robertosfield:
On Sun, 16 Jul 2023 at 17:46, Tim Moore ***@***.***> wrote:

> Supported by whom?


The intent of your code wasn't at all clear, despite reviewing it many
times the intent of extra complexity was opaque. I'm an experienced
engineer used to trying to decode intent from code, if I struggle then it's
usually an indication the code is going able things in an awkward way.


> Vulkan obviously supports it. It wasn't supported by VSG's
> GraphicsPipelineConfigurator, or I wouldn't have had to go my own route.


The previous rev of GraphicsPipelineConfigurator was a first stepping
stone, trying to decouple definition of shaders and associated state from
building the scene graph is non trivial so taking it step by step is
required, with the plan to steadily figure out the range of usage cases
required and expand in a coherent way to encompass them.


> I assume that binding individual descriptor sets high in the scene graph
> does work,


That's been working with GraphicsPiplineConfigurator right from the start,
it was designed to facilitate this, the vsg::TileDatatbase/vsg::tile
functionality has been using this from before 1.0.


> even in the presence of transparency. It could possibly be better
> documented in vsgCs, but this is fairly new code.
>
> many times and was clueless that this is what you intended. The usage case
> you describe should be supported by the VSG's
> MultiDescriptorSetConfiguration branch but may require some additional
> defines if the extra functionality in high value DescriptorSet is left
> implicitly on otherwise.
>
> I still think it is important to be able to construct partial layouts, as
> done by makePipelineLayout. If you don't see its value, then it doesn't
> need to be in core VSG.
>
I can see value, just not in max set parameter approach which adds an
implicit convention to how ShaderSet should be treated.  The ShaderSet's
design is for it to explicitly reflect the functionality provided in the
shaders and the types that should be used with them.  If there are
components of shaders that won't always been required then they should all
be protected by the approach #defines.

It should also be possible to use multiple ShaderSet that share
vsg::DescriptorSetLayout via vsg::SharedObjects, as along as the settings
are the same the the same objects should be shared for them,

robertosfield:
@timoore One of the topics I've been thinking about is flexibility of which descriptors get assigned to which descriptor sets. In the vsgCs shaders you are hard-wiring the set numbers at the ShaderSet level, but using #define in the shaders,.so you are step towards abstracting away from descriptors being tied a a specific.  

The use of built in ShaderSet in the core VSG lacks any flexibility, it's hardwired in the declaration of the ShaderSet and hardwired in the shader.  

This morning it occurred to one approach might be to change the AttributeBinding::location and the UniformBinding::set & UniformBinding::binding from uint32_t to std::string, then have a map from string to uint32_t.  The std::string names could be used in the shaders and passed in automatically along with the #prama(tic) shader defines.

To afford backwards compatibility we'd need to use new versions of AttributeBinding and UniformBinding and keep the current uint32_t versions.  That would be a bit of cludge design/implementation wise but perhaps that's what we need to do in the short term/pre VSG-1.2.

Thoughts?

robertosfield:
Another element I'm wondering about is changing the standard VSG ShaderSet so that the PbrData and LightData are enclosed in #define's, for instance the standard_pbr.frag shader has:

~~~ glsl
layout(binding = 10) uniform PbrData
{
    vec4 baseColorFactor;
    vec4 emissiveFactor;
    vec4 diffuseFactor;
    vec4 specularFactor;
    float metallicFactor;
    float roughnessFactor;
    float alphaMask;
    float alphaMaskCutoff;
} pbr;

layout(set = 1, binding = 0) uniform LightData
{
    vec4 values[64];
} lightData;
~~~

Which force the shaders to required that the DescriptorSet's are assigned to set 0 and 1 respectively so even if they are not explicitly assigned the creation of the DescriptorSetLayout will need to encompass them.

Another possibility is to have the ViewDependentState DescriptorSet have default values for materials such as the PbrMaterial.  I'm also thinking it would be nice to standardize the material data structure as right now we have vsg::material, vsg::PbrMaterial and PhongMaterial, if we were to put a default in ViewDependentState then we'd need standardized first.

On the topic of makePipelineLayout/ShaderSet::createPipelineLayout() it just occurred to me that it might be useful to have a ShaderSet::createDescriptorSetLayout(uint32_t set, std::vector<string> defines).  It would also be useful to have means for getting defines required based on the named attributes & uniforms.

Lots of different possibilities for the future, for now I'd like to settle MultiDescriptorSetConfiguration branch and get it merged with VSG master.  I don't want to do this before vsgCs looks to be working OK with the branch.  If any of the above less intrusive suggestions would help in the short term let me know.

robertosfield:
@timoore  I have added three new vsg::ShaderSet methods that aren't a drop in replacement for makePipelineLayout() but will provide equivalent functionality along with more flexibility: 

b3973366c029a3db8ae8554ba929333c23b63897

I have added doxygen comment to the ShaderSet header to explain what each method does, the key difference is the use of a std::pair<uint32_t, uint32_t> range rather than a int max_set value.  This avoids need to juggle types and enables users to pick different subsets of the descriptor set pipeline.

~~~ cpp
       /// return the <mininum_set, maxiumum_set+1> range of set numbers encompassed UniformBindings
        std::pair<uint32_t, uint32_t> descriptorSetRange() const;

        /// create the descritor set layout.
        virtual ref_ptr<DescriptorSetLayout> createDescriptorSetLayout(const std::set<std::string>& defines, uint32_t set) const;

        /// create the pipeline layout for all descriptor sets enabled by specified defines or required by default.
        inline ref_ptr<PipelineLayout> createPipelineLayout(const std::set<std::string>& defines) { return createPipelineLayout(defines, descriptorSetRange()); }

        /// create pipeline layout for specified range {minimum_set, maxiumum_set+1> of descriptor sets that are enabled by specified defines or required by default.
        virtual ref_ptr<PipelineLayout> createPipelineLayout(const std::set<std::string>& defines, std::pair<uint32_t, uint32_t> range) const;
~~~

I now feel that the MultiDescriptorSetConfiguration branch should be ready to merge with VSG master.  I will wait for feedback on whether things work with vsgCs before going ahead an merging with VSG master.

timoore:
> @timoore I have added three new vsg::ShaderSet methods that aren't a drop in replacement for makePipelineLayout() but will provide equivalent functionality along with more flexibility:
> 
> [b397336](https://github.com/vsg-dev/VulkanSceneGraph/commit/b3973366c029a3db8ae8554ba929333c23b63897)
> 
These  look good, and I will have a go at using them in vsgCs.

>         /// create pipeline layout for specified range {minimum_set, maxiumum_set+1> of descriptor sets that are enabled by specified defines or required by default.
>         virtual ref_ptr<PipelineLayout> createPipelineLayout(const std::set<std::string>& defines, std::pair<uint32_t, uint32_t> range) const;
> ```

Just a note: if this is called with `t > 0`, then descriptor sets numbered less than `t` will be invalidated when the layout is used to bind a pipeline or descriptor set. This is fine for binding a specific shader set and its descriptors, but might have unintended consequences elsewhere.
> 
> I now feel that the MultiDescriptorSetConfiguration branch should be ready to merge with VSG master. I will wait for feedback on whether things work with vsgCs before going ahead an merging with VSG master.

robertosfield:
t > 0?  I didn't use t in any of the parameters or member variables, what are you referencing?

timoore:
> t > 0? I didn't use t in any of the parameters or member variables, what are you referencing?

I misread your range argument, whoops. I meant minimum_set.

timoore:
In trying the new interface with vsgCs, I'm running into a problem with an empty descriptor set. vsgCs skips descriptor set 1, reserving it for future tile set variables. This did work with the old code and my subclasses, but maybe it was a coincidence. I can see a couple of places where there's a problem; I'll see if I can fix them and make a pr.

robertosfield:
In the ShaderSet::createDescriptorSetLayout(..) originally thought that an empty set of bindings would be best mapped to a null ref_ptr<> but after merging it realized that when building into a set of DescriptorSetLayouts that you pass in when setting up the PipelineLayout the ordering in implictly based on the order in the vector of descriptor layouts so an empty set of bindings should be mapping to a default constructed DescriptorSetLayout.

I had planned on change this tomorrow, but will see if I can get it working here on my laptop before I head to bed.

robertosfield:
@timoore could you just pulling the latest VSG MultiDescriptorSetConfiguration branch and see if that works better. I have checked in commit: https://github.com/vsg-dev/VulkanSceneGraph/commit/416419ecd5c0b232523857bc31595a4f749d2a02

robertosfield:
This morning I did a review of changes last night and now realize the ShaderSet::createPipelineLayout(..) was buggy too.  I've checked in another fix: 1c6a3d5f2e2f5665573c78835faa1f5223ae12e7

timoore:
With the pull request I just made on the MultiDescriptorSetConfiguration branch, I can compile and run vsgCs. I'll push my  vsgCs changes to the vsg-desc-set branch; I'll refrain from merging them to master until @robertosfield makes a bump release.

robertosfield:
@timoore I am planning on merging the various MultiDescriptorSetConfiguration branches with master today.  Does this all look fine to you?

timoore:
It looks fine to me. As I said before, I'm going to wait on merging the vsgCs changes to its master until you bump the VSG release version.

robertosfield:
I'm tempted to do it today, but would need at least some testing on Windows to confirm that it's OK.

robertosfield:
I have merged all the respective MultiDescriptorSetConfiguration with VSG, vsgXchange, vsgPoints and vsgExamples master. I have also done some house cleaning removing no longer required branches.

I have another PR to review and merged this morning.  Once that's done I've merge the respective masters back into the 1.0  branches and create a 1.0.8-rc1 tag.

PR #876, opened 2023-07-18
[FIXED] VK_LOD_CLAMP_NONE computed mipmap levels

siystar:
# Pull Request Template

## Description

Fixed incorrect value returned by vsg::computeNumMipMapLevels when maxLod is set to VK_LOD_CLAMP_NONE.

## Type of change

## How Has This Been Tested?

Before:
```
computeNumMipMapLevels(width=256, maxLod=10) = 9
computeNumMipMapLevels(width=256, maxLod=VK_LOD_CLAMP_NONE) = 1000
```

After:
```
computeNumMipMapLevels(width=256, maxLod=10)
 = 9
computeNumMipMapLevels(width=256, maxLod=VK_LOD_CLAMP_NONE) = 9
```

## Checklist:

### DIFF
--- a/src/vsg/state/ImageInfo.cpp
+++ b/src/vsg/state/ImageInfo.cpp
@@ -93,29 +93,25 @@ FormatTraits vsg::getFormatTraits(VkFormat format, bool default_one)
 
 uint32_t vsg::computeNumMipMapLevels(const Data* data, const Sampler* sampler)
 {
-    uint32_t mipLevels = sampler != nullptr ? static_cast<uint32_t>(ceil(sampler->maxLod)) : 1;
-
-    // clamp the mipLevels so that its no larger than what the data dimensions support
-    uint32_t maxDimension = std::max({data->width(), data->height(), data->depth()});
-    if (mipLevels == VK_LOD_CLAMP_NONE)
+    uint32_t mipLevels = 1;
+    if (sampler)
     {
-        mipLevels = 1;
-
-        while ((1u << (mipLevels - 1)) < maxDimension)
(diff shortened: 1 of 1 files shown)


robertosfield:
I have just done quick review, as I wasn't familiar with VK_LOD_CLAMP_NONE so looked it up:

https://registry.khronos.org/vulkan/specs/1.3-extensions/man/html/VK_LOD_CLAMP_NONE.html

I noticed the use of float so looked up the specs on samplers:

https://registry.khronos.org/vulkan/specs/1.3-extensions/man/html/VkSamplerCreateInfo.html

The relevant part is:

~~~ cpp
    float                   minLod;
    float                   maxLod;
~~~

The vsg::computeNumMipMapLevels(...) function uses the maxLod value to help guide the number of mipLevels:

uint32_t vsg::computeNumMipMapLevels(const Data* data, const Sampler* sampler)
{
    uint32_t mipLevels = sampler != nullptr ? static_cast<uint32_t>(ceil(sampler->maxLod)) : 1;
    if (mipLevels == 0)
    {
        mipLevels = 1;
    }

It does so via a cast to uin32_t, and you changes compare this against the float VK_LOD_CLAMP_NONE which may work but it's not something I'd recommend.  Might I suggest comparing sampler->maxLod to VK_LOD_CLAMP_NONE would be more reliable.

Given this check would also need an additional sampler not null check it may be better to restructure the function so the logic is all clearer w.r.t how mipLevels is set.  I can have a bash at this, but I'm in the middle of other work so can't do this right away.

PR #877, opened 2023-07-18
[FIXED] array texture mipmaps

siystar:
# Pull Request Template

## Description

Fixed invalid baseArrayLayer and layerCount parameters passed to VkImageBlit when generating array texture mipmaps.

The Vulkan specification states: "The srcSubresource.baseArrayLayer + srcSubresource.layerCount of each element of pRegions must be less than or equal to the arrayLayers specified in VkImageCreateInfo when srcImage was created." (VUID-vkCmdBlitImage-srcSubresource-01707)

## Type of change

## How Has This Been Tested?

Tested that array texture mipmaps no longer contain artifacts

## Checklist:


robertosfield:
I have just done a quick review, I will need to investigate the issue further before merging so I understand the issue at hand.  Do you have a means for recreating the problem that I could use to test things before/after?

There is also similar code in TransferTask.cpp that will need updating as well.  Or, we refactor both so they use the same code to avoid divergence.

PR #878, opened 2023-07-19
Deal with "missing" descriptor sets in ShaderSet::assignDescriptor

timoore:
Also, add a comment to createPipelineLayout

I found some of the same problems as @robertosfield did last night. The fix to assignDescriptor is necessary for vsgCs' usage.


src/vsg/utils/GraphicsPipelineConfigurator.cpp
```diff
@@ -152,17 +152,21 @@ bool DescriptorConfigurator::assignUniform(const std::string& name, const Buffer
...
-    auto& ds = descriptorSets[set];
-    if (!ds)
+    if (auto currentSize = descriptorSets.size(); set >= currentSize)
     {
-        ds = vsg::DescriptorSet::create();
-        ds->setLayout = DescriptorSetLayout::create();
+        descriptorSets.resize(set + 1);
+        for (auto i = currentSize; i <= set; i++)
+        {
+            if (!descriptorSets[i])
```
robertosfield, line 160: The descriptorSets[i] will always be null in this case so this check can be removed,

robertosfield:
I've merged but I'll now simplify the code a bit.

robertosfield:
Small simplification: 41998332035ead0679541a1e49645fd81b2376c4

PR #906, opened 2023-08-14
Properly implemented vsg::getDirectoryContents on the Windows platform

### LINKED ISSUE #905: vsg::getDirectoryContents is broken on Windows
On the Windows platform this function simply does not do anything; to me it feels like someone just copied an API call and then only made sure it compiled ;)

Right now, the **FindFirstFileNameW/FindNextFileNameW** function pair is used, but with invalid parameters (required in/out parameters are simply set to nullptr). Note that I think these are actually the wrong functions, since these enumerate **hard links** only.

I think that actually the **FindFirstFileW/FindNextFileW** functions should be used.

I have a change here, I will see about creating a PR for this one

LeeRiva:
This change uses the FindFirstFileW/FindNextFileW API function pair (with correct parameters) instead of FindFirstFileNameW/FindNextFileNameW (which only enumerates hard links and was using invalid parameters) to close #905


robertosfield:
Could post the changes as separate PRs, bug fixes are critical and need to tracked separately, the typos should also be separate to personal preferences/style changes. Thanks.

PR #914, opened 2023-08-15
[FIXED] mipmap device support checks

siystar:
# Pull Request Template

## Description

Added check for the device support required for generating mipmaps before the VkImage is allocated, so that it can be allocated with the appropriate mip count, which is now stored per device in Image::_vulkanData.

Removed previously used workaround, a potentially harmful assignment to Sampler::maxLod from within ImageInfo::computeNumMipMapLevels().

Note, the Vulkan spec 34.3. lists required format support. Many formats lack guaranteed support for blit operations. For example, VK_FORMAT_R8G8_SNORM guarantees only SAMPLED_IMAGE and BLIT_SRC support, with support for BLIT_DST being up to the implementation.

## Type of change

## How Has This Been Tested?

Tested various images with Sampler::maxLod set

## Checklist:

- [X] have performed a self-review of my own code

### DIFF
--- a/include/vsg/core/Array2D.h
+++ b/include/vsg/core/Array2D.h
@@ -51,6 +51,7 @@ namespace vsg
         {
             if (_width != 0 && _height != 0)
             {
+                // note, doesn't allocate space for mipmaps
                 _data = _allocate(_width * _height);
                 auto dest_v = _data;
                 for (auto& v : rhs) *(dest_v++) = v;
@@ -60,6 +61,7 @@ namespace vsg
 
         Array2D(uint32_t width, uint32_t height, Properties in_properties = {}) :
             Data(in_properties, sizeof(value_type)),
+            // note, doesn't allocate space for mipmaps
             _data(_allocate(width * height)),
             _width(width),
             _height(height) { dirty(); }
@@ -72,6 +74,7 @@ namespace vsg
 
         Array2D(uint32_t width, uint32_t height, const value_type& value, Properties in_properties = {}) :
             Data(in_properties, sizeof(value_type)),
+            // note, doesn't allocate space for mipmaps
             _data(_allocate(width * height)),
             _width(width),
             _height(height)
--- a/include/vsg/core/Array3D.h
+++ b/include/vsg/core/Array3D.h
@@ -136,6 +136,7 @@ namespace vsg
 
             if (input.matchPropertyName("data"))
             {
+                // TODO: need to adapt dimensions for array imageViewType
                 size_t new_size = computeValueCountIncludingMipmaps(w, h, d, properties.maxNumMipmaps);
 
                 if (_data) // if data exists already may be able to reuse it
@@ -184,6 +185,7 @@ namespace vsg
             output.writeEndOfLine();
         }
 
+        // TODO: need to adapt dimensions for array imageViewType
         size_t size() const { return (properties.maxNumMipmaps <= 1) ? (static_cast<size_t>(_width) * _height * _depth) : computeValueCountIncludingMipmaps(_width, _height, _depth, properties.maxNumMipmaps); }
 
         bool available() const { return _data != nullptr; }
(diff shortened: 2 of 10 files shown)


robertosfield:
I have just done a first pass review, but will need to come back to it.  First reaction is that there are lots of different interdependent changes which makes me very wary about unexpected issues sneaking, this is type of set of changes that tend to fix X usage case but may break Y usage cases.

Given this I think it would be useful to take a step back and come up with tests that illustrate the problem usage case that the changes are intended to address, then we can discuss what might be the best solution - which may be this PR, but could also be resolvable another way.

robertosfield:
I have merged as a siystar-check_mip_support branch:

   https://github.com/vsg-dev/VulkanSceneGraph/tree/siystar-check_mip_support

I have also merged the corresponding PR to vsgExamples as branch and can confirm vsgheadless with this patch resolves the test cases.

I can't work all weekend so I'll stop at this point.  I will do more throughout review and think about the topic later in the coming working week.

robertosfield:
@siystar I am back on the case and reviewing these changes.  

One thing that popped up as an issue to resolve is that the VSG supports multiple vk/vsg::Device and the vsg::Image,  vsg::ImageInfo  and vsg::ImageView all have support for having separate Vulkan objects on each device.

Several of the number methods you have introduced set the local object state based on the Device that is passed in, this leaves open the possibility that different devices will support different capabilities and result in different values being assigned, the value that gets assigned will be the last device that these methods were invoked with, so could end up with the settings that simply don't work on other devices.

I think the correct way to resolve this type of divergence is to use one set of settings that work on all devices required, so the minimum features possible.  If applications want more capabilities on the most capable of devices then they really need to have a dedicated vsg::Image etc. for that device so there isn't any conflict.

The code in VSG master has the consistent treatment across devices by not bothering to check capabilities, which as you have found is a problem, so we'll need to come up with a solution that allows refinement of settings based on the available vk/vsg::Device capabilities.    I don't have any ideas yet, but I thought this was an issue worthy raising the red flag for so we can start brain storming. 

I think this red flag is likely a big enough issue hat this particular PR is useful for reference purposes rather than something ready to be merged with VSG master..  A general discussion on vsg-users might be appropriate on how to resolve the issues with VSG master.  Creating clarity on what usage cases work with VSG master, and what situations it fails would be helpful in this discussion.

robertosfield:
I have been looking at the PR and thinking about the topic and I'm inclined towards having some mechanism for checking the validity of the combination of vsg::Image and vsg::Sampler settings for a given Device.  If this validity checks fails for any of the devices then a vsg::Sampler that doesn't cause problem needs to be assigned.  

If a vsg::Sampler is shared between vsg::ImageInfo where some mipmaps can be generated correctly but others don't support it then we'd need to specialize the the ones that need mipmap generation but don't support it but still share between all those instances where it's not supported, and leave the cases where it is supported,

This refinement of scene graph setting to fit the capabilities of the hardware will be more general than just this case of whether generate mipmaps will succeed or not.  Presently we have a CollectResourceRequirements visitor and a CompileTraversal but neither of these can tweak the scene graph settings to avoid issues.  Do we need a ValidateSettings traversal of some kind?

robertosfield:
I have been reflecting on this topic and feel that validating a scene graph settings against the capabilities of the hardware/drivers is really what this PR is about, albeit in a specific feature/usage case.  

I think the right approach is for us to come up with a general scene graph validation approach rather than tackle different classes usage case by usage case.  Exactly what form general scene graph validation should take isn't something I have an answer for, perhaps adding another Visitor and virtual methods for validation might be the way forward.  

Exactly how to handle refinement of scene graph setting to work on given hardware will need to be part of this validation process, and while we can come up with some reasonable defaults it's something that will have an application specific component, so while the VSG might be able to provide some defaults, they'll need to be something that can be overridden by users.

I am going to close this PR and the associated branch that I checked in because I feel it's headed in a direction that is too usage case specific specific and hidden from users control.

robertosfield:
I have created a discussion thread for the topic of scene graph validation and settings refinement:

https://github.com/vsg-dev/VulkanSceneGraph/discussions/929

timoore:
Even though this is closed, I think the right way to go about this is to write the support for generating mipmaps even when blitting isn't supported, e.g. via a compute shader, render to texture, or even falling back to code on the CPU. I realize that this might be tedious to do for all the supported formats.

PR #917, opened 2023-08-17
Added support for the X1/X2 mouse buttons on the Windows platform.

LeeRiva:
## Description

This change adds (Windows-specific) code to determine whether the X1 or X2 mouse button was clicked. These new buttons are not actually used for anything yet.

## Type of change

## How Has This Been Tested?

I compiled the 'MyFirstVsgApplication' against my changed VSG library and verified using the debugger that the correct code paths are executed.

## Checklist:

### DIFF
--- a/include/vsg/platform/win32/Win32_Window.h
+++ b/include/vsg/platform/win32/Win32_Window.h
@@ -119,16 +119,40 @@ namespace vsgWin32
         return static_cast<vsg::ButtonMask>(mask);
     }
 
-    inline uint32_t getButtonDownEventDetail(UINT buttonMsg)
+    inline uint32_t getButtonDownEventDetail(UINT buttonMsg, WORD wParamHi)
     {
-        return buttonMsg == WM_LBUTTONDOWN ? 1 : (buttonMsg == WM_MBUTTONDOWN ? 2 : buttonMsg == WM_RBUTTONDOWN ? 3
-                                                                                                                : (buttonMsg == WM_XBUTTONDOWN ? 4 : 0)); // need to determine x1, x2
+        switch (buttonMsg)
+        {
+        case WM_LBUTTONDOWN: return 1;
+        case WM_MBUTTONDOWN: return 2;
+        case WM_RBUTTONDOWN: return 3;
+        case WM_XBUTTONDOWN:
+            if (wParamHi == XBUTTON1)
+                return 4;
+            else if (wParamHi == XBUTTON2)
+                return 5;
+            [[fallthrough]];
+        default:
+            return 0;
+        }
     }
 
-    inline uint32_t getButtonUpEventDetail(UINT buttonMsg)
+    inline uint32_t getButtonUpEventDetail(UINT buttonMsg, WORD wParamHi)
     {
-        return buttonMsg == WM_LBUTTONUP ? 1 : (buttonMsg == WM_MBUTTONUP ? 2 : buttonMsg == WM_RBUTTONUP ? 3
-                                                                                                          : (buttonMsg == WM_XBUTTONUP ? 4 : 0));
+        switch (buttonMsg)
+        {
+        case WM_LBUTTONUP: return 1;
+        case WM_MBUTTONUP: return 2;
+        case WM_RBUTTONUP: return 3;
+        case WM_XBUTTONUP:
+             if (wParamHi == XBUTTON1)
+                return 4;
+            else if (wParamHi == XBUTTON2)
+                return 5;
+            [[fallthrough]];
+        default:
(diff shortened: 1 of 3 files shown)


include/vsg/ui/PointerEvent.h
```diff
@@ -22,11 +22,11 @@ namespace vsg
...
         BUTTON_MASK_OFF = 0,
-        BUTTON_MASK_1 = 256,
-        BUTTON_MASK_2 = 512,
-        BUTTON_MASK_3 = 1024,
-        BUTTON_MASK_4 = 2048, /// mouse wheel up
-        BUTTON_MASK_5 = 4096  /// mouse wheel down
+        BUTTON_MASK_1   = 0x0100, /// left mouse button
+        BUTTON_MASK_2   = 0x0200, /// middle mouse button
+        BUTTON_MASK_3   = 0x0400, /// right mouse button 
+        BUTTON_MASK_4   = 0x0800, /// X1 mouse button
```
robertosfield, line 28: This conflicts with the original meaning, mouse wheel up/down which follows X11 convention of button 4 & 5 being reserved for scroll wheel up/down.  We'd need to be careful before changing this.

Changing 256 to 0x0100 etc.  I we are to change then 1 << x would probably be better.

include/vsg/platform/win32/Win32_Window.h
```diff
@@ -119,16 +119,40 @@ namespace vsgWin32
...
+        {
+        case WM_LBUTTONDOWN: return 1;
+        case WM_MBUTTONDOWN: return 2;
+        case WM_RBUTTONDOWN: return 3;
+        case WM_XBUTTONDOWN:
+            if (wParamHi == XBUTTON1)
+                return 4;
+            else if (wParamHi == XBUTTON2)
+                return 5;
+            [[fallthrough]];
```
robertosfield, line 134: return 0; is easier to read than [[fallthrough]] as it's more explicit about how this condition is meant to be treated, it's also less code.  Using a compiler hint to quieten a compile warning when just writing the code more explicitly to what you intend to happen is better.

robertosfield:
I have logitech vertical mouse and the X11/Xcb is generating button mask 1,2,3 for the main mouse button, 4 & 5 for the scroll when up/down and 8 & 9 for the extra buttons under the thumb.  I haven't ever uses these buttons for anything.

What is the story behind how the X1/X2 button events are used and map to devices?

LeeRiva:
Sorry I'm Windows-centric :shrug:

From [https://learn.microsoft.com/en-us/windows/win32/inputdev/about-mouse-input#xbuttons](MSDN)

> Windows supports a mouse with five buttons. In addition to the left, middle, and right buttons there are XBUTTON1 and XBUTTON2, which provide backward and forward navigation when using your browser.

So these have nothing to do with scrolling, I assumed it was the same for other platforms when there was talk about X1/X2 (also, mouse scrolling is handled using vsg::ScrollWheelEvent so I didn't even think about this). I'd assume your device under windows would generate X1/X2 button events for the under-the-thumb buttons

So maybe we should just forget about this change, right now there's no use case for this anyway?

robertosfield:
Tryout out my mouse buttons under my thumb I find firefox is going forward and back through the pages browsed, and the KDE kate editor is switch between the different tabs opened.

What happens with the X buttons under Windows applications?

LeeRiva:
> What happens with the X buttons under Windows applications?
It's up to the application, really. In browers it's forward/back (that seems to have been the initial intention, see my MSDN quote above), in Visual Studio it seems to toggle (randomly?) between tabs

Maybe it's just best to keep assuming a three-button mouse with mouse wheel, since that seems to be handled consistently across all platforms

robertosfield:
I have begun cherry-picking bits of this PR to come up with final version to merge:

https://github.com/vsg-dev/VulkanSceneGraph/tree/Win32_XBUTTON

I have started looking at how to get the XCB implementation to treat the XBUTTON 1&2 the same as Windows does, once that's done I'll merge with master.

@LeeRiva Could you test out the Win32_XBUTTON branch to make sure I haven't missed anything.

LeeRiva:
> @LeeRiva Could you test out the Win32_XBUTTON branch to make sure I haven't missed anything.

This looks good to me (I noticed two typos but I could not figure out how to attach this to this PR so it's now #920). I tested this using the vsgInput and vsgCameras examples (the latter I changed to select next/previous cameras using the X1/X2 buttons), I will commit this once everything's been merged back in.

robertosfield:
Thanks for the testing.  I have merged the typo fix and will merge the Win32_XBUTTON branch with master next so will close this PR.

PR #918, opened 2023-08-17
Fixed unchecked pointer access in GpuLayoutTechnique::setup which could lead to an access violation.

LeeRiva:
## Description

This commit fixes an unchecked pointer access in the GpuLayoutTechnique::setup function, possibly causing an access violation.

## Type of change

## How Has This Been Tested?

I ran the vsginput example (which crashes if the specified font file could not be found) from the vsgExamples repository.

## Checklist:

### DIFF
--- a/src/vsg/text/GpuLayoutTechnique.cpp
+++ b/src/vsg/text/GpuLayoutTechnique.cpp
@@ -44,10 +44,10 @@ void assignValue(T& dest, const T& src, bool& updated)
 
 void GpuLayoutTechnique::setup(Text* text, uint32_t minimumAllocation, ref_ptr<const Options> options)
 {
-    if (!text)
-        return;
     auto layout = text->layout;
     if (!layout) return;
+
+    if (!text->font) return;
 
     textExtents = layout->extents(text->text, *(text->font));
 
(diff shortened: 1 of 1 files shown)


robertosfield:
The GpuLoyoutTechnique shouldn't ever  be called with a null Text pointer so I think the issue must be further up in the chain.

LeeRiva:
Right, sorry about that. It's the text->font pointer that's nullptr, and it's being dereferenced in the 'setup' function (and this is where it crashes). Now whether that's something the caller should check or the function should account for is something I cannot tell. My original PR is invalid though

robertosfield:
I have copied in the checks in the CpuLayoutTechnique.cpp and this fixes the vsginput crash when no font is found:

0fe145eb7fdf94a5fc3e38ed64511665d92e2540

I will also add a check to vsginput to issue a warning when no found has been found.

robertosfield:
I have added the same font file check from vsgtext and vsgtextgroup into vsginput:

https://github.com/vsg-dev/vsgExamples/commit/4e74b2086daea5a01eb9194d9842f7555cea3c9a

PR #939, opened 2023-08-25
[REMOVED] assignment to Sampler::maxLod

siystar:
# Pull Request Template

## Description

Removed assignment to possibly shared Sampler during ImageInfo::computeNumMipMapLevels. Fixed handling of Sampler::maxLod values beyond the Image's mipLevels.

## Type of change

## How Has This Been Tested?

Tested vsgExamples check_mip_support:
```
info: test1 passed
```

Tested application

## Checklist:

### DIFF
--- a/src/vsg/app/TransferTask.cpp
+++ b/src/vsg/app/TransferTask.cpp
@@ -255,7 +255,7 @@ void TransferTask::_transferImageInfo(VkCommandBuffer vk_commandBuffer, Frame& f
     auto height = data->height();
     auto depth = data->depth();
     auto mipmapOffsets = data->computeMipmapOffsets();
-    uint32_t mipLevels = vsg::computeNumMipMapLevels(data, imageInfo.sampler);
+    uint32_t mipLevels = std::min(vsg::computeNumMipMapLevels(data, imageInfo.sampler), imageInfo.imageView->image->mipLevels);
 
     auto source_offset = offset;
 
--- a/src/vsg/commands/CopyAndReleaseImage.cpp
+++ b/src/vsg/commands/CopyAndReleaseImage.cpp
@@ -30,7 +30,7 @@ CopyAndReleaseImage::~CopyAndReleaseImage()
 CopyAndReleaseImage::CopyData::CopyData(ref_ptr<BufferInfo> src, ref_ptr<ImageInfo> dest, uint32_t numMipMapLevels) :
     source(src),
     destination(dest),
-    mipLevels(numMipMapLevels)
+    mipLevels(std::min(numMipMapLevels, dest->imageView->image->mipLevels))
 {
     if (source->data)
     {
--- a/src/vsg/state/ImageInfo.cpp
+++ b/src/vsg/state/ImageInfo.cpp
@@ -153,8 +153,8 @@ void ImageInfo::computeNumMipMapLevels()
             const auto& properties = data->properties;
             if (properties.blockWidth > 1 || properties.blockHeight > 1 || properties.blockDepth > 1)
             {
-                sampler->maxLod = 0.0f;
                 mipLevels = 1;
+                generateMipmaps = false;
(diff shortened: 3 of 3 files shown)


psi29a:
Probably should go into RC3

robertosfield:
I have merged this PR as a siystar-sampler branch:

https://github.com/vsg-dev/VulkanSceneGraph/tree/siystar-sampler

When testing the vsgvalidate test program in vsgExamples vsgvalidate branch I now see the output:

~~~ sh
info: test1 passed
Warning: test2 failed: image will be created with 6 mipLevels, but we have no way to fill them in.
~~~

robertosfield:
My brain is still a bit in holiday mode, but on reviewing the changes and the general way that ImageInfo::computeNumMipMapLevels()  used to overwrite the Sampler::maxLod and now just overwrites the Image::mipLevels I am left wondering it it shouldn't do the later either.

At least from the perspective of ImageInfo "has a" Sampler and "has a" ImageView which "has a" Image, and there is plenty of scope for a single Image to be shared between multiple ImageView and ImageInfo and each one to have it's own Sampler.  

All this is so loosely coupled, having ImageInfo handle the lower levels down may have been historical convenient for me to have implemented it to get things up and running but now perhaps it's a red flag that this bit of the design and implementation needs to be revisited.

My head isn't yet in right place to figure our what we should do short term, w,r,t this PR, or longer term.  Establishing some distinct usage cases that we need to handle is probably a good way to provide some framework for what should probably be done.  I'll have a look at the other PR's right now and then return to this topic.

Happy to take input on the usage cases to use to help guide this review.

robertosfield:
@siystar I have returned to this topic and while I haven't come up with any clear plan I'm leaning towards vsg::Image::mipLevels taking a more central role in determine whether generation of map is required.

Right now we have Image::mipLevels being set implicitly from either the vsg::Data::computeMipmapOffsets() if vsg::Image::Image(ref_ptr<Data>) constructor is used on later by ImageView::computeNumMipMapLevels(). 

The problem with the ImageView::computeNumMipMapLevels() setting Image::mipLevels is multiple ImageView can reference the same vk/vsg::Image and each one can have their own associated vsg::Sampler which can have it's own maxLod value which would then invalidate the vsg::Image::mipLevels.  

This issue is essentially the same issue as this PR attempts to fix, but in fixing it from vsg::Sampler it breaks it from vsg::Image.  For this reason I now feel that this PR will have to be rejected.

What I am currently thinking is that vsg::Image::mipLevels need to be the maximum mipLevel required by the vsg::Sampler that will be used with that vsg::Image and the maximum number supported vsg::Image::extents/vsg::Data.

I think it may also be worth having a vsg::Image::generateMipmaps(CommandBuffers&) or similar method for recording the Vulkan commands required to copy the various image levels down.  It may make sense to have a vsg::Command for doing this as well. For instance if one renders to a level 0 and then wants to create all the extra mipmap levels.

The question of how to set the Image::mipLevels automatically if it hasn't been explicitly set yet is an open question - ideally we'd want scene graph builders to explicitly set everything up, but this is super long winded so the VSG so far has jumped in with what are probably sensible defaults for most usage cases.  I think some scheme that allows explicit setup of the Image::mipLevels but when it's not find the maximum Sampler::maxLod of the all the vsg::Sampler that will be associated with a vsg::Image via vsg::ImageView, and combine this with the Image::extents.

robertosfield:
I'm going to close this PR as I don't think it adequately resolves the problem, it's more a workaround for a specific usage case that is currently broken rather than an fix for the underlying problem - the coupling of vsg::ImageView/Sampler/Image in ways that break in certain usage cases.

robertosfield:
I have created a Issue #945 to provide a new starting place for figuring out how to handle the issue of sharing vsg::Image and vsg::Sampler between vsg::ImageInfo.

PR #943, opened 2023-08-27
Allocate DescriptorBuffer memory from pools

### LINKED ISSUE #942: DeviceMemory::DeviceMemory() throws error with VK_ERROR_TOO_MANY_OBJECTS
vsgCs throws this error when using Google Photorealistic Terrain and zooming in on an area.

Hardware: AMD Integrated Graphics (Renoir)
Driver version: 2.0.193
API version: 1.2.182

This driver has a relatively modest value for maxMemoryAllocationCount: 4096. The Google 3D Tiles are complex objects.

I don't totally understand the DeviceMemory code, but it looks to me like a call to vkAllocateMemory is made for each VSG buffer object compiled. Is that right? I see a MemorySlots object in DeviceMemory that looks like it is intended to satisfy small allocations from one DeviceMemory object, but I don't see how it is used.

timoore:
Allocate buffers that usually contain uniform data from memory pools, instead of directly calling vkAllocateMemory. Some drivers have very low limits on the number of Vulkan memory objects that can be allocated.

This fixes #942.

### DIFF
--- a/src/vsg/state/DescriptorBuffer.cpp
+++ b/src/vsg/state/DescriptorBuffer.cpp
@@ -11,6 +11,7 @@ THE SOFTWARE IS PROVIDED "AS IS", WITHOUT WARRANTY OF ANY KIND, EXPRESS OR IMPLI
 </editor-fold> */
 
 #include <vsg/core/compare.h>
+#include <vsg/core/Exception.h>
 #include <vsg/io/Logger.h>
 #include <vsg/io/Options.h>
 #include <vsg/state/DescriptorBuffer.h>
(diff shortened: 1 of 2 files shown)


src/vsg/state/DescriptorBuffer.cpp
```diff
@@ -177,15 +178,16 @@ void DescriptorBuffer::compile(Context& context)
...
+                                                                         VK_MEMORY_PROPERTY_HOST_VISIBLE_BIT | VK_MEMORY_PROPERTY_HOST_COHERENT_BIT);
+                    if (deviceMemory)
                     {
-                        bufferInfo->buffer->bind(memory, offset);
+                        bufferInfo->buffer->bind(deviceMemory, offset);
                     }
                     else
                     {
-                        warn("DescriptorBuffer::compile(..) unable to allocate buffer within associated DeviceMemory.");
+                        throw Exception{"Error: DescriptorBuffer::compile(..) unable to allocate buffer."};
```
robertosfield, line 190: If we are to throw an exception then the returning a relevant Vulkan error would be helpful.

Alternatively we could use vsg::fatal().

timoore: > If we are to throw an exception then the returning a relevant Vulkan error would be helpful.
> 
I agree. That would require changing the memory pool interface,  but perhaps that is desirable.
> Alternatively we could use vsg::fatal().

robertosfield: This close to a release I want to avoid getting diverted into lots of other work - occasionally I need to get some paid work done :-)

robertosfield:
For testing purposes and help refine the code I have created a timoore-buffer-memory-pools branch:

https://github.com/vsg-dev/VulkanSceneGraph/tree/timoore-buffer-memory-pools

I have also changed the way the flags are passed into the reserveMemory(..) method to help make the code more readable and easier to adapt the settings: 00138ad3b3f03793ff3ebc5a06ba49f304fe8c14

We may way to change to using device local storage for the DescriptorBuffer and than host visible.  I am wary of making such a change off the the cuff though so I'l leave as is and we can optimize later.

robertosfield:
Testing looks fine, so I have now merged the timoore-buffer-memory-pools branch with VSG master.,

robertosfield:
I hadn't come across the article, thanks for the link.  The explanation of how AMD APUs are managing device local is particularly relevant as I am using one right now! I will need to reflect on the topic.  

Further done the line it might be worth having some scheme to set preferences and fallbacks to cope with how different drivers/hardware are best used w.r.t memory allocations.

Issue #954, opened 2023-09-01
Storage buffer support for Configurators and ShaderSet


theodoregoetz:
We would like to create a (large) fragment shader *storage buffer* along side the `GraphicsPipelineConfigurator` and `ShaderSet` classes in the same way we handle uniform buffers. From the interface it looks like only attributes, uniforms and textures are suppported.

We see two possible solutions:

## 1. Allow the `VK_DESCRIPTOR_TYPE_STORAGE_BUFFER` descriptorType to propagate through the Configurators in `GraphicsPipelineConfigurator.cpp`

`ShaderSet` accepts a storage buffer descriptorType in `ShaderSet::addUniformBinding()`. To use this in the Configurators, we found that it need only be passed to `DescriptorBuffer::create()` in the two `DescriptorConfigurator::assignUniform()` methods found in `GraphicsPipelineConfigurator.cpp`.

```diff
 bool DescriptorConfigurator::assignUniform(
     const std::string& name, 
     ref_ptr<Data> data, 
     uint32_t dstArrayElement)
 {
     if (auto& uniformBinding = shaderSet->getUniformBinding(name))
     {
         assigned.insert(name);
 
         // set up bindings
         if (!uniformBinding.define.empty()) 
             defines.insert(uniformBinding.define);
 
         // create uniform and associated DescriptorSets and binding
         return assignDescriptor(
             uniformBinding.set, 
             uniformBinding.binding, 
             uniformBinding.descriptorType, 
             uniformBinding.descriptorCount, 
             uniformBinding.stageFlags,
             DescriptorBuffer::create(
                 data ? data : uniformBinding.data, 
                 uniformBinding.binding,
                 dstArrayElement,
+                uniformBinding.descriptorType
             )
         );
     }
     return false;
 }
```

While this does work, it's not ideal since we are using methods like `assignUniform()` for buffers that are not actually "uniform" type.

## 2. Add `ShaderSet::addStorageBinding()` and `GraphicsPipelineConfigurator::assignStorage()` methods

This involves adding a `StorageBinding` struct to the `ShaderSet` class along with several new methods and possibly asserting on the descriptorTypes passed in to ensure that a compatible value is used (i.e. disallow using `VK_DESCRIPTOR_TYPE_STORAGE_BUFFER` with `ShaderSet::addUniformBinding()`.

## Note

[...]

PR #956, opened 2023-09-04
renamed "Uniform" to "DescriptorBuffer"

theodoregoetz:
Using the name "DescriptorBuffer" instead of "Uniform" with ShaderSet and GraphicsPipelineConfigurator better conveys that these methods work with both uniform and storage (and sometimes even image/texture) bindings.

Part of #954 

This is a rename only and no functionality was changed. The public member variable `ShaderSet::uniformBindings` was renamed to `ShaderSet::bufferBindings` and a reference was created to keep the the API from breaking. Also, updated names in the read/write methods are set to be used when the version hits `1.0.10`.

- [x] This change requires a documentation update

## Testing

See the associated storage buffer example: https://github.com/vsg-dev/vsgExamples/tree/master/examples/utils/vsgstoragebuffer

## Checklist:

### DIFF
--- a/include/vsg/utils/GraphicsPipelineConfigurator.h
+++ b/include/vsg/utils/GraphicsPipelineConfigurator.h
@@ -45,18 +45,25 @@ namespace vsg
         bool two_sided = false;
 
         bool enableTexture(const std::string& name);
-        bool enableUniform(const std::string& name);
+        bool enableBuffer(const std::string& name);
 
         bool assignTexture(const std::string& name, ref_ptr<Data> textureData = {}, ref_ptr<Sampler> sampler = {}, uint32_t dstArrayElement = 0);
         bool assignTexture(const std::string& name, const ImageInfoList& imageInfoList, uint32_t dstArrayElement = 0);
 
-        bool assignUniform(const std::string& name, ref_ptr<Data> data = {}, uint32_t dstArrayElement = 0);
-        bool assignUniform(const std::string& name, const BufferInfoList& bufferInfoList, uint32_t dstArrayElement = 0);
+        bool assignBuffer(const std::string& name, ref_ptr<Data> data = {}, uint32_t dstArrayElement = 0);
+        bool assignBuffer(const std::string& name, const BufferInfoList& bufferInfoList, uint32_t dstArrayElement = 0);
 
         bool assignDescriptor(uint32_t set, uint32_t binding, VkDescriptorType descriptorType, uint32_t descriptorCount, VkShaderStageFlags stageFlags, ref_ptr<Descriptor> descriptor);
 
-        /// call after all the textures/uniforms have been explictly assigned to add in textures/uniforms descriptors that are enabled by default (define == "").
+        /// call after all the textures and buffers have been explictly assigned to add in texture/buffer descriptors that are enabled by default (define == "").
         bool assignDefaults();
+
+        /// deprecated. use enableBuffer() instead
+        bool enableUniform(const std::string& name);
+        /// deprecated. use assignBuffer() instead
+        bool assignUniform(const std::string& name, ref_ptr<Data> data = {}, uint32_t dstArrayElement = 0);
+        /// deprecated. use assignBuffer() instead
+        bool assignUniform(const std::string& name, const BufferInfoList& bufferInfoList, uint32_t dstArrayElement = 0);
 
         std::set<std::string> assigned;
         std::set<std::string> defines;
@@ -105,11 +112,11 @@ namespace vsg
 
         bool enableArray(const std::string& name, VkVertexInputRate vertexInputRate, uint32_t stride, VkFormat format = VK_FORMAT_UNDEFINED);
         bool enableTexture(const std::string& name);
-        bool enableUniform(const std::string& name);
+        bool enableBuffer(const std::string& name);
 
         bool assignArray(DataList& arrays, const std::string& name, VkVertexInputRate vertexInputRate, ref_ptr<Data> array);
         bool assignTexture(const std::string& name, ref_ptr<Data> textureData = {}, ref_ptr<Sampler> sampler = {});
-        bool assignUniform(const std::string& name, ref_ptr<Data> data = {});
+        bool assignBuffer(const std::string& name, ref_ptr<Data> data = {});
 
         // setup by assign calls
         ref_ptr<ShaderCompileSettings> shaderHints;
@@ -127,6 +134,11 @@ namespace vsg
         ref_ptr<PipelineLayout> layout;
         ref_ptr<GraphicsPipeline> graphicsPipeline;
         ref_ptr<BindGraphicsPipeline> bindGraphicsPipeline;
+
+        /// deprecated. use enableUniform() instead
+        bool enableUniform(const std::string& name);
+        /// deprecated. use assignBuffer() instead
+        bool assignUniform(const std::string& name, ref_ptr<Data> data = {});
 
     protected:
         void _assignShaderSetSettings();
--- a/include/vsg/utils/GraphicsPipelineConfigurator.h
+++ b/include/vsg/utils/GraphicsPipelineConfigurator.h
@@ -58,11 +58,13 @@ namespace vsg
         /// call after all the textures and buffers have been explictly assigned to add in texture/buffer descriptors that are enabled by default (define == "").
         bool assignDefaults();
 
-        /// deprecated. use enableBuffer() instead
+        [[deprecated("use enableBuffer()")]]
         bool enableUniform(const std::string& name);
-        /// deprecated. use assignBuffer() instead
+
+        [[deprecated("use assignBuffer()")]]
         bool assignUniform(const std::string& name, ref_ptr<Data> data = {}, uint32_t dstArrayElement = 0);
-        /// deprecated. use assignBuffer() instead
+
+        [[deprecated("use assignBuffer()")]]
         bool assignUniform(const std::string& name, const BufferInfoList& bufferInfoList, uint32_t dstArrayElement = 0);
 
         std::set<std::string> assigned;
@@ -135,9 +137,10 @@ namespace vsg
         ref_ptr<GraphicsPipeline> graphicsPipeline;
         ref_ptr<BindGraphicsPipeline> bindGraphicsPipeline;
 
-        /// deprecated. use enableUniform() instead
+        [[deprecated("use enableBuffer()")]]
         bool enableUniform(const std::string& name);
-        /// deprecated. use assignBuffer() instead
+
+        [[deprecated("use assignBuffer()")]]
         bool assignUniform(const std::string& name, ref_ptr<Data> data = {});
 
     protected:
--- a/include/vsg/utils/GraphicsPipelineConfigurator.h
+++ b/include/vsg/utils/GraphicsPipelineConfigurator.h
@@ -45,26 +45,29 @@ namespace vsg
         bool two_sided = false;
 
         bool enableTexture(const std::string& name);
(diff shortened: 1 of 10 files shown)


theodoregoetz:
Looking at the documentation (vsg-dev/vsgTutorial), I don't think anything there needs to change.

robertosfield:
Thanks for the changes.  I have done a first pass review and it all looks consistent with what I had in mind with a renaming.

One thing that has just struct me is the changes introduce a new naming conflict.  In Vulkan and the VSG we have the vk/vsg::Buffer which is used for a far wider usage than just DescriptorBuffer, so we are solving one naming issue but introducing a new one that could cause further confusion.

I will need to reflect on this overlap and doing further passes over the changes, happy to take suggestions.

robertosfield:
After a second read through I'm wondering if DescriptorBinding would be better than UniformBinding/BufferBinding.  This would also fit with the DescriptorConfigurator naming, and the DescriptorBuffer/DescrptorImage naming.

Thoughts?

theodoregoetz:
Agreed. I'll rework this for "Uniform" -> "Descriptor".

robertosfield:
Thanks.. I am sorry that using Descriptor as the base name didn't occur to me in discussions, alas sometimes only when you see something implemented does how coherent it is come to fore.

theodoregoetz:
After a closer look, I figured that the Buffer term was good to have in GraphicsPipelineConfigurator because it differentiates the descriptor from the images (textures) and vertex attributes (arrays). But for sure the ShaderSet's methods were better using the term "descriptor" because it holds onto textures, uniforms and storage buffers. I also fixed up the read/write to be backwards compatible and triggered on version `1.0.10` which doesn't exist yet - I'm not really sure how to want to handle that.

Additionally, I started to use the `[[deprecated]]` C++ attribute and found that the errors I get from gcc are quite nice. Let me know if this causes any problems.

include/vsg/utils/GraphicsPipelineConfigurator.h
```diff
@@ -58,11 +58,13 @@ namespace vsg
         /// call after all the textures and buffers have been explictly assigned to add in texture/buffer descriptors that are enabled by default (define == "").
         bool assignDefaults();
 
-        /// deprecated. use enableBuffer() instead
+        [[deprecated("use enableBuffer()")]]
```
robertosfield, line 61: Shouldn't this be enableDescriptor()?

W.r.t [[deprecated]], do you know whether doxygen can handle it in a sensible way?

theodoregoetz: no. `GraphicsPipelineConfigurator::enableBuffer()` is still good in my opinion. see my previous comment posted at the same time as yours just now.

src/vsg/utils/GraphicsPipelineConfigurator.cpp
```diff
@@ -151,7 +151,7 @@ bool DescriptorConfigurator::assignTexture(const std::string& name, const ImageI
 
 bool DescriptorConfigurator::enableBuffer(const std::string& name)
```
robertosfield, line 152: Shouldn't this be enableDescriptor(..)?

src/vsg/utils/GraphicsPipelineConfigurator.cpp
```diff
@@ -200,7 +200,7 @@ bool DescriptorConfigurator::assignUniform(const std::string& name, ref_ptr<Data
 
 bool DescriptorConfigurator::assignBuffer(const std::string& name, const BufferInfoList& bufferInfoList, uint32_t dstArrayElement)
```
robertosfield, line 201: Shouldn't this be assignDescriptor(..)?

src/vsg/utils/GraphicsPipelineConfigurator.cpp
```diff
@@ -471,8 +471,8 @@ bool GraphicsPipelineConfigurator::assignBuffer(const std::string& name, ref_ptr
 /// deprecated. use GraphicsPipelineConfigurator::assignBuffer() instead
 bool GraphicsPipelineConfigurator::assignUniform(const std::string& name, ref_ptr<Data> data)
 {
-    warn("GraphicsPipelineConfigurator::enableUniform() has been deprecated."
-         " use GraphicsPipelineConfigurator::enableBuffer() instead.");
+    warn("GraphicsPipelineConfigurator::assignUniform() has been deprecated."
```
robertosfield, line 474: Is the warning redundant if  we are using the [[deprecated]] compiler hint?

theodoregoetz: Possibly. I don't know how this attribute and how VSG's logging plays with release and debug. I figured you might have some opinion on which is best. I can take this out if you think it best.

src/vsg/utils/GraphicsPipelineConfigurator.cpp
theodoregoetz: no. same reason as for in `GraphicsPipelineConfigurator` since we want to differentiate between buffers (uniform or storage) and images (textures).

include/vsg/utils/GraphicsPipelineConfigurator.h
robertosfield: I think enableBuffer() is more confusing that the original enableUniform().  Buffer has a distinct mean in VSG and Vulkan, it refers to a whole buffer not a subset that is can be as a uniform and storage.

enableDescriptor() is more generic than enableTexture()/enableUniform() but it does at least still map to DescriptorBuffer/DescriptorImage.  If we want to replace enableUniform()/enableTexture() then it would probably be best as enableDescriptorBuffer()/enableDescriptorImage().

theodoregoetz: Right, I came to nearly the same conclusion but decided at the time that shorter method names would be better. Note that the word "descriptor" shows up twice in `DescriptorConfigurator::enableDescriptorBuffer()` so why not just make it `DescriptorConfigurator::enableBuffer()`?

But really I could have gone either way. `DescriptorConfigurator::enableDescriptorBuffer()` is not too bad in length.

theodoregoetz: Also, I want to keep this PR to the uniform/storage issue and follow-up with a rename for the texture/image methods right after.

robertosfield: Perhaps just DescriptorConfigurator::enable(..) then?

Though, as we'll need to keep it consistent with the GraphicsPipelineConfigurator API.

src/vsg/utils/GraphicsPipelineConfigurator.cpp
robertosfield: Once this PR is settled a bit further I'll merge this PR as a branch and then do testing with it.

theodoregoetz: for deprecations, I believe it best to be obnoxious about it. Otherwise it's easier to ignore.

theodoregoetz:
The last commit renames the configurator methods to `...DescriptorBuffer()`. I also added `/// deprecated: ` comments to account for doxygen which seems to ignore the `[[deprecated]]` attribute in my (lmited) tests.

include/vsg/utils/GraphicsPipelineConfigurator.h
theodoregoetz: If you can stand the slight inconsistency for this PR, I'd like to make this change first, keeping it as is at the moment (`enableDescriptorBuffer()`, etc) and come back with another PR to rename `enableTexture()` to `enableDescriptorImage()`. This is because the down-stream changes are many and separating this out keeps it all straight in my head.

robertosfield:
FYI, I've checked in some work that includes a bump of the VSG master version to 1.0.10-rc1 so the IO checks in this PR should now work with VSG master.

theodoregoetz:
I believe this PR is complete now, unless @robertosfield you really want to have the rename from `Texture` to `DescriptorImage` done at the exact same time (keep in mind that this would more than double the down-stream updates)

robertosfield:
I have not another review and want to make a few changes so will merge this PR as a theodoregoetz-buffer-binding branch.  I need to decide upon the naming and want to simplify the deprecated entries - I think one is enough, we certainly don't need [[deprecated]], doxygen comment and console message all saying the same thing.

The branch is:
https://github.com/vsg-dev/VulkanSceneGraph/tree/theodoregoetz-buffer-binding

robertosfield:
I have decided to create a parallel DescriptorBinding branch so I can incrementally cherry-pick changes from the  theodoregoetz-buffer-binding branch:

https://github.com/vsg-dev/VulkanSceneGraph/tree/DescriptorBinding

So far I have just copied across the changes to alignment and DescriptorBuffer doxygen comment as these don't affect the API and fix a bug.  I'll merge these changes with VSG master right away as they uncontroversial.  I'll then copy across the changes that change the API.

robertosfield:
I now think I have got to the new DescriptorBinding branch parity with this PR and the theodoregoetz-buffer-binding branch, these are the changes so far:

https://github.com/vsg-dev/VulkanSceneGraph/compare/DescriptorBinding

I have updated vsgXchange, vsgPoints and vsgExamples to not use the deprecated methods:

https://github.com/vsg-dev/vsgXchange/compare/DescriporBinding
https://github.com/vsg-dev/vsgPoints/compare/DescriptorBindings
https://github.com/vsg-dev/vsgPoints/compare/DescriptorBindings

The main differences to this PR is that I have only used the [[deprecated]] compile hint and have not assign any doxygen comments or vsg::warn(..) messages, and the enable/assignUniform(..) methods have been replaced by enable/assignDescriptor(..) methods rather than enable/assignDescriptorBuffer(..).  

I decided just using enable/assignDescriptor rather than enable/assignDescriptorBuffer as I feel that it fits better with the enable/assignArray(..) naming, and offers the path of just overloading the assignDescriptor(..) methods to have an optional ref_ptr<Sampler> which can then replace the enable/assignTexture(..) method with a standard enableDescriptor/assignDescriptor() naming.

I think the DescriptorBinding branch is ready to merge with master, but will reflect on the possible change of enable/assignTexture(..) before I do anything further. Happy to take suggestions.

theodoregoetz:
Sounds good to me except I don't quite understand how you will handle textures with just `assignDescriptor` - uniform/storage takes a `BufferInfoList` and the texture takes an `ImageInfoList`.

robertosfield:
I have had a look at unifying the enableTexture/assignTexture to use the enableDescriptor/assignDescriptor() usage but to make it work well we'll need to check the vkDescriptorType and use that to choose with a DescriptorImage, DescriptorBuffer or [DescriptorTexelBufferView](https://github.com/vsg-dev/VulkanSceneGraph/blob/master/include/vsg/state/DescriptorTexelBufferView.h) should be used.

The vsg::DescriptorTexelBufferView isn't yet supported by ShaderSet/DescriptorConfigurator/GraphicsPipelineConfigurator so this would be best tackled as separate lock of work.  I am lots of other work to get on with so I won't go pursue another avenue of unplanned work.  Resolving the enableTexture/assignTexture() is a coherent way will also take more time so I'll now merged the respective DescriptorBinding branches with VSG etc. master and leave these matters to future work.

robertosfield:
> Sounds good to me except I don't quite understand how you will handle textures with just `assignDescriptor` - uniform/storage takes a `BufferInfoList` and the texture takes an `ImageInfoList`.

Oh, that's simple - you just have overload assignDescriptor(..) methods.  That's the easy part, the harder part is implementing a generate enableDescirptor() that handles all the different Descriptor types.

PR #967, opened 2023-09-16
[REMOVED] unnecessary const_cast

siystar:


robertosfield:
Did you even try to compile the code before creating a PR?  

It breaks the build, the cast is very much needed as the viewID is declared const.

siystar:
My apologies for the noise. I thought I'd compiled the PR, but evidently I haven't. I must have lost track because of the number of active branches/PRs in my pipeline, some of them were tested on their own and some were tested in combination with others. In general, the only PRs I don't test are the typo fixes.

PR #988, opened 2023-10-04
Added cascaded shadow maps suppot for directional lights.

robertosfield:


include/vsg/state/ViewDependentState.h
```diff
@@ -96,38 +104,65 @@ namespace vsg
...
         ref_ptr<DescriptorSetLayout> descriptorSetLayout;
         ref_ptr<DescriptorBuffer> descriptor;
         ref_ptr<DescriptorSet> descriptorSet;
 
+        // shadow map hints
+        double maxShadowDistance = 1e8;
+        double shadowMapBias = 0.005;
+        double lambda = 0.5;
+
+        // Shadow backend.
```
LeeRiva, line 154: lower-case comment?

src/vsg/app/CommandGraph.cpp
```diff
@@ -18,6 +18,10 @@ THE SOFTWARE IS PROVIDED "AS IS", WITHOUT WARRANTY OF ANY KIND, EXPRESS OR IMPLI
 
 using namespace vsg;
 
+CommandGraph::CommandGraph()
```
LeeRiva, line 21: Could this be simply left as =default; in the header?

src/vsg/state/ViewDependentState.cpp
```diff
@@ -103,41 +142,332 @@ void BindViewDescriptorSets::compile(Context& context)
...
+            rendergraph->framebuffer = fbuf;
+
+            rendergraph->clearValues.resize(1);
+            rendergraph->clearValues[0].depthStencil = VkClearDepthStencilValue{0.0f, 0};
+
+            ++layer;
+        }
+
+        // use an image barrier to transition the initial shadow map array layout to VK_IMAGE_LAYOUT_DEPTH_STENCIL_READ_ONLY_OPTIMAL
+        // so that whole shadow map is usable in fragment shader even when only portions of it have been set using a render to texture pass
```
LeeRiva, line 456: so that THE whole shadow..

src/vsg/app/RecordTraversal.cpp
```diff
@@ -484,7 +485,14 @@ void RecordTraversal::apply(const CommandGraph& commandGraph)
...
     {
         auto cg = const_cast<CommandGraph*>(&commandGraph);
-        cg->record(recordedCommandBuffers, _frameStamp, _databasePager);
+        if (cg->device)
+        {
+            cg->record(recordedCommandBuffers, _frameStamp, _databasePager);
+        }
+        else
+        {
+            warn("RecordTraversal::apply(const CommandGraph& commandGraph) cannot traverse as commandGraph->device = ", cg->device);
```
LeeRiva, line 494: cg->device would be always be nullptr here, so maybe simplify that warning by getting rid of the (always-nullptr) parameter and write that directly into the text string?

LeeRiva (commented):
I don't claim to really understand how shadows work but I noticed some trivial things, so did this unsolicited review. Very excited about that change!

src/vsg/state/ViewDependentState.cpp
```diff
@@ -103,41 +142,332 @@ void BindViewDescriptorSets::compile(Context& context)
...
+            dependencies[0].dstSubpass = 0;
+            dependencies[0].srcStageMask = VK_PIPELINE_STAGE_FRAGMENT_SHADER_BIT;
+            dependencies[0].dstStageMask = VK_PIPELINE_STAGE_EARLY_FRAGMENT_TESTS_BIT;
+            dependencies[0].srcAccessMask = VK_ACCESS_SHADER_READ_BIT;
+            dependencies[0].dstAccessMask = VK_ACCESS_DEPTH_STENCIL_ATTACHMENT_WRITE_BIT;
+            dependencies[0].dependencyFlags = VK_DEPENDENCY_BY_REGION_BIT;
+
+            // This is the heart of what makes Vulkan offscreen rendering
+            // work: render passes that follow are blocked from using this
+            // passes' color attachment in their fragment shaders until all
```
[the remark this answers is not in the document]
robertosfield: old hang up started with example as starting place for the render to texture, so have deleted this comment block.

src/vsg/app/CommandGraph.cpp
robertosfield: It could, but it doesn't really gain anything.  I like have even empty code blocks there so it's easy to add/remove debug output from the constructors.

PR #989, opened 2023-10-04
Prevent self-destructing build system

AnyOldName3:
## Description

Prevent clobber and uninstall targets running when pressing the *Build Solution* hotkey in Visual Studio.

This function is *usually* analogous to `make all`, but there's a slight difference with when transitive dependencies of things get built, so for consistency, CMake creates an `ALL_BUILD` VS project in the VS solution that does exactly what `make all` would, and leaves the hotkey doing what it normally does.

For under-the-hood reasons, targets that are excluded from all (the `EXCLUDE_FROM_ALL` target property, on by default for custom targets) have `EXCLUDE_FROM_DEFAULT_BUILD` implicitly off if they're a dependency of another target, and implicitly on if they're not. This means that using *Build Solution* skips `uninstall` and `clobber`, but runs `uninstall-vsg` (which will fail if an install hasn't happened yet as `install_manifest.txt` won't exist) and `clobber-vsg` (which wipes out the build system, causing all kinds of havoc).

Explicitly setting `EXCLUDE_FROM_DEFAULT_BUILD` for the overall and project-specific `uninstall` and `clobber` targets prevents them from running when they're not explicitly requested. Unfortunately, thanks to VS being dumb here, this also applies to when you run the overall target - you didn't explicitly request its project-specific dependencies, so they aren't built. You can get around this by using `cmake --build . --target uninstall` instead of running the uninstall project from the VS GUI, though, so it's not a *huge* problem like a surprise `git clean -dxf` can be when the `clobber` target runs unexpectedly.

The only alternatives I've thought of would be entirely removing the `clobber` target (`git clean -dxf` isn't a *wildly* difficult command to remember, and can be a pretty big problem if run accidentally, which is more likely when there's a build system target for it) and modifying the uninstall target so it doesn't fail the whole build, or adding CMake option(s) to disable them (and potentially defaulting it to true when using a VS generator).

Fixes N/A - there's nothing on the tracker for this.

## Type of change

- [x] Breaking change (fix or feature that would cause existing functionality to not work as expected)

## How Has This Been Tested?

I've built the project a bunch of times.

**Test Configuration**:
* Firmware version: N/A
* Hardware: N/A
* Toolchain: VS 17 2022, CMake 3.27.4
* SDK: N/A

## Checklist:

### DIFF
--- a/cmake/vsgMacros.cmake
+++ b/cmake/vsgMacros.cmake
@@ -280,11 +280,12 @@ macro(vsg_add_target_clobber)
         else()
             if (NOT TARGET clobber)
                 add_custom_target(clobber)
+                set_target_properties(clobber PROPERTIES EXCLUDE_FROM_DEFAULT_BUILD TRUE)
             endif()
             add_custom_target(clobber-${PROJECT_NAME}
                 COMMAND ${GIT_EXECUTABLE} -C ${PROJECT_SOURCE_DIR} clean -d -f -x
             )
-            set_target_properties(clobber-${PROJECT_NAME} PROPERTIES FOLDER "${PROJECT_NAME} Folder")
+            set_target_properties(clobber-${PROJECT_NAME} PROPERTIES FOLDER "${PROJECT_NAME} Folder" EXCLUDE_FROM_DEFAULT_BUILD TRUE)
             add_dependencies(clobber clobber-${PROJECT_NAME})
         endif()
     endif()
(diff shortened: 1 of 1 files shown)


robertosfield:
Other users have been working with VisualStudio without major problems.  Are you doing something different or using a different build combination?  Is the "Build Solution" a new feature that others aren't using?

I would like to know what is going with other users vs what you are doing to know whether this is local issue or a general issue that others just workaround.

robertosfield:
It would seem to be me that if we do include EXCLUDE_FROM_DEFAULT_BUILD TRUE entries, then perhaps a helper macro might be appropriate, as there are a range of build targets meant to be run explicitly and not  done by default.

AnyOldName3:
I've built VSG with this build system before without issue, but I was doing so from the command-line, where CMake was running its own `ALL_BUILD` target instead of VS' 'default build', and for tinkering with the examples, I was setting individual projects as the debugging target, so only them and their direct dependencies were being built. I don't think there's any particular reason why I did a *Build Solution* today and not in the past - it's as likely as anything that I just pressed the hotkey because it's faster than finding an individual project in the IDE, right-clicking it, and building it individually, especially if you've only changed a few files, so it doesn't need to rebuild everything (and the build system doesn't delete itself so it has to start again). The button and hotkey have been there as long as I've been using VS, and the CMake option controlling it looks like it was added over a decade ago, so it's not a new feature.

MSVC is much easier to use from the command-line than it once was, especially for CMake projects, so it might just be that most Windows developers working on VSG stuff *aren't* using VS as their IDE, and those that are are just building individual targets and their direct dependencies most of the time instead of everything at once.

robertosfield:
I have now had time to do another review and see the changes are made to the macros so will help the other projects as well.  I'll merge as is, but it looks to me like other targets would be appropriate to add as well.

PR #1003, opened 2023-10-24
Don't create descriptor set bindings with null data

timoore:
In GraphicsPipelineConfigurator::copyTo(), don't create a BindDescriptorSet command for descriptor sets that contain null buffer or image data, and emit a warning if a descriptor set contains a mix of null values and valid data. The common case with GraphicsPipelineConfigurator is to bind a pipeline and all the necessary descriptor sets at the same time (in the same StateGroup); this commit supports binding some descriptor sets earlier than the pipeline by not rebinding them with some default value.

### DIFF
--- a/src/vsg/utils/GraphicsPipelineConfigurator.cpp
+++ b/src/vsg/utils/GraphicsPipelineConfigurator.cpp
@@ -10,6 +10,7 @@ THE SOFTWARE IS PROVIDED "AS IS", WITHOUT WARRANTY OF ANY KIND, EXPRESS OR IMPLI
 
 </editor-fold> */
 
+#include <vsg/core/visit.h>
 #include <vsg/io/Logger.h>
 #include <vsg/io/Options.h>
 #include <vsg/nodes/StateGroup.h>
@@ -483,6 +484,60 @@ void GraphicsPipelineConfigurator::init()
     bindGraphicsPipeline = vsg::BindGraphicsPipeline::create(graphicsPipeline);
 }
 
+// Determine if a descriptor set contains null values and therefore should not be written, updated,
+// or bound. This might conflict someday with Vulkan features and extensions for null descriptors.
+
+class DescriptorSetContainsNull : public ConstVisitor
+{
+public:
+    bool containsNull = false;
+    void apply(const DescriptorSet& ds) override
+    {
+        if (ds.descriptors.empty())
+        {
+            containsNull = true;
+        }
+        else
+        {
+            ds.traverse(*this);
+        }
+    }
+
+    void apply(const DescriptorBuffer& db) override
+    {
+        containsNull = containsNull || db.bufferInfoList.empty();
+    }
+
+    void apply(const DescriptorImage& di) override
+    {
+        containsNull = containsNull || di.imageInfoList.empty();
+    }
+    // DescriptorTexelBufferView?
+};
+
+// The inverse, mainly for issuing a warning if a descriptor set contains nulls and values
+
+class DescriptorSetContainsData : public ConstVisitor
+{
+public:
+    bool containsData = false;
+    void apply(const DescriptorSet& ds) override
+    {
+        ds.traverse(*this);
+    }
+
+    void apply(const DescriptorBuffer& db) override
+    {
+        containsData = containsData || !db.bufferInfoList.empty();
+    }
+
+    void apply(const DescriptorImage& di) override
+    {
+        containsData = containsData || !di.imageInfoList.empty();
+    }
+    // DescriptorTexelBufferView?
+};
+
 void GraphicsPipelineConfigurator::copyTo(ref_ptr<StateGroup> stateGroup, ref_ptr<SharedObjects> sharedObjects)
 {
     // create StateGroup as the root of the scene/command graph to hold the GraphicsPipeline, and binding of Descriptors to decorate the whole graph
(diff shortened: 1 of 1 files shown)


robertosfield:
Do you have an example that illustrates the usage case that this PR addresses?

I would like to step through the usage case so I can't see the purpose behind it, it might be that some other approach should be used for this type of usage case, or a different modification to ShaderSet/GraphicsPipelineConfigurator might be more appropriate.

timoore:
The use case is the fast / slow descriptor set scheme of vsgCs. Some descriptor sets defined by the shader set are bound higher in the scene graph than the state group that binds the pipeline; if a default DS is bound with the pipeline, then that will hide the correct binding. So, I need some way to say “don’t bind a DS”.

vsgCs doesn’t use the pipeline configurator in this way, but if I want to load vsgXchange models and use the same shaders, or at least the same DS scheme, then I need this functionality in GraphicsPipelineConfigurator.

robertosfield:
Is there a vsgCs example that adds a loaded model like the usage case you describe?

From your usage case description it sounds like we may want a way of specifying that descriptor sets will be assigned elsewhere.  I'd rather an explicit scheme for this than one that works implicitly based on knowledge of an internal implementation detail.

robertosfield:
I have merged this PR as the branch:

[https://github.com/vsg-dev/VulkanSceneGraph/tree/timoore-empty-descriptor-sets](https://github.com/vsg-dev/VulkanSceneGraph/tree/timoore-empty-descriptor-sets)

I will use this in testing of the vsgCs style DescriptorSet usage with the aim for coming up with a more explicit scheme to specifying when ShaderSet/GraphicsPipelineConfigurator should use inherited BindDescriptorSet rather than defining one locally when setting up a subgraph.

src/vsg/utils/GraphicsPipelineConfigurator.cpp
```diff
@@ -496,18 +551,28 @@ void GraphicsPipelineConfigurator::copyTo(ref_ptr<StateGroup> stateGroup, ref_pt
         {
             if (auto ds = descriptorConfigurator->descriptorSets[set])
             {
-                if (sharedObjects)
+                if (visit<DescriptorSetContainsNull>(ds).containsNull)
                 {
-                    sharedObjects->share(ds);
+                    if (visit<DescriptorSetContainsData>(ds).containsData)
```
robertosfield, line 556: Is this checked for a case where one DescriptorBuffer/DescriptorImage is empty so triggers the first condition, but there is another DescriptorBuffer/DescriptorImage on the same DesciptorSet that isn't empty?

Is this not a condition where defaults is the appropriate to fill in the missing entries rather than have valid entries just ignored?

Or is this really a case of an invalid application usage?

timoore: > Is this checked for a case where one DescriptorBuffer/DescriptorImage is empty so triggers the first condition, but there is another DescriptorBuffer/DescriptorImage on the same DesciptorSet that isn't empty?

Yes.
> 
> Is this not a condition where defaults is the appropriate to fill in the missing entries rather than have valid entries just ignored?

In this case there aren't any defaults, or they would already be in the list of descriptors. In other words, the only way a descriptor can have an empty buffer or image info list is if the shader set does not provide a default, or the user assigns such an empty descriptor.

> 
> Or is this really a case of an invalid application usage?

The comment alludes to the possibility of null as a valid value if certain features and extensions are enabled, but I haven't yet wrapped my around the functionality that's actually provided.

The motivation for this whole ceremony is that the descriptor set layout for the pipeline needs to be constructed, but the user may leave some descriptor sets empty. I submitted ShaderSet::createPipelineLayout() in another PR which builds the layout directly from the list of defines. I might be useful here.

robertosfield: I wonder if a debug method of some kind might be helpful to give users better insight into how the ShaderSet/GraphicsPipelineConfigurator usage is incomplete.  I think it's a bit premature for me to start looking at this thought, first I need to get the extra controls we need for the vsgCs usage approach working.

robertosfield:
FYI, I have started writing a vsgcustomshaderset example for vsgExamples that will test out the DescriptorSet hierarchy used in vsgCs.

My current feeling is that we should hint via vsg::Options and/or GraphicsPipelineConfigurator that there are StateCommands's that we can assume will be inherited from StateGroup higher up in the scene graph to where the newly created subrgaph will be added.  

So BindViewDependentState and other BindDescriptorSets or BindGraphicsPipeline can be specified as being set higher up in the scene graph so when a loader/model builder creates a subgraph the GraphicsPipelineConfigurator can choose not to create/assign the associated element if those inherited elements are compatible.

robertosfield:
I am working on the state inheritance support in ShaderSet/GraphicsPipelineConfigurator and while this work isn't complete yet when getting the new vsgExamples [vsgcustomshaderset](https://github.com/vsg-dev/vsgExamples/tree/CustomShaderSet) to work cleanly I've had to implement support creating empty DescrptorSetLayout and not creating/assigning DescriptorSet that aren't needed for a particular ShaderSet/GraphicsPipelineConfigurator usage.

The work is checked into VSG [StateInhertance](https://github.com/vsg-dev/VulkanSceneGraph/tree/StateInheritance) branch, and changes so far are:

https://github.com/vsg-dev/VulkanSceneGraph/compare/master...StateInheritance

I'm thinking this provides the capability of a ShaderSet declaring both sparse and unassigned DescriptorSet entries, which overlaps with what this PR implements.  It tackles it in a different way - higher up in the DescriptorConfigurator so avoids any unnecessary DescritporSet creations so should be more efficient.

This StateInheritance is probably is a far enough state to merge with VSG master, though probably makes sense to do some more testing.  This work isn't complete though as I haven't implemented the state inheritance yet. all the functionality so far is building towards this.

@timoore thoughts?  Will this be sufficient for vsgCs and so could we close this PR?

timoore:
I'll look at this soon and see if it does the trick for me.

One thing that is important for my usage is that the descriptor set layout be correct, even if there empty descriptor sets. That's why I submitted ShaderSet::createPipelineLayout(), which doesn't use DescriptorConfigurator at all to build the pipeline layout.

 Another point is that the pipeline layout is only partly driven by the bindings used by a particular shader program. If there's a layout(set=...,binding=...) declaration, then there has to be a corresponding entry in the pipeline layout, but the converse is not true; it is legal for there to be bindings declared by the layout which don't appear in the shader at all.

robertosfield:
Thanks for mentioning ShaderSet::createPipelineLayout(..), I forgotten about it, that will be useful for the vsgcustomshaderset. I manually set up the top most BIndVIewDependentState's pipeline layout so this ShaderSet::createPipelineLayout(..) would probably be a clean way of doing it.

Today I'll add support for setting the state that will be inherited by a subgraph so doesn't need to be set locally. This inheritance will need to check compatibility of pipeline and descriptor set layouts.

robertosfield:
I have added use of the ShaderSet::createPipelineLayout() in the vsgcustomshader example, still learning/adapting code as I go.  This is certainly cleaner than what I originally had.

https://github.com/vsg-dev/vsgExamples/tree/CustomShaderSet

I haven't yet implemented inheritance yet, control over whether sharing or uniform or keeping a uniform/texture unique is something I'll look at as well.  These changes I'll chack in the StateInheritance branch.

robertosfield:
I am closing this PR as it should no longer be necessary thanks to the StateInheritance work.

PR #1006, opened 2023-10-25
Change assignDescriptor and assignTexture to template functions

timoore:
This makes all the overloads provided by DescriptorConfigurator available to users of GraphicsPipelineConfigurator.

### DIFF
--- a/include/vsg/utils/GraphicsPipelineConfigurator.h
+++ b/include/vsg/utils/GraphicsPipelineConfigurator.h
@@ -113,8 +113,22 @@ namespace vsg
         bool enableTexture(const std::string& name);
 
         bool assignArray(DataList& arrays, const std::string& name, VkVertexInputRate vertexInputRate, ref_ptr<Data> array);
-        bool assignDescriptor(const std::string& name, ref_ptr<Data> data = {});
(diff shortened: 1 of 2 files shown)


timoore:
Ack, this is quite broken... I didn't understand exactly how create_if works. Will push a fix later.

robertosfield:
create_if(bool, params) will return an {} if bool is false.  The original if (!object) object = Object::create() style code was appropriate so just reinstate that in your template code.

timoore:
Yeah. vsgCs has a
```
    template<typename T, typename TSource, typename... Args>
    vsg::ref_ptr<T> create_or(const vsg::ref_ptr<TSource>& source, Args&&... args)
    {
        auto downcast = ref_ptr_cast<T>(source);
        if (downcast.valid())
        {
            return downcast;
        }
        return T::create(args...);
    }
```

which I was thinking of.

robertosfield:
Perhaps if these convenience methods can cause confusions we should have them and just lets folk explicitly set stuff.

robertosfield:
I have implemented the functionality of this PR by just implementing the missing functions so will now close this PR. 

The changes is: f7b412ad1d7125cf7ea58deeb558ac821ebc8d68

PR #1014, opened 2023-11-03
Always check if the overriding multisample state is correct for pass

timoore:
With shadows enabled and 4 samples in the main framebuffer, a Vulkan validation error was emitted: pipelines were being created whose multisample state was not compatible with the renderpass used to create the pipeline.

### DIFF
--- a/src/vsg/app/CompileTraversal.cpp
+++ b/src/vsg/app/CompileTraversal.cpp
@@ -247,6 +247,21 @@ void CompileTraversal::apply(Geometry& geometry)
     geometry.traverse(*this);
 }
 
+namespace
+{
+    ref_ptr<MultisampleState> getContextMSState(const ref_ptr<Context>& context)
+    {
+        for (const auto& state : context->overridePipelineStates)
+        {
+            if (auto overideMs = state.cast<MultisampleState>())
+            {
+                return overideMs;
+            }
+        }
+        return {};
+    }
+}
+
 void CompileTraversal::apply(CommandGraph& commandGraph)
 {
     auto traverseRenderedSubgraph = [&](vsg::RenderPass* renderpass, VkExtent2D renderArea) {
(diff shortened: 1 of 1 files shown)


robertosfield:
What would the best way to reproduce the problem that this PR addresses?

timoore:
On Fri, Nov 3, 2023 at 2:51 PM Robert Osfield ***@***.***>
wrote:

> What would the best way to reproduce the problem that this PR addresses?
>
> Run with shadows and multisampling enabled and look at the debug
(validation) output.

robertosfield:
vsgshadow example didn't have support for enabling shadows so I've enabled this with the addition of a --samples <n> command line parsing and can confirm the Vulkan debug error appears:

   Objects: 3
        [0] 0x55f3ee62e850, type: 6, name: NULL
        [1] 0x59f7450000000038, type: 19, name: NULL
        [2] 0x27d60e0000000019, type: 18, name: NULL
VUID-vkCmdDrawIndexed-multisampledRenderToSingleSampled-07284(ERROR / SPEC): msgNum: 1466193060 - Validation Error: [ VUID-vkCmdDrawIndexed-multisampledRenderToSingleSampled-07284 ] Object 0: handle = 0x55f3ee62e850, type = VK_OBJECT_TYPE_COMMAND_BUFFER; Object 1: handle = 0x59f7450000000038, type = VK_OBJECT_TYPE_PIPELINE; Object 2: handle = 0x27d60e0000000019, type = VK_OBJECT_TYPE_RENDER_PASS; | MessageID = 0x576454a4 | vkCmdDrawIndexed():  In VkPipeline 0x59f7450000000038[] the sample count is VK_SAMPLE_COUNT_8_BIT while the current VkRenderPass 0x27d60e0000000019[] has VK_SAMPLE_COUNT_1_BIT and they need to be the same. The Vulkan spec states: If rasterization is not disabled in the bound graphics pipeline, and none of the VK_AMD_mixed_attachment_samples extension, the VK_NV_framebuffer_mixed_samples extension, or the multisampledRenderToSingleSampled feature is enabled, then rasterizationSamples for the currently bound graphics pipeline must be the same as the current subpass color and/or depth/stencil attachments (https://vulkan.lunarg.com/doc/view/1.3.268.0/linux/1.3-extensions/vkspec.html#VUID-vkCmdDrawIndexed-multisampledRenderToSingleSampled-07284)

Once I've wrapped up my work on StateInheritance I'll check out this PR and do proper review.

robertosfield:
As a heads up, I'm currently reviewing and testing the combination of multi-sampling with shadows so should get either this PR or an alternate solution merged this morning.

robertosfield:
I have come up with a fix for nested RenderGraph/RenderPass having a different number of samples, mostly it required removing code!

The fix is checked into VSG master: c89774cb8d051d7d31c100b30a14965161a81bec

With this fix _vsgshadow -d --samples 8_ runs cleanly.

Could you update to VSG master and let me know if it now runs without any errors.  I'll close this PR now, if there are still issues I can investigate based on what the above commit fails to catch.

PR #1030, opened 2023-11-17
Store the ImageInfo for font textures in the font

timoore:
This is in place of only storing the Data for the font texture and relying on SharedObjects to find an already created font texture. The font glyph metrics weren't stored as data at all, but were recreated on every creation of a GpuLayoutTechnique.

At the moment this change results in a massive memory savings in complex scenes. There seems to be a bug in SharedObjects in that it can't pick up the images in a font if they are recreated on each use, resulting in gigabytes of extra texture use. Even if SharedObjects is fixed, it is cleaner to store the images in the font.

The glyph metrics texture should probably be a storage buffer instead.

If this is accepted, then the code in vsgXchange that creates a font using freetype should probably change to call createFontImages().

### DIFF
--- a/include/vsg/text/Font.h
+++ b/include/vsg/text/Font.h
@@ -14,6 +14,7 @@ THE SOFTWARE IS PROVIDED "AS IS", WITHOUT WARRANTY OF ANY KIND, EXPRESS OR IMPLI
 
 #include <vsg/core/Data.h>
 #include <vsg/io/Options.h>
+#include <vsg/state/ImageInfo.h>
 #include <vsg/text/GlyphMetrics.h>
 #include <vsg/utils/SharedObjects.h>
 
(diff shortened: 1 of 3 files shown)


robertosfield:
Changes look good.  It looks like createFontImages() is already created when needed, so what would the benefit of vsgXchange::freetype calling it explicitly?

timoore:
I was worried about multithreading issues and didn't want to add a mutex to the font or something, and so thought it best if the font-building code completely constructed the images.

robertosfield:
It is indeed safer to create as much up front before objects are shared to prevent possible race conditions.  

There remains an open ended question of how much of the compile and record traversals need objects protected to prevent race conditions, so far I have left it high level coordinates of threads to prevent them stepping on each others toes with shared data, but it's fragile peace.

PR #1037, opened 2023-11-28
[#1035] glslang support

### LINKED ISSUE #1035: Inability to compile vsg with gslang support in build environments without network access
**Describe the bug**
glslang support in vsg seems tied to a specific fork of glslang (upstream + patches). This is problematic for several reasons:

- packaging vsg downstream with glslang support is impossible as build environments have no network access
- downstream's glslang do not have those patches
- projects that require glslang support can't also be packaged due to lack of glslang support
- vsg's cmake requires git to pull in the fork then hooks into it via cmake in an odd way

**Expected behavior**

vsg should be making use of the system glslang, and if not available but still required, then the project should make use of cmake's FetchContent: https://cmake.org/cmake/help/latest/module/FetchContent.html
FetchContent is more secure as it also requires a hash of the release or commit version you wish to download to limit supply-chain attacks.

**Additional context**
How necessary are the patches to glslang and why are they needed?
It is possible to use FetchContent to apply patches so there is no need to fork.

Being able to package vsg with glslang support Debian and Ubuntu is currently blocked. 

Building projects, like vsgopenmw, using flatpak are also blocked as they also do not have network access while in the build environment.

psi29a:
# Pull Request Template

## Description

Adds support for selecting use of system glslang (by default) or using cmake's FetchContent
glslang compilation support is now by default available

Fixes #1035 

## Type of change

## How Has This Been Tested?

- [ ] Validated to work with system installed lib: Ubuntu and macOS (brew)
- [x] Validated with compiled from source via FetchContent

## Checklist:

### DIFF
--- a/.github/workflows/ci-android.yml
+++ b/.github/workflows/ci-android.yml
@@ -4,7 +4,7 @@ on:
   pull_request:
 env:
   BuildDocEnabled: ${{ github.event_name == 'push' && github.ref == 'refs/heads/master' }}
-  CMakeVersion: 3.10.x
+  CMakeVersion: 3.24.x
 jobs:
   build:
     runs-on: ${{ matrix.os }}
@@ -18,7 +18,7 @@ jobs:
     steps:
     - uses: actions/checkout@v2
     - name: Setup cmake
-      uses: jwlawson/actions-setup-cmake@v1.12
+      uses: jwlawson/actions-setup-cmake@v1.14.1
       with:
         cmake-version: ${{ env.CMakeVersion }}
     - name: Setup NDK  
--- a/.github/workflows/ci-android.yml
+++ b/.github/workflows/ci-android.yml
@@ -28,5 +28,5 @@ jobs:
     - name: Build and Install VSG
       shell: bash
       run: |
-        cmake . -DCMAKE_TOOLCHAIN_FILE=$(dirname $(which ndk-build))/build/cmake/android.toolchain.cmake -DANDROID_ABI=${{matrix.android-abi}} -DANDROID_PLATFORM=${{matrix.android-platform}} -DBUILD_SHARED_LIBS=${{matrix.build-shared}} -DCMAKE_INSTALL_PREFIX=./install
+        cmake . -DCMAKE_TOOLCHAIN_FILE=$(dirname $(which ndk-build))/build/cmake/android.toolchain.cmake -DANDROID_ABI=${{matrix.android-abi}} -DANDROID_PLATFORM=${{matrix.android-platform}} -DBUILD_SHARED_LIBS=${{matrix.build-shared}} -DCMAKE_INSTALL_PREFIX=./install -DVSG_USE_SYSTEM_GLSLANG=0
         cmake --build . --target install --config Release
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -4,7 +4,7 @@ on:
   pull_request:
 env:
   BuildDocEnabled: ${{ github.event_name == 'push' && github.ref == 'refs/heads/master' }}
-  CMakeVersion: 3.10.x
+  CMakeVersion: 3.24.x
 jobs:
   build:
     runs-on: ${{ matrix.os }}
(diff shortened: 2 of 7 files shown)


timoore:
Why does your commit remove VSG_SUPPORTS_ShaderCompiler? That's an important requirement for some use cases.

psi29a:
> Why does your commit remove VSG_SUPPORTS_ShaderCompiler? That's an important requirement for some use cases.

Yeah, change ABI is bad, but look at this way, it's always on now, so no need test for its support.

@robertosfield what should the policy for deprecation be?

timoore:
> > Why does your commit remove VSG_SUPPORTS_ShaderCompiler? That's an important requirement for some use cases.
> 
> Yeah, changing ABI out of the blue is bad, but look at this way, it's always on now, so no need test for its support.
>

My point is this: why do you think it should be deprecated, especially as part of a PR that does something else? Running VSG in an environment where there are only precompiled SPIRV files available should be supported.

robertosfield:
A case of one step forward, ten steps back.  This is an easy reject.

Forcing an update to a much later CMake version is something I really want to avoid unless we *really* have to.  This change could break the build where the VSG compiles fine yet now.  It's changes like this that could create a lot of work for us downstream.

The change for VSG_SUPPORTS_ShaderCompiler is really bad too.  We still want to be able to compile the VSG without glslang.  It's perfectly possible to use the VSG without glslang, it's just some usage cases that rely upon shader compilation.

I understand the desire to make things more flexible but this is huge step back in flexibility.

.github/workflows/ci-android.yml
```diff
@@ -4,7 +4,7 @@ on:
   pull_request:
 env:
   BuildDocEnabled: ${{ github.event_name == 'push' && github.ref == 'refs/heads/master' }}
-  CMakeVersion: 3.10.x
+  CMakeVersion: 3.24.x
```
timoore, line 7: If you do this, shouldn't you change the minimum version in CMakeLists.txt?

timoore:
Just a sidebar: I don't think a recent version of CMake is an onerous requirement. I just tried installing CMake on Ubuntu 22.04.2 using snap and it happily installed cmake v3.27. 8. The FetchContent support is better in 3.24 and directly supports using find_package as an alternative to downloading and building the content.

I also don't think that there's anything inherently wrong with building against a custom version of glslang, but I would prefer using FetchContent to pull it from your repo.

robertosfield:
Any extra hurdle that folks need to jump has to be judged by the benefit it provides.  Slightly better FetchContent support is not enough reason. find_package with glslang has been problematic in the past as well so it's a potentially reintroducing more problems than is solves.  We have to be very careful about any of these changes.  As I've said too often now, this is the far from the first time around trying to get glslang to work cleanly, trying stuff I've already tried and failed with is really not helpful at moving us forward.

psi29a:
Why reject? Just say what needs to change? :)

.github/workflows/ci-android.yml
psi29a: I can, yes.

robertosfield:
I tried FetchContent before I tried just pulling stuff in directly.  I resorted to the solution we have because FetchContent and the other solutions I tried didn't work well enough in all the different situations that users use the VSG.  How many times do I need to say the same thing?  I've spent many days trying many different combinations of ways to get glslang to work well within the VSG.

I don't like having glslang built as part of the VSG, it's cludge to get around the fact that gslang was breaking build/runtime on too many different user systems.

The only thing that might change things is if the glslang project itself changes things - allow us to specify the version requirements easily and robustly without breaking when an older, problematic in various ways, version is available but not usable.  If glslang starts behaving itself sufficiently as a standalone project then we can ditch the internal glslang completely and just use find_package().

psi29a:
If you'd prefer `find_package`, that's fine too as it works fine on macos and ubuntu. The odd one out would be windows I think.

What bothers me is that there is an implicit attempt to grab glslang and if it can't do that, it's ignored. It's a best effort attempt.

Can this go away? 

Or do you still wish to do a best-effort attempt to fetch it if it's not found?

I'm asking because I'd prefer it to be an optional param that is off by default, as you said, some projects don't want it. Rather be explicit than implicit.

robertosfield:
If the attempt to pill in glslang fails it very much isn't ignored.

The VSG sets #define about support for glslang in the automatically generated headers so it can be checked at compile time, as well as a function that enables a check at runtime.  As Tim pointed out your PR actually broke this...

If the glslang project has fixed the problems that prevent using it via find_package then we may be able to ditch our internal build and just link to it as we do Vulkan.

PR #1038, opened 2023-11-28
Debug utils extension

timoore:
Add support for the VK_EXT_debug_utils instance extension.

### DIFF
--- a/include/vsg/vk/Instance.h
+++ b/include/vsg/vk/Instance.h
@@ -78,7 +78,18 @@ namespace vsg
             if (pNameFallback) procAddress = reinterpret_cast<T>(vkGetInstanceProcAddr(_instance, pNameFallback));
             return (procAddress);
         }
-
+        // VK_EXT_debug_utils
+        PFN_vkSetDebugUtilsObjectNameEXT vkSetDebugUtilsObjectNameEXT = nullptr;
+        PFN_vkSetDebugUtilsObjectTagEXT vkSetDebugUtilsObjectTagEXT = nullptr;
+        PFN_vkQueueBeginDebugUtilsLabelEXT vkQueueBeginDebugUtilsLabelEXT = nullptr;
+        PFN_vkQueueEndDebugUtilsLabelEXT vkQueueEndDebugUtilsLabelEXT = nullptr;
+        PFN_vkQueueInsertDebugUtilsLabelEXT vkQueueInsertDebugUtilsLabelEXT = nullptr;
+        PFN_vkCmdBeginDebugUtilsLabelEXT vkCmdBeginDebugUtilsLabelEXT = nullptr;
+        PFN_vkCmdEndDebugUtilsLabelEXT vkCmdEndDebugUtilsLabelEXT = nullptr;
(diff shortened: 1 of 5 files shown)


include/vsg/vk/Instance.h
```diff
@@ -78,7 +78,18 @@ namespace vsg
             if (pNameFallback) procAddress = reinterpret_cast<T>(vkGetInstanceProcAddr(_instance, pNameFallback));
             return (procAddress);
         }
-
+        // VK_EXT_debug_utils
```
robertosfield, line 81: Is there a reason why these extensions aren't set up as part of vsg::Extentsions?

https://github.com/vsg-dev/VulkanSceneGraph/blob/master/include/vsg/vk/Extensions.h

timoore:
> Is this changes required to work with RenderDoc?

Yes. I didn't include a scene graph node to record text for RenderDoc et al., thinking that it would go in vsgExamples, but perhaps a utility class should go in core VSG.

include/vsg/vk/Instance.h
timoore: The only reason is that it's an instance extension. If you don't care about the distinction (and not sure what the difference is on practice), then the pointers could go in vsg: Extensions. I guess that device extensions are particular to one device.

robertosfield: I will have think about where best to place things.  It may be appropriate to just create Instance extensions and Device extensions structuress.

robertosfield:
If you create a vsgExample test program with the custom node I can test this out an make decision whether to merge it into core VSG or not.

Tomorrow I'll merge this PR as branch and have a tinker.

robertosfield:
I have merged this PR as a branch:

https://github.com/vsg-dev/VulkanSceneGraph/tree/timoore-debug-utils

I will now review the changes.  Some instructions/vsgExample on how to test it with RenderDoc would be helpful.

robertosfield:
I have made set of changes to the [timoore-debug-utils](https://github.com/vsg-dev/VulkanSceneGraph/tree/timoore-debug-utils) branch.  

The key one is moving the instance extensions out from vsg::Instance into a dedicated vsg::InstanceExtensions class like the existing vsg::Extensions class.  Once we merge this work I think vsg::Extensions should be renamed to vsg::DeviceExtensions to make things clear what is what.

When setting up the functions I have dropped the check to see if the extension is supported or not when setting up the extensions, the simplifies the code and I don't think there is undue risk in have these setup.

robertosfield:
While my system has been building the automated build has picked up on Windows build issues that I' hopefully have fixed now, but I await the automated build giving it the thumbs up.  If that all looks OK I'll be happy enough to merge with VSG master as it mostly looks uncontroversial.  I would rather have something to test against though.

robertosfield:
The branch now has a clean build so I'll close this PR and then generate a new PR from the branch.

timoore:
I was just in the process of making a pull request in vsgExamples using my original version, but I'll back up and do it with your latest.

robertosfield:
Thanks, my new PR is #1042.

The only changes you should need to do in the form:

auto extensions = instance->getExtensions();
extension->callMyFunc(...);

At least that's what I'm hoping :-)

PR #1042, opened 2023-11-29
Added vsg::InstanceExtensions and support for VK_EXT_debug_utils

robertosfield:


timoore:
With your change that always initializes the function pointers, how does client code efficiently test if the extension is supported, or desired? In my example usage, which you can see in https://github.com/timoore/vsgExamples/tree/annotation, the annotation feature is a noop if the extension wasn't requested. I guess a careful reading of the spec suggests that the function pointer will be null if the instance extension isn't supported.

robertosfield:
Just had a quick look at the example.  What value does the template helper class provide over just implementing a decorator node that does the required extension calls?

For timing stats I have the same thing in mind.  A decorator that we can insert above any subgraphs we want to track.

robertosfield:
I have merged the timoore-debug-utils branches of VSG and vsgExamples.

@timoore I think it would be worth letting the community know that this capability is now available, I could do this but as you're the main author and know more about the detail of RenderDoc you might be better placed to announce it.

timoore:
I'll write something for the discussion group.

PR #1114, opened 2024-03-05
Support per-light angle subtended

AnyOldName3:
Adds an extra parameter to directional lights, and exposes it as part of shadow settings so soft shadow implementations can use it.


AnyOldName3:
Here are two directional lights with different penumbra sizes:

image

robertosfield:
@AnyOldName3 this PR was missing version guards so I've added them with this commit: 321c1e2ed4087b347907eaf02998cddb6c23d59b

The version guards are essentially for retaining backwards compatibility of .vsgb and .vsgt files.

PR #1186, opened 2024-05-07
Set version to 1.1.4 so I can do feature test macros for physics allocator affinity

AnyOldName3:
I don't know if 1.1.4 is the right version as the allocator type option was removed at the same time, and that's a breaking change.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -1,7 +1,7 @@
 cmake_minimum_required(VERSION 3.7)
 
 project(vsg
-    VERSION 1.1.3
+    VERSION 1.1.4
     DESCRIPTION "VulkanSceneGraph library"
     LANGUAGES CXX
 )


robertosfield:
Bumping the version number is appropriate so I'll go ahead and merge.  I will make tag 1.1.4 dev release later this week or early next week.

If you need the version marcos then you can always just copying them to your local code and do an #ifndef to use your local version.  Once we are further down the line then you'll be able to remove this local definition.

AnyOldName3:
> If you need the version marcos then you can always just copying them to your local code and do an #ifndef to use your local version. 

That'll probably be cleaner - despite short-circuiting correctly, `#if !defined(VSG_API_VERSION_LESS) || VSG_API_VERSION_LESS(1, 1, 4)` emits lots of compiler warnings for pre-helper-macro VSG.

PR #1191, opened 2024-05-16
Fix button value in ButtonPressEvent and ButtonReleaseEvent when doub…

Mikalai:
## Description

Windows has dedicated message for double click. Old behavior with this handler 
```
class IntersectionHandler : public vsg::Inherit<vsg::Visitor, IntersectionHandler>
{
public:

    void apply(vsg::ButtonReleaseEvent& evt) override {
        std::cout << "Release button = " << evt.button << " mask = " << evt.mask << std::endl;
    }

    void apply(vsg::ButtonPressEvent& evt) override
    {
        std::cout << "Press button = " << evt.button << " mask = " << evt.mask << std::endl;
    }

};
```
is 
```
Press button = 1 mask = 256
Release button = 1 mask = 0
Press button = 0 mask = 256
Release button = 1 mask = 0
```

Where button 0 doesn't look correct value for left button double click. With this changes:
```
Press button = 1 mask = 256
Release button = 1 mask = 0
Press button = 1 mask = 256
Release button = 1 mask = 0
```
And this depicts correctly the action.

## Type of change

## How Has This Been Tested?

It was tested with handler above.

**Test Configuration**:
Windows 11

## Checklist:

### DIFF
--- a/include/vsg/platform/win32/Win32_Window.h
+++ b/include/vsg/platform/win32/Win32_Window.h
@@ -132,9 +132,13 @@ namespace vsgWin32
     {
         switch (buttonMsg)
         {
+        case WM_LBUTTONDBLCLK:
         case WM_LBUTTONDOWN: return 1;
+        case WM_MBUTTONDBLCLK:
         case WM_MBUTTONDOWN: return 2;
+        case WM_RBUTTONDBLCLK:
         case WM_RBUTTONDOWN: return 3;
+        case WM_XBUTTONDBLCLK:
         case WM_XBUTTONDOWN:
             if (wParamHi == XBUTTON1)
                 return 4;


robertosfield:
I'm not familiar with Win32 event handling (Win32_Window.cpp by @tomhog) so I need learn the background a bit before I can properly review/test Win32 submissions.    

When Windows generates the WM_LBUTTONDBLCLK etc. event does it it entirely replace the  WM_LBUTTONDOWN event? Or does it come in as well?  Of as well does it come in before or after.

Also when WM_LBUTTONDBLCLK etc. event  domes in does it get matched the WM_LBUTTONUP or similar event once the button is finally released?

This topic also touches upon whether it's time for use to add a vsg::DoubleClickEvent alongside the existing vsg::ButtonDownEvent/ButtonReleaseEvent. 

While Win32 should be straight forward to map to this type of event, other UI libraries like Xcb don't have a double click event type so to we'd need to simulate the behavior and generate the DoubleClickEvent in Xcb_Window.cpp etc to ensure user applications can check for double click events in a coherent way across platforms.

Thoughts?

Mikalai:
Regarding the behavior on windows:

> Only windows that have the CS_DBLCLKS style can receive WM_LBUTTONDBLCLK messages, which the system generates whenever the user presses, releases, and again presses the left mouse button within the system's double-click time limit. Double-clicking the left mouse button actually generates a sequence of four messages: [WM_LBUTTONDOWN](https://learn.microsoft.com/en-us/windows/win32/inputdev/wm-lbuttondown), [WM_LBUTTONUP](https://learn.microsoft.com/en-us/windows/win32/inputdev/wm-lbuttonup), WM_LBUTTONDBLCLK, and WM_LBUTTONUP.
https://learn.microsoft.com/en-us/windows/win32/inputdev/wm-lbuttondblclk

Having a dedicated vsg::DoubleClickEvent is a good idea.

robertosfield:
To help with testing features like this I have extended the vsginput example to have a --print command line option that enables the writing of the event strings out to the console, making it easier to step through what events the VSG has generated.  

https://github.com/vsg-dev/vsgExamples/commit/266faac47cbd1e4b7bd667e810b40037b0988125

PR #1195, opened 2024-05-17
only install cmake on non-window runners

sbrkopac:
…  and added control back for cmake version - windows will default to cmake provided by the runner

# Pull Request Template

## Description

I prefer this fix over with removing the variable that controls cmake completely. Completely removing it will default all runners to latest and greatest which may not be desirable for linux package managers.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -4,6 +4,7 @@ on:
   pull_request:
 env:
   BuildDocEnabled: ${{ github.event_name == 'push' && github.ref == 'refs/heads/master' }}
+  CMakeVersion: 3.10.x
 jobs:
   build:
     runs-on: ${{ matrix.os }}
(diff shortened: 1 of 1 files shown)


robertosfield:
It's not clear to me this is an improvement, it's just a different choice of defaults.  The automated builds are an early warning system for possible issues, but I don't consider them all inclusive, we still need the community to test platforms specific to their own needs.

If there is an issue with linux package managers then we can figure it out when the issues arise.  As of now the automated build is running cleanly on all the target combinations.

sbrkopac:
That's fine. We should remove the cmake installation step since all runners come with cmake.

robertosfield:
The ci.yml is pretty cryptic, it's not clear to me what each bit really does. If we can remove stuff that isn't required then so much the better.

PR #1203, opened 2024-05-29
Ensure minimum maxSets is at least one

theodoregoetz:
Following the discussions in #960 and #961 this sets the minimum maxSets to 1 so that when switching out graphics pipelines that use different descriptor sets, the warning about Context::reserve() is avoided. See also 

See issues:

* #960 
* #961 

And in vsgExamples:

* https://github.com/vsg-dev/vsgExamples/issues/312
* https://github.com/vsg-dev/vsgExamples/pull/311

This PR is a possible fix for the warning, but it's unclear to this author if this is the correct and robust solution. I expect a discussion on the merits of this w.r.t. other possible changes to follow here or in the discussion board.

### DIFF
--- a/src/vsg/vk/Context.cpp
+++ b/src/vsg/vk/Context.cpp
@@ -10,6 +10,9 @@ THE SOFTWARE IS PROVIDED "AS IS", WITHOUT WARRANTY OF ANY KIND, EXPRESS OR IMPLI
 
 </editor-fold> */
 
+#include <algorithm>
+#include <cassert>
+
 #include <vsg/commands/Commands.h>
 #include <vsg/commands/CopyAndReleaseBuffer.h>
 #include <vsg/commands/CopyAndReleaseImage.h>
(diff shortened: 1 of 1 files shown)


robertosfield:
I am busy with rewriting vsg::Allocator right now, and have client work backed up so can't look into the topic in detail right away.  This topic that I will look at before 1.2, it may be that I need to rewrite the DescriptorPool management.

robertosfield:
I am now testing/reviewing the bug, I have test case that recreates it reliably so will use that for testing.

As part of the testing/reviewing process I have merged this PR as a branch:

https://github.com/vsg-dev/VulkanSceneGraph/tree/theodoregoetz-context-descriptor-pool-maxsets

I haven't yet looked into the bug and the associated code enough to know what the correct fix will be, hopefully I'll have this resolved in the next day or so and will then make another developer release for it.

robertosfield:
Running my test against this PR it does at least quieten the console warning, at least that's confirmed,  can't say whether it's just hiding an issue or resolve it yet though.

robertosfield:
I'm now quite advanced with a refactor of how vsg::DescriptorPool are managed with the introduction of new vsg::DescriptorPools class.  The implementation should already address the issues tackled by this PR so should make this PR redudent.  Could you please test the [DescriptorPool ](https://github.com/vsg-dev/VulkanSceneGraph/tree/DescriptorPool)branch of the VSG to check that things are working correctly for you application? Thanks.

rolandhill:
Hi Robert,

I tested the DescriptorPool branch on both Kubuntu and Windows and both
worked fine. Only the Kubuntu version was running with the debug layer and
there were no messages. Looks good to me!

Regards,
Roland

PR #1248, opened 2024-07-26
Fixed various issues with handling 'extended' keys on the Windows pla…

### LINKED ISSUE #1198: Probably incorrect handling of left/right ALT key on Windows
**Describe the bug**
I've noticed that on Windows I don't receive vsg::KEY_Alt_L and vsg::KEY_Alt_R for vsg::KeyPressEvent and vsg::KeyReleaseEvent. I've checked Win32_Window.cpp and there is VirtualKeyToKeySymbolMap _vk2vsg which is used for mapping. But for VK_LMENU and VK_RMENU it is mapped to

```
{VK_LMENU                             ,              KEY_Menu},
{VK_RMENU                             ,              KEY_Menu},
```
Though according to https://learn.microsoft.com/en-us/previous-versions/windows/embedded/ms927178(v=msdn.10)?redirectedfrom=MSDN
```
VK_LMENU | 0xA4 | Left ALT
VK_RMENU | 0xA5 | Right ALT
```

That is why current behavior doesn't look correct. It's not clear why KEY_Menu was used as mapping to vsg::KEY_Alt_L and vsg::KEY_Alt_R seems to be more appropriate.

**To Reproduce**
Check vsg::KeyPressEvent or vsg::KeyReleaseEvent when left or right ALT is pressed on Windows.

**Expected behavior**
Receive  vsg::KEY_Alt_L and vsg::KEY_Alt_R when left or right ALT is pressed on Windows.

 - Windows 11

appcodegen:
## Description

Fixed various issues with handling 'extended' keys on the Windows platform
- the code now distinguishes between 'left' and 'right' ALT and CTRL modifier keys
- the code now handles the 'Applications' key (VK_MENU) correctly
- added the various keys on the numpad (VK_NUMPADx, VK_MULTIPLY, VK_ADD etc)
- 'scroll lock' is now handled
- 'print screen' is now correctly handled (VK_SNAPSHOT instead of VK_PRINT)

Fixes #1198 

## Type of change

## How Has This Been Tested?

I ran the vsgInput example

**Test Configuration**:
* SDK: Windows 10, Visual Studio 2022

## Checklist:


src/vsg/platform/win32/Win32_Window.cpp
```diff
@@ -122,7 +122,7 @@ KeyboardMap::KeyboardMap()
             {VK_RETURN                            ,              KEY_Return},
             {VK_SHIFT                             ,              KEY_Undefined},
             {VK_CONTROL                           ,              KEY_Undefined},
-            {VK_MENU                              ,              KEY_Undefined},
+            {VK_MENU                              ,              KEY_Alt_L},
```
robertosfield, line 125: Why wouldn't this many to KEY_Menu?

src/vsg/platform/win32/Win32_Window.cpp
```diff
@@ -149,32 +149,32 @@ KeyboardMap::KeyboardMap()
             {VK_RIGHT                             ,              KEY_Right    },
             {VK_DOWN                              ,              KEY_Down     },
             {VK_SELECT                            ,              KEY_Select},
-            {VK_PRINT                             ,              KEY_Print},
+            {VK_SNAPSHOT                          ,              KEY_Print},
```
robertosfield, line 152: If VK_PRINT shouldn't map to KEY_Print would should VK_SNAPSHOT?

What does VK_PRINT and KEY_Print meaning and are you assuming?

src/vsg/platform/win32/Win32_Window.cpp
```diff
@@ -149,32 +149,32 @@ KeyboardMap::KeyboardMap()
...
             {VK_EXECUTE                           ,              KEY_Execute},
             {VK_SNAPSHOT                          ,              KEY_Undefined},
             {VK_INSERT                            ,              KEY_Insert},
             {VK_DELETE                            ,              KEY_Delete},
             {VK_HELP                              ,              KEY_Help},
             {VK_LWIN                              ,              KEY_Super_L},
             {VK_RWIN                              ,              KEY_Super_R},
-            {VK_APPS                              ,              KEY_Undefined},
+            {VK_APPS                              ,              KEY_Menu},
```
robertosfield, line 160: Um... VK_Menu isn't a menu but VK_APPS is?

robertosfield:
While it's great to see some obvious mapping that should be included originally, some of the changes look very odd and not suitable for merging without further discussion and testing under Windows.  FYI, I didn't write the Win32 support in the VSG as use Unices predominantly and Windows only occasionally recently, so for Win32 changes I rely upon the community, and need to follow up with official docs to settle any questions if they remain.

src/vsg/platform/win32/Win32_Window.cpp
appcodegen: [VK_MENU on Windows is the (left) 'Alt' key](https://learn.microsoft.com/en-us/windows/win32/inputdev/virtual-key-codes). KEY_Menu is (according to [KeyEvent.h](https://github.com/vsg-dev/VulkanSceneGraph/blob/2d238246a09c3af1d4f268e02bc232442c4c0156/include/vsg/ui/KeyEvent.h#L161)) supposed to be "VK_APPS, the context-menu key"

src/vsg/platform/win32/Win32_Window.cpp
appcodegen: Please see [this comment above](https://github.com/vsg-dev/VulkanSceneGraph/pull/1248#discussion_r1694315657)

src/vsg/platform/win32/Win32_Window.cpp
appcodegen: OK that's a good/valid question..
- I assumed KEY_Print would be the 'Print screen' hotkey/functionality.
- while the MSDN lists the VK_PRINT constant all it says is that it is the 'PRINT key' (I searched around a bit but I could only find [this stackoverflow discussion](https://stackoverflow.com/questions/5815471/whats-the-purpose-of-vk-print) where it's stated that VK_PRINT is basically a nowadays unused remnant from 83/84 key IBM XT/AT machines). However looking at they keyboard in front of me, pressing the 'PrtSc' button I could see that the VK_SNAPSHOT event was created by Windows :)

robertosfield: I just tried under Linux and pressing "Press Screen" is mapping to VK_PRINT, so if Windows generates VK_SNAPSHOT for this key then that would be correct mapping.

robertosfield:
To help with testing I've added a map<vsg::KeySymbol, std::string> to the vsginput example that provides the enum entry name when we press/release keys so we can more easily see how the VSG is treating keys: 

https://github.com/vsg-dev/vsgExamples/commit/278c6a3a2c18121a07174247fbe605df5faa14eb

My aim with the vsg::KeyRelease/keyPress events is the vsg::KeySymbol reported will be same across platforms.

I have also merged this PR as a branch for myself and others to test/make amendments before anything final gets merged with VSG master.

https://github.com/vsg-dev/VulkanSceneGraph/tree/appcodegen-WindowsKeyHandling

If possible it would be helpful for users to try out multiple platforms to see whether we are consistently getting the mapping they would expect and where it differs due to bugs we can directly fix them, but if it looks like an inconsistency between platforms then we'll need to figure out what to standardize on.

robertosfield:
I have booted my Windows machine checked out this branch and tested against the updated vsginput and can confirm that Linux and Windows now behave much more closely thanks to this PR :-)

I have merged this PR but note that there a still a few missing mappings:  

1. Pressing Pause/Break doesn't result in any key being reported by vsginput
2. Pressing NumLock doesn't result in any key being reported by vsginput
3. Pressing Fn doesn't result in any key being reported by vsginput
4. Pressing F10 passing vsginput, no idea why this might be. 

I have other work to get on with today so I'll leave these as outstanding.

rolandhill:
Hi Robert,

I was looking at differences in how Linux and Windows treat the numpad keys
last week. I found that pressing numpad numbers in Linux does not produce
any output in ImGui::InputText controls (you can test in the ImGui demo
window).

I made the table below.
The first section shows the contents of the 'base', 'modified' & 'modifier'
members when you press the NORMAL 3 key. Both Linux and Windows produce the
same result.
The second section shows the same thing when you press the NUMPAD 3 key.
Windows appears to be converting the numpad number into a normal number,
except with a different base. Linux produces both KP category base and
modified members.

I'm not sure which is correct or what the intention is. I do know that
Numpad keys don't work in ImGui on Linux (on my machine).

*** Keyboard code differences
Normal 3 (Both the same)
Base,  Modified,  Modifier
Windows: KEY_3,   KEY_3,   MODKEY_NumLock
Linux: KEY_3,  KEY_3,  MODKEY_NumLock

Numpad 3
Base,  Modified,  Modifier
Windows: KEY_NEXT,   KEY_3,   MODKEY_NumLock
Linux: KEY_KP_NEXT,   KEY_KP_3,   MODKEY_NumLock


Regards,
Roland

PR #1249, opened 2024-07-26
Added missing header

appcodegen:
## Description

I regenerated my build config and this change popped up.. automatically done by CMake I think?


robertosfield:
Thanks, the VSG's Cmake build systems generates a _build_all_h_ target that can be run to put all the heads into the all.h. I must have missing running this recently when adding DescriptorPools.h.  I have now run this and checked it in so this PR should be moot.

PR #1250, opened 2024-07-26
Trivial typo

appcodegen:
## Description

Trying to (beginning to) understand cmake I saw this.. it's really trivial


robertosfield:
Thanks.  I'm afraid I have mild dyslexia so often don't spot typos.  Thankfully the compiler is able to catch some of them, others I or eagle eyed others have to come along and correct these silly typos :-)

Issue #1254, opened 2024-07-30
vsgWin32::Win32_Window continues using HWND after it's been destroyed


AnyOldName3:
`vsgWin32::Win32_Window` attempts to destroy its window handle, `_window`, in its destructor by calling `DestroyWindow(_window)`. However, because `vsgWin32::Win32WindowProc` (the window message handling `CALLBACK`) calls `DefWindowProc` (the default window message handler), when a `WM_CLOSE` message is received, the window's closed and destroyed before the destructor runs.

This is especially bad because `HWND`s are reused for new windows, so the destructor could potentially destroy another process' unrelated window. Thankfully, it's unlikely as Windows actively defends against this kind of bug, using some of a HWND's bits for the window table index and some for a generation number so the same index won't turn into the same HWND when it's reused until those bits have been exhausted.

I used this patch to identify the unwanted window murderer, based on this OldNewThing post: https://devblogs.microsoft.com/oldnewthing/20111026-00/?p=9263
[who-killed-my-window.patch](https://github.com/user-attachments/files/16431965/who-killed-my-window.patch)

robertosfield:
Is the WM_CLOSE message a signal that the window is closed or is it a request to close the window?

The Win32_Window::handleWin32Messages(..) method currently treats the WM_CLOSE as a request to close rather than attempting to handle anything more.  

Is the ::DefWindowProc(hwnd, msg, wParam, lParam) call in Win32WindowProc responding to the WM_CLOSE and doing the window destruction?

Can we prevent the window from being destroyed by anyone else other than the destructor.  If ::DefWindowProc() is the one doing the window destruction then can we just not pass on the message to ::DefWindowProc() when Win32_Window::handleWin32Messages(..)  handles the event?

AnyOldName3:
> Is the WM_CLOSE message a signal that the window is closed or is it a request to close the window?

It's a request to close, fired by things like pressing the X in the corner of a window.

> Is the ::DefWindowProc(hwnd, msg, wParam, lParam) call in Win32WindowProc responding to the WM_CLOSE and doing the window destruction?

Yep. The default handling for the close message is to close and destroy the window.

> Can we prevent the window from being destroyed by anyone else other than the destructor. If ::DefWindowProc() is the one doing the window destruction then can we just not pass on the message to ::DefWindowProc() when Win32_Window::handleWin32Messages(..) handles the event?

[...]

Issue #1133, opened 2024-03-21
vsg::convert_utf does not work properly on platforms where wchar_t is not equivalent to char32_t


AnyOldName3:
On some platforms, `wchar_t` is a 32-bit type, and the same width as `char32_t`, intended to hold UCS4/UTF-32 code points as fixed-width strings. On others, in particular, ones that attempted to support Unicode in the 90s, when UTF-8 and UTF-16 hadn't been invented and The Unicode Consortium thought that sixteen bits would be enough to hold any character from any writing system humans had ever used, `wchar_t` is a 16-bit type and the same width as `char16_t`, intended to hold UCS2 fixed-width strings or UTF-16 variable-width strings.

`src/vsg/io/convert_utf.cpp` works under the assumption that `wchar_t` can hold an entire Unicode code point on its own, which isn't guaranteed. This can be easily demonstrated by attempting to convert strings containing emoji between `std::string` and `std::wstring` in either direction on Windows, as most emoji occupy code points above 65536, and Windows is one of the platforms where `wchar_t` is sixteen bits. When converting wide strings to narrow, the unpaired surrogates are converted to three bytes each, giving six nonsensical code units per code point, instead of glued to their partner and converted to a combined four correct code units. When converting narrow strings to wide, the four code units are correctly converted to the right code unit held in a `uint32_t`, then `static_cast`ed into `wchar_t`, which truncates the most significant sixteen bits, which works for the first 65536 code points (which is most non-emoji text, hence why it's not been noticed), and then wraps around.

I noticed this because I was poking around, and have seen this bug lots of times in different projects, and not because I'm affected by it, so there's no pressing need to fix this immediately, but it'll end up affecting someone eventually.

LeeRiva:
I wonder if one could simply check for the size of wchar_t and then assume UTF-16? Or check whether it's a Windows build (I think that information is easily available?) and then assume UTF-16?

PR #1255, opened 2024-07-30
Check and report errors for Win32 functions that may fail

AnyOldName3:
Skips a couple in the `Win32_Window` destructor. One will throw because of #1254, so this should be temporary. The other's because there's a relatively benign failure as the API won't let you unregister a window class while there are still windows using it, which is what we want.

### DIFF
--- a/src/vsg/platform/win32/Win32_Window.cpp
+++ b/src/vsg/platform/win32/Win32_Window.cpp
@@ -19,6 +19,27 @@ THE SOFTWARE IS PROVIDED "AS IS", WITHOUT WARRANTY OF ANY KIND, EXPRESS OR IMPLI
 using namespace vsg;
 using namespace vsgWin32;
 
+namespace
+{
+    void reportWin32Error()
+    {
+        DWORD errorCode = GetLastError();
+        if (errorCode == NOERROR)
+            return;
+        // Hopefully the error will be representable with the current eight-bit code page as Exception doesn't support wide strings
+        LPSTR buffer;
+        // Ignore the dodgy cast from pointer-to-pointer to pointer
+        // the argument should really be a union of those types as it's interpreted differently depending on the flags passed
+        std::size_t length = FormatMessageA(FORMAT_MESSAGE_ALLOCATE_BUFFER | FORMAT_MESSAGE_FROM_SYSTEM | FORMAT_MESSAGE_IGNORE_INSERTS,
+            nullptr, errorCode, MAKELANGID(LANG_NEUTRAL, SUBLANG_DEFAULT), (LPSTR)&buffer, 0, nullptr);
+
+        std::string message(buffer, length);
+        LocalFree(buffer);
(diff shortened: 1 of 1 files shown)


robertosfield:
Thanks for the changes.  I feel there is some more evolution of the code that is needed before merging with VSG master so I've merged as a branch: https://github.com/vsg-dev/VulkanSceneGraph/tree/AnyOldName3-check-win32-errors

One of the changes I'm thinking about is changing the Win32_Window::resize() method so it only updates the _extend2D and calls buildSwapchain() when GetClientRect(..) return true.

Another is moving the GetClientRect(..) usage in to within the Win32_Window::handleWin32Messages(..) into the blocks that actually need it rather than for all event messages.

Your comment about vsg::Exception not supporting wide strings.  One thing I've been wondering about is looking at changing vsg::Exception so it subclasses from one of the std exception classes and just adding to them.  I don't know if that would aid with the whole widestring situation under Windows as I haven't looked into this topic in detail.

AnyOldName3:
Deriving from `std::exception` would generally be a good idea (lots of applications have a global `std::exception` catch block that displays a fatal error message etc. before exiting), but it wouldn't help the Windows Unicode situation as `std::exception::what` returns a C string. There are still LTS versions of Windows under support until 2029 that don't support UTF-8 as the narrow string encoding, so the *put the application into UTF-8 mode if you care about Unicode and can't be bothered with wide strings* solution isn't viable for enterprise-friendly software yet. There *is* the option to internally use UTF-8 everywhere and convert to UTF-16 every time a function talks to the Win32 API, but in modern C++, that requires some dodgy-looking `reinterpret_cast`s as `char` and `char8_t` are distinct types and there are a few places where STL functions assume `char` uses the OS' eight-bit encoding and can't be convinced otherwise.

It's particularly annoying as you can't even blame Microsoft for the mess. They followed the Unicode Consortium's advice before UTF-8 had been invented and before they'd noticed there were more than 65536 unique letters in all human languages, so were all-in on 16-bit wide strings before Unix systems supported Unicode at all.

AnyOldName3:
Temporarily ignoring the fact that that conversion is broken (see #1133), the conversion itself isn't the tricky part, it's ensuring it happens in the right places and gets communicated to the people who need to know. It does look like lots of other bits of the VSG are working under the assumption that `char`-based strings are UTF-8, in which case it would make sense to always do UTF-16 conversions and always call the `W`-suffixed versions of Win32 functions instead of calling the `A`-suffixed ones or letting the `UNICODE` define pick. Otherwise, the Win32 API will most likely expect and return strings in a fixed-width eight-bit encoding like CP1252, which is only compatible with a small subset of UTF-8.

robertosfield:
I'm closing this PR as I believe all the key functionality is merged in the https://github.com/vsg-dev/VulkanSceneGraph/tree/Win32_Window_Refinements branch.

PR #1264, opened 2024-08-21
Move numReadThreads to be a class member so that it can be set extern…

rolandhill:
Make numReadThreads a member of DatabasePager so that it can be set externally if desired. It used to be hardwired in the start() method. The default value is kept the same so there should be no effect on existing code.

### DIFF
--- a/include/vsg/io/DatabasePager.h
+++ b/include/vsg/io/DatabasePager.h
@@ -99,6 +99,8 @@ namespace vsg
         std::atomic_uint numActiveRequests{0};
         std::atomic_uint64_t frameCount;
 
+        int numReadThreads{4};
+
         ref_ptr<CulledPagedLODs> culledPagedLODs;
 
         /// for systems with smaller GPU memory limits you may need to reduce the targetMaxNumPagedLODWithHighResSubgraphs to keep memory usage within available limits.
--- a/src/vsg/io/DatabasePager.cpp
+++ b/src/vsg/io/DatabasePager.cpp
@@ -133,8 +133,6 @@ void DatabasePager::assignInstrumentation(ref_ptr<Instrumentation> in_instrument
 
 void DatabasePager::start()
 {
-    int numReadThreads = 4;
-
     //
     // set up read thread(s)
     //


robertosfield:
The problem with setting the value this way is once the database pager has started the number of read threads will be fixed so subsequent changes to the numReadThreads won't have an affect.

In you app what is your usage pattern?  Perhaps we can come up with a way that allows you the controlability you need without having a number member that only has an effect in one particular usage patter.

rolandhill:
At the moment I only intend to change the number of threads once as the
application starts. I hadn't really thought about adjusting it on-the-fly.
I do a reasonable amount of processing when a database tile/octree cell is
requested and the queue can get up over 900 requests pending, which takes
time to clear. Experimenting to use 8 read threads looks to help, at least
on my machine. Ideally I could set it according to the user's CPU resources.

More generally, my users need to be able to change their 'project' at any
time so I've derived my own DBPager from your DatabasePager and included my
own functions for start(), stop(), flush(), updateSceneGraph(). When they
change project I stop the DatabasePager, flush it and then restart. This
also means that I can create my DBPager manually (without it starting), set
the number of threads, then start it. There is nothing in my scene graph
when I call the first compile(), so it doesn't get set up automatically.
All content is added dynamically after the frame loop is running.

If you can think of a better way then I'm happy to go with it.

rolandhill:
It's just after - see below. I can set the number of read threads in the
DbPager constructor with the changes in this PR. Nodes attached at the time
compile() is called are minimal.

_viewer->compile();

// Install the custom database pager
_dbPager = DbPager::create();
_dbPager->compileManager = vsg::CompileManager::create(*_viewer,
resourceHints);
for(auto& task : _viewer->recordAndSubmitTasks)
{
task->databasePager = _dbPager;
}
_dbPager->start();

robertosfield:
Perhaps we could pass in the settings for the DatabaserPager via the viewer::compile(resourceHints);  So add the number of threads to use into the resourceHints.  This way you wouldn't need any database pager setup code or need to create a new CompileManager or DatabasePager.

rolandhill:
I'm happy to try that. As long as the number of threads isn't hard coded as
a local variable in the start() method, then it would be fully flexible for
anyone who wants to customise it.

robertosfield:
My work on utilizing TransferTask within the CompileTraversal I've ended working on DatabasrePager to fix the occasional high DatabaserPager::updateSceneGraph() times that was causing big frame drops, this was resolved by introducing a dedicated DeleteQueue that now has a delete thread in DatabasePager.  This work's now done, and as I was already working on DatabasrPager I've implemented support for specifying the number of read threads to create with commit: c2a2559f94e3a21ec3f7d7c7e197c98cd9cd7efc

This can be found in the Compile_Using_TransferData branch:  https://github.com/vsg-dev/VulkanSceneGraph/tree/Compile_Using_TransferData

I will do some more testing, if that all goes well I'll merge with VSG master tomorrow.

With these changes you can now do databasePager->start(numReadThreads)l, and as the vsg::Viewer::compile() calls databasePager->start(..) for users I've added a ResourceHints::numDatabasePagerReadThreads member that defaults to 4, you can pass this in the form viewer->compile(hints);

I have checked in some changes to vsgviewer in vsgExampes to enable users to read and output a ResourceHints file using --rh resource.vsgt or --orh resource.vsgt command line options: 

https://github.com/vsg-dev/vsgExamples/commit/b9f3c903d2cfdcc31a9275f53ab3529663f6a5c0

rolandhill:
Hi Robert,

I tried the Compile_Using_TransferData branch on Ubuntu 24.04 and setting
the threads using start(8). No problems - all went as expected.

Cheers,
Roland

robertosfield:
Thanks for the testing.  I'll close this PR as it looks like the feature required is working fine.

PR #1302, opened 2024-10-14
Gamma correctness phase one

AnyOldName3:
As discussed here https://github.com/vsg-dev/VulkanSceneGraph/discussions/1291, there are quite a few gamma-correctness issues and annoying-to-work-around obstacles in the VSG. Between this, and related PRs https://github.com/vsg-dev/vsgXchange/pull/204 and https://github.com/vsg-dev/vsgExamples/pull/323, things should be more manageable.

This PR:
* Changes the default surface image format from a `UNORM` one to an `SRGB` one, enabling automatic linear-to-sRGB conversions when data is written to the framebuffer. We're already telling the driver that we're writing sRGB data to the surface as we're giving it an sRGB colour space. This should avoid things like alpha blending issues caused by doing the sRGB conversion first. The vsgExamples PR includes an update to the PBR shader to remove the now-redundant conversion it was doing at the end.
* Fixes the conversion functions in `color.h` as they did the opposite of what they said. The call site I found seemed to be calling the one with the right implementation but wrong name, so I switched it over.
* Adds extra fields to `vsg::Options` to control when vsgXchange converts colour spaces in loaded data.

Things with a potential to break existing applications can be opted out of - you can still explicitly ask for a `UNORM` surface format and set the source and destination colour spaces via `vsg::Options`.

### DIFF
--- a/include/vsg/io/Options.h
+++ b/include/vsg/io/Options.h
@@ -87,6 +87,22 @@ namespace vsg
         /// Coordinate convention to assume for specified lower case file formats extensions
         std::map<Path, CoordinateConvention> formatCoordinateConventions;
 
+        enum ColorSpace
+        {
+            sRGB,
+            linearRGB
+        };
+
+        /// Color space to use for vertex colors for scene graph
+        ColorSpace sceneVertexColorColorSpace = ColorSpace::linearRGB;
+        /// Color space to use for materials for scene graph
+        ColorSpace sceneMaterialColorSpace = ColorSpace::linearRGB;
+
+        /// Color space to assume for vertex colors for specified lower case file formats extensions
+        std::map<Path, ColorSpace> formatVertexColorColorSpaces;
+        /// Color space to assume for materials for specified lower case file formats extensions
+        std::map<Path, ColorSpace> formatMaterialColorSpaces;
+
         /// User defined ShaderSet map, loaders should check the available ShaderSet using the name of the type of ShaderSet.
         /// Standard names are :
         ///     "pbr" will substitute for vsg::createPhysicsBasedRenderingShaderSet()
--- a/include/vsg/maths/color.h
+++ b/include/vsg/maths/color.h
@@ -99,28 +99,28 @@ namespace vsg
     template<typename T>
     constexpr t_vec4<T> linear_to_sRGB(const t_vec4<T>& src)
     {
-        const T exponent = static_cast<T>(2.2);
+        const T exponent = static_cast<T>(1.0 / 2.2);
         return t_vec4<T>(std::pow(src.r, exponent), std::pow(src.g, exponent), std::pow(src.b, exponent), src.a);
     }
 
     template<typename T>
     constexpr t_vec4<T> linear_to_sRGB(T r, T g, T b, T a)
     {
-        const T exponent = static_cast<T>(2.2);
+        const T exponent = static_cast<T>(1.0 / 2.2);
         return t_vec4<T>(std::pow(r, exponent), std::pow(g, exponent), std::pow(b, exponent), a);
     }
 
     template<typename T>
     constexpr t_vec4<T> sRGB_to_linear(const t_vec4<T>& src)
     {
-        const T exponent = static_cast<T>(1.0 / 2.2);
+        const T exponent = static_cast<T>(2.2);
         return t_vec4<T>(std::pow(src.r, exponent), std::pow(src.g, exponent), std::pow(src.b, exponent), src.a);
     }
 
     template<typename T>
     constexpr t_vec4<T> sRGB_to_linear(T r, T g, T b, T a)
     {
-        const T exponent = static_cast<T>(1.0 / 2.2);
+        const T exponent = static_cast<T>(2.2);
         return t_vec4<T>(std::pow(r, exponent), std::pow(g, exponent), std::pow(b, exponent), a);
     }
 
--- a/src/vsg/app/Window.cpp
+++ b/src/vsg/app/Window.cpp
@@ -41,7 +41,7 @@ Window::Window(ref_ptr<WindowTraits> traits) :
 {
     if (_traits && (_traits->swapchainPreferences.surfaceFormat.format == VK_FORMAT_B8G8R8A8_SRGB || _traits->swapchainPreferences.surfaceFormat.format == VK_FORMAT_B8G8R8_SRGB))
     {
-        _clearColor = linear_to_sRGB(_clearColor);
+        _clearColor = sRGB_to_linear(_clearColor);
         info("Selected sRGB window ", _clearColor);
     }
 }
--- a/include/vsg/vk/Swapchain.h
+++ b/include/vsg/vk/Swapchain.h
@@ -27,7 +27,7 @@ namespace vsg
     };
 
     extern VSG_DECLSPEC SwapChainSupportDetails querySwapChainSupport(VkPhysicalDevice device, VkSurfaceKHR surface);
-    extern VSG_DECLSPEC VkSurfaceFormatKHR selectSwapSurfaceFormat(const SwapChainSupportDetails& details, VkSurfaceFormatKHR preferredSurfaceFormat = {VK_FORMAT_B8G8R8A8_UNORM, VK_COLOR_SPACE_SRGB_NONLINEAR_KHR});
+    extern VSG_DECLSPEC VkSurfaceFormatKHR selectSwapSurfaceFormat(const SwapChainSupportDetails& details, VkSurfaceFormatKHR preferredSurfaceFormat = {VK_FORMAT_B8G8R8A8_SRGB, VK_COLOR_SPACE_SRGB_NONLINEAR_KHR});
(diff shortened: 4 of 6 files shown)


src/vsg/app/Window.cpp
```diff
@@ -41,7 +41,7 @@ Window::Window(ref_ptr<WindowTraits> traits) :
 {
     if (_traits && (_traits->swapchainPreferences.surfaceFormat.format == VK_FORMAT_B8G8R8A8_SRGB || _traits->swapchainPreferences.surfaceFormat.format == VK_FORMAT_B8G8R8_SRGB))
     {
-        _clearColor = linear_to_sRGB(_clearColor);
+        _clearColor = sRGB_to_linear(_clearColor);
```
robertosfield, line 44: If the swapchain is SRGB then why would one need to convert from sRGB_to_linear?

The default _clearColor has been linear up to this point.

include/vsg/maths/color.h
```diff
@@ -99,28 +99,28 @@ namespace vsg
     template<typename T>
     constexpr t_vec4<T> linear_to_sRGB(const t_vec4<T>& src)
     {
-        const T exponent = static_cast<T>(2.2);
+        const T exponent = static_cast<T>(1.0 / 2.2);
```
robertosfield, line 102: What testing did you use to determine that the conversion was wrong?  You've flipped vsg::Window constructor and this local code so that Window.cpp code is now doing exactly the same as before but with flipped meaning.

robertosfield:
I'm happy to see us migrate to sRGB frame buffers, but some of these changes have left me scratching my head - I have added comments to the code changes that will hopefully highlight why I'm confused by the changes.

I think it would be useful to have a test example in vsgExamples to test the conversions between linear and sRGB.

include/vsg/maths/color.h
AnyOldName3: I was feeding `sRGB_to_linear` sRGB colours with the vsgXchange PR, and getting nonsensical results. When switched, I got correct results. It's pretty easy to find sources saying these functions should work the way this PR makes them work. For a start, the conversion functions in the standard VSG PBR shader work this way.

robertosfield: I noticed that shaders used the ratios in your PR.

The change you've made to Window.cpp still seem odd.  If the frame buffer is sRGB when would you convert to sRGB to linear?

src/vsg/app/Window.cpp
AnyOldName3: The image format was previously `_UNORM` with the colour space set to sRGB, so there was no automatic from-sRGB conversion happening during the clear before this PR, and the value of `_clearColor` was what would end up on an sRGB screen. As `linear_to_sRGB` was really doing a `sRGB_to_linear` conversion before this PR, that meant you'd end up with a nonsense colour, but it looked fine because it's not like the precise maths to derive the clear colour particularly matters. When the framebuffer image format was changed to `_SRGB`, it made Vulkan start doing an automatic linear-to-sRGB conversion during the clear, which combined with the fix to make the colour space conversion functions match their names, made the clear colour really dark due to a double linear-to-sRGB conversion. As people typically type out colours in perceptually uniform sRGB space (e.g. they'd chosen it with a colour picker in an image editor, which nearly always use sRGB by default), and Vulkan was now expecting a linear colour, it seemed most reasonable to do a sRGB-to-linear conversion here.

include/vsg/maths/color.h
AnyOldName3: I've left a comment on that thread about it. When the framebuffer has an sRGB format, it means Vulkan expects to be fed linear colour and will then do a to-sRGB conversion any time anything is written to the framebuffer.

robertosfield: Um.... OK, so that kinda explains the perverse conversion in Window.cpp on the _clearColor, but it's not a sRGB_to_linear conversion, it's really a preconversion of a linear colour to something that remains the approrpiate colour when the framebuffer operation converts the linear colour to sRGB.  Mathematically I can see that it's the same as sRGB_to_linear but it's not actually doing that conversion.

src/vsg/app/Window.cpp
robertosfield: I am now inclined to merge this PR as a branch and start testing.  I think the changes to Window.cpp really need explanation as what is going on why it looks like something perverse is being done.

AnyOldName3: The original function usage didn't make sense. The original `_clearColor` value was being interpreted as sRGB due to the `_UNORM` image format paired with `_SRGB_NONLINEAR` colour space, and yet the `{0.2f, 0.2f, 0.4f, 1.0f}` was going through an operation claiming to be an sRGB-to-linear conversion. Obviously, you don't convert something from the colour space it's about to be interpreted as to a different one.

If `{0.2f, 0.2f, 0.4f, 1.0f}` is linear, then the correct original behaviour would be a linear-to-sRGB conversion (which is what the function call said it was doing, but it was instead doing the opposite) and the new correct behaviour would be to do no conversion at all.

If `{0.2f, 0.2f, 0.4f, 1.0f}` is sRGB, then the correct original behaviour would be to do no conversion at all, and the new correct behaviour would be to do an sRGB-to-linear conversion.

If the displayed colour on the screen was correct, then that would suggest that `{0.2f, 0.2f, 0.4f, 1.0f}` was in a made-up colour space with a 4.84 gamma value (as it became sRGB when undergoing an sRGB-to-linear conversion), and the sensible thing to do would be to write it out in a real colour space (whether sRGB or linear) instead.

As it seemed unlikely that you'd knowingly have used a mad colour space, called `sRGB_to_linear` on a colour you thought was already linear, or ignored a clear colour that was wildly different to the one you thought you'd specified, it seemed most plausible that `{0.2f, 0.2f, 0.4f, 1.0f}` was sRGB, so the code now does what's most appropriate for an sRGB colour.

AnyOldName3: Okay, it seems *something's* slightly off in my explanation. I made a test image:

image

Both with and without this PR, the clear colour is displayed like the middle row on my machine. That makes sense with this PR, but as it was previously doing a sRGB-to-linear conversion (mislabeled as linear-to-sRGB) and then supposedly no further conversions, it doesn't make sense without this PR. I'm digging further to see what's going on.

AnyOldName3: Well that was dumb. I'd somehow missed that the conversion was happening in an `if` block predicated on the swapchain format. I'll take another look and revise some of what I said.

include/vsg/io/Options.h
```diff
@@ -87,6 +87,22 @@ namespace vsg
         /// Coordinate convention to assume for specified lower case file formats extensions
         std::map<Path, CoordinateConvention> formatCoordinateConventions;
 
+        enum ColorSpace
+        {
+            sRGB,
+            linearRGB
+        };
+
+        /// Color space to use for vertex colors for scene graph
```
robertosfield, line 96: Is this essentially the target colour space that users want to create their scene graphs so they match with the framebuffer colour spaces settings?

src/vsg/app/Window.cpp
robertosfield: Thanks for trying to come up with coherent explanation.

The clearColor values used are simply what worked for the OSG and I wanted to keep the background consistent to avoid too much of jump for users going between the OSG and VSG.  The implicit assumption was that the clear colour was in linear RGB colour space, the frame buffer and final display would also be in linear RGB colour space.

From my reading of what you've written is the assumption that the final display would honour the linear RGB colour space of the frame buffer is misplaced, and that monitors would be treating the linear RGB frame buffer as straight input without conversion and would be displaying it in sRGB, but... these monitors are assuming the frame buffers are already in sRGB rather linear RGB, so the whole chain of conflicting assumptions is what we are essence trying to untangle.

robertosfield:
Do we need to start putting the colour space info into the scene graph so when we save and then reload the application can decide whether the vertex and texture data needs converting?  

The VSG doesn't presently record colour space in the scene graph, so existing .vsgb/.vsgt will implicitly be linearRGB.  Does the vsg::Data::Properties struct now need a colour space hint?  Adding to Data::Properties has to be carefully as we don't want to bump over a word boundary and have it's size increase.

robertosfield:
I have merged this PR as branch:

https://github.com/vsg-dev/VulkanSceneGraph/tree/AnyOldName3-gamma-correctness-phase-one

I have also merged the vsgXchange and vsgExamples PR's as branches too.  Then ran make build_ShaderSets to make sure all the built in shaders are updated too.

include/vsg/io/Options.h
AnyOldName3: It's not necessarily to match the framebuffer's image format (it's probably more common to have it as the opposite of the framebuffer's image format), but it *is* to control the colour space that users want their scenegraph to use internally. Users who are doing explicit manual conversions in their shaders or are being intentionally gamma-incorrect (e.g. to faithfully draw legacy content) will probably want sRGB vertex and material colours, whereas most users will want linear vertex and material colours and will have their shaders operate entirely with linear colours, and rely on the automatic conversions on store that come with using an sRGB image format for the framebuffer.

PR #1335, opened 2024-11-26
Add a View resize notification function called from WindowResizeHandler

dov:
- This allows creating views that are redrawn on window resize

# Pull Request Template

## Description

Some `View`'s may need updating when the window is resized. E.g. if doing a pixel exact 2D overlay, we may want to resize the elements to be of a certain size in pixels. I could not find any callback that I could catch in my custom view that notifies of such resizing. This was the reason that I introduced the `resize()` abstract callback.

Another issue with pixel exact overlays, is that we may want to have the projection matrix fixed. Currently, the projection matrix is updated in `WindowResizeHandler::apply(vsg::View&)`.  An overloaded `resize()` method allows resetting the `projectionMatrix` to an expected transparent state.

Fixes # (issue)

## Type of change

## How Has This Been Tested?

- [ ] Test A

I created a test program available here: https://gist.github.com/dov/b4ca06a0b13558b75da05948eb955777

**Test Configuration**:

Tested on Linux .
## Checklist:

### DIFF
--- a/include/vsg/app/View.h
+++ b/include/vsg/app/View.h
@@ -86,6 +86,9 @@ namespace vsg
         /// override states for customization of graphics pipelines for this view
         GraphicsPipelineStates overridePipelineStates;
 
+        // abstract class to notify that the window is resized
+        virtual void resize() {};
+
     protected:
         virtual ~View();
     };
--- a/src/vsg/app/WindowResizeHandler.cpp
+++ b/src/vsg/app/WindowResizeHandler.cpp
@@ -183,4 +183,7 @@ void WindowResizeHandler::apply(vsg::View& view)
     view.traverse(*this);
 
     context->defaultPipelineStates.pop_back();
+
+    // Notify the view about the resize
+    view.resize();
 }


dov:
Just subclassing the `WindowResizeHandle` does not solve the problem of *how* the overrided `WindowResizeHandle` will notify its child views that they were resized. Currently I worked around this by creating a "repository" of resizeble views, and then notifying all of them. My idea of this pull request was to resizable views default, and get around the need for this hack.

I do have the feeling though that I'm missing something, and there must be an easier way to do it. 

Your suggestions are most appreciated.

```c++

// This is like view, but it has an additional virtual method resize() which is
// called the containing window is resized
class ResizableView : public vsg::Inherit<vsg::View, ResizableView>
{
  public:
     explicit ResizableView(vsg::ViewFeatures in_features = vsg::RECORD_ALL)
       : vsg::Inherit<vsg::View, ResizableView>(in_features) {}
     explicit ResizableView(vsg::ref_ptr<vsg::Camera> in_camera, vsg::ref_ptr<vsg::Node> in_scenegraph = {}, vsg::ViewFeatures in_features = vsg::RECORD_ALL)
      : vsg::Inherit<vsg::View, ResizableView>(in_camera, in_scenegraph, in_features) {}
    virtual void resize() {};
};

:

class MyWindowResizeHandler : public vsg::Inherit<vsg::WindowResizeHandler, MyWindowResizeHandler>
{
public:
    MyWindowResizeHandler()
        : vsg::Inherit<vsg::WindowResizeHandler, MyWindowResizeHandler>() {}

    void apply(vsg::View& view) override {
        vsg::Inherit<vsg::WindowResizeHandler, MyWindowResizeHandler>::apply(view);

        if (resizableViews.count(&view) > 0)
            ((ResizableView*)&view)->resize();
    };

    // A "registry" for resizable views
    std::set<vsg::View*> resizableViews;
};
:
class LogoView : public vsg::Inherit<ResizableView, LogoView>
{
}
:
vsg::ref_ptr<MyWindowResizeHandler> resizeHandler = MyWindowResizeHandler::create();
resizeHandler->resizableViews.insert(logoView.get());

// Overridethe renderGraph resize handler
renderGraph->windowResizeHandler = resizeHandler;


```

robertosfield:
You can simply do a dynamic_cast<ResizableView*> in the MyWindowResizeHandler::apply(View& view) or override the accept() method of the ResizeableView::accept(Visitor* visitor) method and do a cast to WindowResizeHandler/MyWindowResizeHandler and do any special work you want there.

robertosfield:
The vsgvisitorcustomtype example might of interest:

https://github.com/vsg-dev/vsgExamples/tree/master/examples/core/vsgvisitorcustomtype

dov:
@robertosfield Thanks! `dynamic_cast` indeed cleans it up, an removes my ugly set of resizable views! I really have to get used to thinking in terms of RTTI!

PR #1345, opened 2024-12-13
Update PolytopeIntersector.cpp

jamierobertson1:
In the PolytopeIntersector constructor with screen coordinates, the eyespace polytope should also be placed in the polytope stack before the worldspace polytope (In the same way as in the LineSegmentIntersector constructor). 

# Pull Request Template

## Description

Change is to to also add the eyespace polytope to the polytope stack before the worldspace polytope

Fixes # 
The omission meant that intersections were not calculated properly for nodes under MatrixTransforms.

## Type of change

Can be tested by modifying the vsgintersection example default box creation to be under a MatrixTransform as follows: 
(wont work without the change)
```
    if (scene->children.empty())
    {
        vsg::GeometryInfo info;
        info.cullNode = true;
        info.dx.set(100.0f, 0.0f, 0.0f);
        info.dy.set(0.0f, 100.0f, 0.0f);
        info.dz.set(0.0f, 0.0f, 100.0f);
        auto box = builder->createBox(info, stateInfo);
        auto transform = vsg::MatrixTransform::create();
        transform->matrix = vsg::translate(vsg::dvec3(1000.0, 1000.0, 1000.0));
        transform->addChild(box);
        scene->addChild(transform);
    }

```

Tested on windows

### DIFF
--- a/src/vsg/utils/PolytopeIntersector.cpp
+++ b/src/vsg/utils/PolytopeIntersector.cpp
@@ -229,6 +229,8 @@ PolytopeIntersector::PolytopeIntersector(const Camera& camera, double xMin, doub
         eyespace.push_back(pl * projectionMatrix);
     }
 
+    _polytopeStack.push_back(eyespace);
+
     vsg::Polytope worldspace;
     for (auto& pl : eyespace)
     {


robertosfield:
I am starting my  code review so first step was to confirm the bug and then use the PR to fix the bug, but I've fallen on the first hurdle in not being able to recreate the bug yet.  I used your suggest code change, with a small tweak to allow me to toggle on/off the use of the MatrixTransform using the --mt command line option.  The changes I made to the vsgintersection are:

~~~ sh
$ git diff
diff --git a/examples/utils/vsgintersection/vsgintersection.cpp b/examples/utils/vsgintersection/vsgintersection.cpp
index 8dd028b1..17fab22a 100644
--- a/examples/utils/vsgintersection/vsgintersection.cpp
+++ b/examples/utils/vsgintersection/vsgintersection.cpp
@@ -296,7 +296,21 @@ int main(int argc, char** argv)
         info.dx.set(100.0f, 0.0f, 0.0f);
         info.dy.set(0.0f, 100.0f, 0.0f);
         info.dz.set(0.0f, 0.0f, 100.0f);
-        scene->addChild(builder->createBox(info, stateInfo));
+
+        if (arguments.read("--mt"))
+        {
+            auto box = builder->createBox(info, stateInfo);
+            auto transform = vsg::MatrixTransform::create();
+            transform->matrix = vsg::translate(vsg::dvec3(1000.0, 1000.0, 1000.0));
+            transform->addChild(box);
+            scene->addChild(transform);
+
+            vsg::info("Used transform");
+        }
+        else
+        {
+            scene->addChild(builder->createBox(info, stateInfo));
+        }
     }
 
     // create the viewer and assign window(s) to it
~~~

I can get intersections reported and shapes inserted at the correct places when I use --mt and when I don't, this is with VSG master.  Thoughts?

robertosfield:
I wasn't testing the PolytopeIntersector when I running vsgintersection --mt, now I have tested it correcly I can reproduce the issue and the PR fixes it so I have gone ahead a merged :-)

Thanks for the fix.

PR #1346, opened 2024-12-16
Set depthAttachment.stencilLoadOp to VK_ATTACHMENT_LOAD_OP_CLEAR

franciszhi:
# Pull Request Template

## Description
Set `depthAttachment.stencilLoadOp` from `VK_ATTACHMENT_LOAD_OP_DONT_CARE` to `VK_ATTACHMENT_LOAD_OP_CLEAR`.
Without this change, stencil buffer is not cleared on new frame.

## Type of change

## Checklist:


robertosfield:
That's the default values, if you have a set up that requires non default setting you explicitly set them.

Could it be that you haven't tried setting the values to what you need in your application when setting up the viewer?  You should just need to set up the RenderPass explicitly rather than falling back to fallback.  Have a look at the vsgsubpass example for inspiration.

franciszhi:
Thank you for the suggestion. `vsg::Window::setRenderPass` is very helpful to apply customized render pass.

PR #1347, opened 2024-12-16
vsg::CoordinateFrame, vsg::CameraSampler & long double support.

robertosfield:


AnyOldName3:
This PR (or the adjustments to the examples that it mandated) seem to have broken support for existing serialised camera animations. The camera path files I used to feed to the vsgShadow example for performance testing don't do anything anymore.

robertosfield:
I chose to not attempt to implement backwards compatibility as it would complicated the code quite a bit, and hoped that that old path .vsgt files wouldn't be widely used.

PR #1372, opened 2025-01-22
fixed conversion of wchar_t to uint32_t

Vic-Min:
The size of wchar_t in Linux and many other non-Windows systems is 4 bytes, and converting it to uint16_t may cause information loss.

### DIFF
--- a/src/vsg/text/GpuLayoutTechnique.cpp
+++ b/src/vsg/text/GpuLayoutTechnique.cpp
@@ -199,9 +199,9 @@ void GpuLayoutTechnique::setup(Text* text, uint32_t minimumAllocation, ref_ptr<c
             allocate(static_cast<uint32_t>(text.value().size()));
 
             auto itr = textArray->begin();
-            for (auto& c : text.value())
+            for (auto c : text.value())
             {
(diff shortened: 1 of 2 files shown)


src/vsg/text/GpuLayoutTechnique.cpp
```diff
@@ -199,9 +199,9 @@ void GpuLayoutTechnique::setup(Text* text, uint32_t minimumAllocation, ref_ptr<c
             allocate(static_cast<uint32_t>(text.value().size()));
 
             auto itr = textArray->begin();
-            for (auto& c : text.value())
+            for (auto c : text.value())
             {
-                assignValue(*(itr++), font.glyphIndexForCharcode(uint16_t(c)), updated);
+                assignValue(*(itr++), font.glyphIndexForCharcode(std::make_unsigned_t<decltype(c)>(c)), updated);
```
robertosfield, line 204: Why not just change the uint16_t to uint32_t?  Far simpler to read.

robertosfield:
I have decided to simply promote all the character usage to uint32_t as this is what the Font class etc. are already built around: ca5a9589fe0fd72a78f36c693c114ccf82a8686c

Issue #1405, opened 2025-02-23
Build with clang-cl/lld-link fails due to missing symbol


Neumann-A:
Error:

```
[237/238] C:\WINDOWS\system32\cmd.exe /C "cd . && E:\vcpkg_cache\downloads\tools\cmake-3.30.1-windows\cmake-3.30.1-windows-i386\bin\cmake.exe -E vs_link_dll --intdir=src\vsg\CMakeFiles\vsg.dir --rc=rc.exe --mt="D:\vcpkg_folders\no_msvc\installed\x64-windows-static\compiler\msvc\WinSDK\Windows Kits\10\bin\10.0.26100.0\x64\mt.exe" --manifests  -- D:\vcpkg_folders\no_msvc\installed\x64-windows-static\compiler\llvm\bin\lld-link.exe  @CMakeFiles\vsg.rsp  /out:lib\vsg-14.dll /implib:lib\vsg.lib /pdb:lib\vsg-14.pdb /dll /version:1.1 /machine:x64 /nologo /DEBUG /INCREMENTAL:NO /OPT:REF /OPT:ICF && cd ."
FAILED: lib/vsg-14.dll lib/vsg.lib 
C:\WINDOWS\system32\cmd.exe /C "cd . && E:\vcpkg_cache\downloads\tools\cmake-3.30.1-windows\cmake-3.30.1-windows-i386\bin\cmake.exe -E vs_link_dll --intdir=src\vsg\CMakeFiles\vsg.dir --rc=rc.exe --mt="D:\vcpkg_folders\no_msvc\installed\x64-windows-static\compiler\msvc\WinSDK\Windows Kits\10\bin\10.0.26100.0\x64\mt.exe" --manifests  -- D:\vcpkg_folders\no_msvc\installed\x64-windows-static\compiler\llvm\bin\lld-link.exe  @CMakeFiles\vsg.rsp  /out:lib\vsg-14.dll /implib:lib\vsg.lib /pdb:lib\vsg-14.pdb /dll /version:1.1 /machine:x64 /nologo /DEBUG /INCREMENTAL:NO /OPT:REF /OPT:ICF && cd ."
[...]

PR #1407, opened 2025-02-25
[Invalid] Try building with ClangCL

Neumann-A:
Try reproducing #1405 in VSG own CI (ClangCL toolset should be installed in the default windows runners if i remember correctly.)

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -36,7 +36,7 @@ jobs:
       shell: bash
       run: |
          if [ "$RUNNER_OS" == "Windows" ]; then
-              cmake . -DBUILD_SHARED_LIBS=${{matrix.build-shared}} -A x64
+              cmake . -DBUILD_SHARED_LIBS=${{matrix.build-shared}} -A x64 -T ClangCL
               cmake --build .
          elif [ "$RUNNER_OS" == "Linux" ]; then
               sudo apt-get -qq update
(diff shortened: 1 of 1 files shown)


.github/workflows/ci.yml
```diff
@@ -36,7 +36,7 @@ jobs:
       shell: bash
       run: |
          if [ "$RUNNER_OS" == "Windows" ]; then
-              cmake . -DBUILD_SHARED_LIBS=${{matrix.build-shared}} -A x64
+              cmake . -DBUILD_SHARED_LIBS=${{matrix.build-shared}} -A x64 -T ClangCL
               cmake --build .
```
Neumann-A, line 40:
```suggestion
              cmake --build . --config Release
```

robertosfield:
Thanks for the CL change.  

I have started looking at the log. Lots of the warnings about incompatibility are really brain dead.  The VSG build system explicitly specifies C++17 so not being compatible with C++98 and C++14 is exactly what is expected, so the compiler warnings are just wrong.  I have seen anyone else report these warnings so perhaps just something about the clang version or default warnings that area enabled.

I'm away on a trip and the internet connections here is bad so it's taking forever to browse through the logs so haven't been able to browse through to the actual compile error.  I'll leave this Draft PR up till me return to office next week.

What version of clang are you using?  I've previously built with clang under Linux (that's my main dev platform) without issues.

Neumann-A:
The error is posted here: https://github.com/vsg-dev/VulkanSceneGraph/issues/1405#issuecomment-2682753292 it is the same error as I have observed. Your VS version/CI uses 18.1.x, I use 19.1.x

PR #1412, opened 2025-02-28
Reorder API dump and validation layers

AnyOldName3:
The validation layers eat and alter some Vulkan calls, so a combined log with both layers will have mismatched information.

In particular, vkSetDebugUtilsObjectNameEXT calls do not make it through the validation layers, so the API dump layer won't be able to see any object names or be able dump the calls that set them, and the validation layer internally remaps object IDs, so the IDs in any validation errors won't match those in the dumped API calls.

Ensuring the calls go through the dump layer before the validation layers ensures all calls are seen and can be dumped, and that the public object IDs that the validation layers expose to the application are the same as end up in the API dump.

I'm currently investigating some problems with an app that go away when it's run through RenderDoc, so I need to rely on the API dump layer to see what's going on, and this change gets rid of a lot of manual steps and guesswork.

### DIFF
--- a/src/vsg/app/WindowTraits.cpp
+++ b/src/vsg/app/WindowTraits.cpp
@@ -110,8 +110,8 @@ void WindowTraits::validate()
     {
         instanceExtensionNames.push_back(VK_EXT_DEBUG_UTILS_EXTENSION_NAME);
     }
-    if (debugLayer) requestedLayers.push_back("VK_LAYER_KHRONOS_validation");
     if (apiDumpLayer) requestedLayers.push_back("VK_LAYER_LUNARG_api_dump");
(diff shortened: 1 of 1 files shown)


AnyOldName3:
As a future note, the validation layer will still log IDs as `0x3fbcd60000000028` whereas the API dump layer would log as `3FBCD60000000028`, so if you do a case-sensitive search or search with the `0x` prefix, it might look like this PR didn't work.

robertosfield:
Early in the VSG project I wrote a small pointermap utility that remaps 0x pointers from input ascii files:

https://github.com/robertosfield/PointerMap

Perhaps this should be part of vsgExamples.

PR #1443, opened 2025-04-13
Update main website link

qpwo:
https://www.vulkanscenegraph.org is down so i changed the  first link to https://vsg-dev.github.io/vsg-dev.io --- all the other links were already updated


robertosfield:
I will look into why vulkanscenegraph.org isn't redirecting.

PR #1464, opened 2025-05-14
Vague layout compatibility checks

### LINKED ISSUE #1394: Descriptor sets aren't rebound when pipeline layout change 'disturbs' them

AnyOldName3:
# Pull Request Template

## Description

Avoids disturbing necessary descriptor sets while still avoiding the need for full layout compatibility checks.

This is achieved with a basic `vector<bool>` saying which sets a pipeline layout has and doesn't have, so layouts it doesn't have won't be bound. This could mean different sets are bound than if a full compatibility check was done, but that should only happen when the scenegraph requested invalid API usage - there should only be one `BindDescriptorSet` for each slot at the top of the state stacks, and if it's for a slot the current pipeline statically uses, it must be compatible. Therefore, we only need to worry about compatibility checks for slots the current pipeline layout doesn't use, which the change here now does.

*There's a slight caveat here - if a pipeline doesn't statically use any bindings in a descriptor set, it's legal to bind nothing to it, so someone might make a scenegraph that leaves its `BindDescriptorSet` out of a stategroup and inherets something incompatible. If a descriptor set is empty, it can't be statically used, so that's treated as if there's no descriptor set. Any handling beyond that would have a larger performance overhead, so I think use cases other than this aren't worth supporting directly. If someone wants to avoid binding useless descriptor sets, there are other means, e.g. adding a dummy statecommand or not having a pointless descriptor set in the first place.*

This is based on https://github.com/vsg-dev/VulkanSceneGraph/pull/1438, even though technically it could be separated and used on its own. This is because my earlier attempts to fix the problem relied on that PR, and using the same baseline made comparing them more straightforward.

This fix is by far the fastest that helps every failure case I identified (before yesterday), with the performance impact of a worst-case scenario ranging from below the noise floor of my measurements to just 0.25% overhead depending on whether I cherry-pick a good run or my worst run. For a scenegraph that isn't designed specifically to make this have as large an impact as possible, it should be free - it does a few instructions extra work for each pipeline layout change and descriptor set binding, both of which take much longer than that already.

Fixes #1394

## Type of change

## How Has This Been Tested?

The main interesting test is the incompatible layout example here: https://github.com/vsg-dev/vsgExamples/compare/master...AnyOldName3:vsgExamples:incompatible-layouts. As discussed in #1394, it replicates the problem I initially discovered in a client app and several related situations that I discovered and thought were likely to have problems. There's lots of detail in this post https://github.com/vsg-dev/VulkanSceneGraph/issues/1394#issuecomment-2762170208. With the master branch, lots of validation errors are emitted in most test cases, but with this branch, it causes no errors at all.

Obviously, I've confirmed that this fixes the problem in the original client app.

As well as this, I've done performance testing. On my machine, the overhead is much smaller than the variation in framerate or execution time from run to run, so simply timing things didn't provide meaningful results. Instead, I found that running for a couple of minutes with a sampling-based profiler (I used the one built into Visual Studio) and comparing the percentage of samples in the modified functions was much more consistent. To get a worst-case scenario and maximise the observable overhead, I generated a scenegraph where the same basic cube is drawn thousands of times cycling between 32 different materials, with the number of draw commands chosen to give around a hundred frames per second. Trying fewer draws and more frames just made the noise from slow functions (particularly `vkCmdBeginRenderPass` and `vkBeginCommandBuffer`) overwhelm the overhead I was trying to measure, but trying a smaller number of slower frames meant they were called less and countered this out. Other variations on this style of scenegraph produced equivalent results. As mentioned above, if I cherry-pick the fastest baseline result and slowest result from this branch, and compare the number of samples in modified functions, that only accounts for 0.25% of all samples. If I don't artificially try and make things look as bad as possible, then the difference is too small to measure.

## Checklist:

- [x] I have added tests that prove my fix is effective or that my feature works. *They're on a vsgExamples branch - now the problem's fixed, it's not necessarily worth keeping the test app around.*

### DIFF
--- a/include/vsg/state/DescriptorSetLayout.h
+++ b/include/vsg/state/DescriptorSetLayout.h
@@ -37,6 +37,8 @@ namespace vsg
         /// VkDescriptorSetLayoutCreateInfo settings
         DescriptorSetLayoutBindings bindings;
 
+        virtual bool empty() const { return bindings.empty(); }
+
         /// map the descriptor bindings to the descriptor pool sizes that will be required to represent them.
         void getDescriptorPoolSizes(DescriptorPoolSizes& descriptorPoolSizes);
 
--- a/include/vsg/state/PipelineLayout.h
+++ b/include/vsg/state/PipelineLayout.h
@@ -34,6 +34,7 @@ namespace vsg
         VkPipelineLayoutCreateFlags flags = 0;
         DescriptorSetLayouts setLayouts;
         PushConstantRanges pushConstantRanges;
+        std::vector<bool> descriptorSetSlots;
 
         /// Vulkan VkPipelineLayout handle
         VkPipelineLayout vk(uint32_t deviceID) const { return _implementation[deviceID]->_pipelineLayout; }
--- a/include/vsg/state/ViewDependentState.h
+++ b/include/vsg/state/ViewDependentState.h
@@ -33,6 +33,8 @@ namespace vsg
     public:
         ViewDescriptorSetLayout();
 
+        bool empty() const override { return false; }
+
         VkDescriptorSetLayout vk(uint32_t deviceID) const override { return _viewDescriptorSetLayout ? _viewDescriptorSetLayout->vk(deviceID) : 0; }
 
         int compare(const Object& rhs_object) const override;
--- a/include/vsg/utils/ShaderSet.h
+++ b/include/vsg/utils/ShaderSet.h
@@ -177,8 +177,11 @@ namespace vsg
         /// create the descriptor set layout.
         virtual ref_ptr<DescriptorSetLayout> createDescriptorSetLayout(const std::set<std::string>& defines, uint32_t set) const;
 
-        /// return true of specified pipeline layout is compatible with what is required for this ShaderSet
+        /// return true if specified pipeline layout is compatible with what is required for this ShaderSet
         virtual bool compatiblePipelineLayout(const PipelineLayout& layout, const std::set<std::string>& defines) const;
+
+        /// return true if specified pipeline layout is partially compatible with what is required for this ShaderSet
+        virtual bool partiallyCompatiblePipelineLayout(const PipelineLayout& layout, const std::set<std::string>& defines, bool onlyPushConstants, uint32_t descriptorSet) const;
 
         /// create the pipeline layout for all descriptor sets enabled by specified defines or required by default.
         inline ref_ptr<PipelineLayout> createPipelineLayout(const std::set<std::string>& defines) { return createPipelineLayout(defines, descriptorSetRange()); }
--- a/include/vsg/vk/CommandBuffer.h
+++ b/include/vsg/vk/CommandBuffer.h
@@ -56,6 +56,7 @@ namespace vsg
         void setCurrentPipelineLayout(const PipelineLayout* pipelineLayout);
 
         VkPipelineLayout getCurrentPipelineLayout() const { return _currentPipelineLayout; }
+        const std::vector<bool>& getCurrentDescriptorSetSlots() const { return _currentDescriptorSetSlots; }
         VkShaderStageFlags getCurrentPushConstantStageFlags() const { return _currentPushConstantStageFlags; }
 
         ref_ptr<ScratchMemory> scratchMemory;
@@ -73,6 +74,7 @@ namespace vsg
         ref_ptr<Device> _device;
         ref_ptr<CommandPool> _commandPool;
         VkPipelineLayout _currentPipelineLayout;
+        std::vector<bool> _currentDescriptorSetSlots;
         VkShaderStageFlags _currentPushConstantStageFlags;
     };
     VSG_type_name(vsg::CommandBuffer);
--- a/src/vsg/state/BindDescriptorSet.cpp
+++ b/src/vsg/state/BindDescriptorSet.cpp
@@ -113,6 +113,13 @@ void BindDescriptorSets::compile(Context& context)
 void BindDescriptorSets::record(CommandBuffer& commandBuffer) const
 {
     //info("BindDescriptorSets::record() ", dynamicOffsets.size(), ", ", dynamicOffsets.data());
+    if (commandBuffer.getCurrentDescriptorSetSlots().size() < firstSet + descriptorSets.size())
+        return;
+    for (size_t slot = firstSet; slot < firstSet + descriptorSets.size(); ++slot)
+    {
+        if (!commandBuffer.getCurrentDescriptorSetSlots()[slot])
+            return;
+    }
     auto& vkd = _vulkanData[commandBuffer.deviceID];
     vkCmdBindDescriptorSets(commandBuffer, pipelineBindPoint, vkd._vkPipelineLayout, firstSet,
                             static_cast<uint32_t>(vkd._vkDescriptorSets.size()), vkd._vkDescriptorSets.data(),
@@ -209,8 +216,11 @@ void BindDescriptorSet::compile(Context& context)
 void BindDescriptorSet::record(CommandBuffer& commandBuffer) const
 {
     //info("BindDescriptorSet::record() ", dynamicOffsets.size(), ", ", dynamicOffsets.data());
-    auto& vkd = _vulkanData[commandBuffer.deviceID];
-    vkCmdBindDescriptorSets(commandBuffer, pipelineBindPoint, vkd._vkPipelineLayout, firstSet,
-                            1, &(vkd._vkDescriptorSet),
-                            static_cast<uint32_t>(dynamicOffsets.size()), dynamicOffsets.data());
+    if (commandBuffer.getCurrentDescriptorSetSlots().size() > firstSet && commandBuffer.getCurrentDescriptorSetSlots()[firstSet])
+    {
+        auto& vkd = _vulkanData[commandBuffer.deviceID];
+        vkCmdBindDescriptorSets(commandBuffer, pipelineBindPoint, vkd._vkPipelineLayout, firstSet,
+                                1, &(vkd._vkDescriptorSet),
+                                static_cast<uint32_t>(dynamicOffsets.size()), dynamicOffsets.data());
+    }
 }
--- a/src/vsg/state/PipelineLayout.cpp
+++ b/src/vsg/state/PipelineLayout.cpp
@@ -130,7 +130,7 @@ void PipelineLayout::compile(Context& context)
         for (auto dsl : setLayouts)
         {
             if (dsl) dsl->compile(context);
-            descriptorSetSlots.push_back(dsl != nullptr);
+            descriptorSetSlots.push_back(dsl != nullptr && !dsl->empty());
         }
         _implementation[context.deviceID] = PipelineLayout::Implementation::create(context.device, setLayouts, pushConstantRanges, flags);
     }
--- a/src/vsg/state/PipelineLayout.cpp
+++ b/src/vsg/state/PipelineLayout.cpp
@@ -30,7 +30,8 @@ PipelineLayout::PipelineLayout(const PipelineLayout& rhs, const CopyOp& copyop) 
     Inherit(rhs, copyop),
     flags(rhs.flags),
     setLayouts(rhs.setLayouts),
-    pushConstantRanges(rhs.pushConstantRanges)
+    pushConstantRanges(rhs.pushConstantRanges),
+    descriptorSetSlots(rhs.descriptorSetSlots)
 {
 }
 
@@ -123,9 +124,13 @@ void PipelineLayout::compile(Context& context)
 {
     if (!_implementation[context.deviceID])
     {
+        descriptorSetSlots.clear();
+        descriptorSetSlots.reserve(setLayouts.size());
+
         for (auto dsl : setLayouts)
         {
             if (dsl) dsl->compile(context);
+            descriptorSetSlots.push_back(dsl != nullptr);
         }
         _implementation[context.deviceID] = PipelineLayout::Implementation::create(context.device, setLayouts, pushConstantRanges, flags);
     }
--- a/src/vsg/state/ViewDependentState.cpp
+++ b/src/vsg/state/ViewDependentState.cpp
@@ -151,7 +151,7 @@ void BindViewDescriptorSets::compile(Context& context)
 
 void BindViewDescriptorSets::record(CommandBuffer& commandBuffer) const
 {
-    if (commandBuffer.viewDependentState)
+    if (commandBuffer.viewDependentState && commandBuffer.getCurrentDescriptorSetSlots().size() > firstSet && commandBuffer.getCurrentDescriptorSetSlots()[firstSet])
     {
         commandBuffer.viewDependentState->bindDescriptorSets(commandBuffer, pipelineBindPoint, layout->vk(commandBuffer.deviceID), firstSet);
     }
--- a/src/vsg/utils/GraphicsPipelineConfigurator.cpp
+++ b/src/vsg/utils/GraphicsPipelineConfigurator.cpp
@@ -552,7 +552,7 @@ void GraphicsPipelineConfigurator::_assignInheritedSets()
 
         void apply(const BindDescriptorSet& bds) override
         {
-            if (!bds.descriptorSet || !bds.descriptorSet->setLayout || !gpc.descriptorConfigurator) return;
+            if (!bds.layout || !gpc.descriptorConfigurator) return;
 
             if (gpc.shaderSet->partiallyCompatiblePipelineLayout(*bds.layout, gpc.shaderHints->defines, false, bds.firstSet))
             {
@@ -562,7 +562,7 @@ void GraphicsPipelineConfigurator::_assignInheritedSets()
 
         void apply(const BindDescriptorSets& bds) override
         {
-            if (!gpc.descriptorConfigurator) return;
+            if (!bds.layout || !gpc.descriptorConfigurator) return;
 
             if (gpc.shaderSet->partiallyCompatiblePipelineLayout(*bds.layout, gpc.shaderHints->defines, false, bds.firstSet + static_cast<uint32_t>(bds.descriptorSets.size()) - 1))
             {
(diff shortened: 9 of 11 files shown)


robertosfield:
I have just done a first pass review and looks good to me so far.  When I'm ready for a second pass I'll create a branch and do performance testing, I'm not expecting any issues based on my initial code reviewer :-)

AnyOldName3:
What kind of timeline can be expected for this getting the full review? It's been a month since it was opened, so it's already been quite a while.

robertosfield:
I am hoping to get most of the outstanding PRs reviewed this week.

robertosfield:
@AnyOldName3 could you give me a heads up on which of your PRs are ready to review and merge with VSG master and the priority/order that they would be best merged, and which ones are lower priority/for discussion ones.

I ask this as I have a whole bunch of work to do on many different development threads so it's hard for me to keep track of everything.

AnyOldName3:
I'll put a list of everything here to avoid having to figure out a new place.

## Soon

https://github.com/vsg-dev/VulkanSceneGraph/pull/1464 (i.e. this one) is the highest-priority PR from my perspective. It also includes the changes from https://github.com/vsg-dev/VulkanSceneGraph/pull/1438 , so it might be easier to review that one first.

https://github.com/vsg-dev/vsgXchange/pull/224 doesn't change the API or fix a bug, so nothing has an absolute dependency on it, but it makes a significant difference to text atlas generation time, so it would be nice to have sooner rather than later.

## Less soon

https://github.com/AnyOldName3/vsgXchange/pull/1 is a minor docs fix that I apparently opened to the wrong repo ages ago. I think it's slightly more useful than the equivalent thing someone else submitted later, but it's not particularly important now it's not replacing a placeholder.

https://github.com/vsg-dev/VulkanSceneGraph/pull/1296 is only a proof-of-concept, and doesn't need immediate attention.

https://github.com/vsg-dev/VulkanSceneGraph/pull/1419 and https://github.com/vsg-dev/vsgExamples/pull/350 are drafts. I've used them to investigate problems by rebasing them and fixing up merge conflicts each time I've needed to, which is a hassle, so I'd rather we figured out a way to make them mergeable but disabled by default eventually, but it's not a priority.

I think we've given up on cgitest, so https://github.com/robertosfield/cgitest/pull/1 can be ignored.

AnyOldName3:
That's a big surprise to me as the simulated worst-case test I did should genuinely have been a worst-case but didn't have nearly that much overhead, and I compared the emitted assembly for multiple compilers to ensure that I wasn't relying on vendor-specific optimisations.

Do you have a test scene you can share so I can see what's different between it and the ones I generated? The only guess I've got is that maybe yours rebinds one material descriptor over and over and it remains in L1 (which would mean doing something explicitly to dirty it), so this branch would make it spill, whereas mine switched between a few materials so they'd always need binding, so they'd be less likely to remain in L1. Either that, or your Vulkan driver has much less overhead than mine, which would be a good thing to know about for future performance testing.

robertosfield:
I don't own the test models I am using so can't share.  They were imported from old OSG models so probably stress different bits of the VSG than more modern PBR etc. models.

Longer term I'd like to have a set of scene generator utilities to create stress testing models programmatically so won't can create a range of tests and scale them up/down as appropriate.

The code looks like it can be optimized - there more if statements and variable sized containers than I think is required or optimal.  The first step I've taken on this path is to reorgnize the checks into two CommandBuffer::enabled(..) methods, this makes the code more readable and centralizes the checks so rewriting them won't change the code that depends upon them.

I have put my changes so far into a optimized-layout-compatibility-checks branch:

https://github.com/vsg-dev/VulkanSceneGraph/tree/optimized-layout-compatibility-checks

The changes so far are just to introduce the CommandBuffer::enabled() method: 334b98da809353e96be1c85b4afd8a99e3f2432f

Next up I'm curious about using a fixed size array for the enabled slots so we can avoid the checks against the container CommandBuffer::_currentDescriptorSetSlots.size().

I'm also thinking that summing the results when iterating over the _currentDescriptorSetSlots and then comparing to the size required might avoid putting if statements in a for loop.  I don't know if the test models I'm using would have BindDescriptorSet's with multiple DescriptorSet's so this might be a optimization that might not be picked up in results.

robertosfield:
Now I have put the checks all into two CommandBuffer methods we focus on where the optimizations need to be made.

First up is the simple single set check:
~~~ cpp
        inline bool enabled(uint32_t firstSet) const
        {
            return static_cast<uint32_t>(_currentDescriptorSetSlots.size()) > firstSet && _currentDescriptorSetSlots[firstSet];
        }
~~~

Getting the container size and doing the comparison could be replaced by just having a fixed sized _currentDescriptorSetSlots that is big enough for all possibilities.  A first step would be to have a std::array<bool> or just use a 32 bit mask.

The multiple set variant is more complex:

~~~ cpp
       inline bool enabled(uint32_t firstSet, uint32_t numSets) const
        {
            if (static_cast<uint32_t>(_currentDescriptorSetSlots.size()) < firstSet + numSets)
                return false;
            for (uint32_t slot = firstSet; slot < firstSet + numSets; ++slot)
            {
                if (!_currentDescriptorSetSlots[slot])
                    return false;
            }
            return true;
        }
~~~

It has a three separate conditionals that affect control flow.  We could avoid the for loop and all the conditionals by replacing it all with a set of bit mask operations.

AnyOldName3:
The implementation here's all vector of bool stuff, so it's already all bitmask operations because `std::vector` is specialised for `bool` to be a bitset. E.g. `commandBuffer.getCurrentDescriptorSetSlots()[firstSet]` should just turn into a couple of instructions to select the right address (which there are a lot more of for the other variables before the `vkCmdBindDescriptorSets` call) and then a single `bt` instruction to extract the right bit.

I'd expected the loop in `BindDescriptorSets::record` to be unrolled into a bitmask operation that would at least check a whole word worth of slots at once (which would mean only one loop iteration unless an application used more than 32 slots), and I thought I'd determined that that was happening when I looked at Godbolt, but either I changed the code since I last looked and it's disabled the optimisation or I misread the assembly as that doesn't look like it's happening for any of the major compilers with the code as it exists on my branch right now.

If we're happy to limit the VSG to 32 or 64 descriptor slots instead of the `uint32_t` that the Vulkan spec permits, then the `std::vector<bool>` could be swapped for a `std::bitset<32>`, which shaves a few instructions off the assembly.

Annoyingly, that's still not enough to convince the big three compilers to fully unroll the loop in `BindDescriptorSets::record`, but I expect I can have that working in a few minutes now I've seen it's a problem.

robertosfield:
I have replaced the use of std::vector<bool> with a uin32_t and bitmask maths and performance on my test models is back to where it was, at 1.4% faster than the original PR.  The changes require 2 less lines of code as well.  The changes are in commit: 32eb8c49d5933d1b75a980256675fc5cd6873fd7.

So the new code works faster and in my preliminary testing I can't see any regressions, but I'm just testing with standard models that worked prior to this code anyway.  So my next step is check the layout tests.  I don't recall them off the top of my head so will have to look back over past communications.

@AnyOldName3 could you do a code review of my https://github.com/vsg-dev/VulkanSceneGraph/tree/optimized-layout-compatibility-checks branch?  Thanks.

robertosfield:
inlining is hint to the compiler to inline to code where possible, if it's a f*cking awful compiler that can't inline very simple code, that this code is, then it's utter shameful mess that everyone should steer clear of.

This new code is written to be VERY easy for the compiler to optimize.  No if statements, no loops, just simple bit math.  This is how you should have written the code in the first place if you had wanted it be efficient.

I have merged your vsgincompatiblepipelinelayouts example as branch of vsgExamples and updated it to compile against the recent changes to the built-in ShaderSets.  Further details on: https://github.com/vsg-dev/vsgExamples/pull/359

I've now done as much as I can do to get this work ready to be merged with VSG master.  I now need you to provide some guidance on how to test using vsgincompatiblepipelinelayouts and to testing yourself.

AnyOldName3:
It turns out that Godbolt had changed some of my compiler flags at some point, so inlining of non-templates was disabled. Either way, now that's sorted out, I get nearly identical assembly with all three major compilers with your `uint32_t`-based implementation as the `std::bitset`-based implementation I was working on myself at the same time. That implementation's available at https://github.com/AnyOldName3/VulkanSceneGraph/tree/optimized-layout-compatibility-checks-bitset.

The `std::bitset`-based implementation currently works, but the `uitn32_t`-based one doesn't. I've isolated the problem to `PipelineLayout::compile`. It's not setting the $`n`$th bit when dealing with descriptor set $`n`$, it's setting the bit that's $`total - n`$ from the LSB. I've pushed a fixed implementation to https://github.com/AnyOldName3/VulkanSceneGraph/tree/optimized-layout-compatibility-checks-fixed.

I'll do performance testing with the scenegraphs I tested before, but I'm not anticipating any problems. On all the platforms I tried, the `vector<bool>` implementation was already a bitset and the final bit extraction was done with the `bt` instruction, so all that's changed is how it decides which address to extract a bit from, and that the size check has been removed.

robertosfield:
Sorry. I have ended up having to complete other work first. That is almost
complete. Hoping it'll be out today bed tomorrow.

AnyOldName3:
When I've been doing performance testing, I've generally found that using any real data has given me much, much smaller impacts than you've had from the same changes, typically *far* too small to measure against the variance I see from run to run, so to try and get results that indicate what you'll see, I've been designing scenegraphs that hammer the modified code paths as often as possible, and then altering them so different iterations read different data to see how much impact that has - obviously running a hundred instructions in a loop will go at different speeds if it's only reading from L1 cache. I've got a patch for vsgviewer that adds flags to duplicate either the whole scenegraph or nodes within it.

If I don't add lots of duplicates of the same data, then the frametime is nearly entirely spent in `vkBeginCommandBuffer` and vkBeginRenderpass`, which take nearly a millisecond for each call on my driver. I see that they're spending most of their runtime in `NtGdiDdDDILock2`, which is an undocumented NT function, but it's got *lock* in the name, so I think it's a safe bet that there's some blocking synchronisation between the usermode and kernel mode parts of the driver. One thought is that the VSG could be feeding commands to the driver faster than it can deal with them - if the whole scenegraph is in cache, then we can generate commands quickly, but it doesn't eliminate the need for the driver to build the command buffer in RAM and then copy it to VRAM, so we're always going to be constrained by PCIe bandwidth with dedicated GPUs even if the record traversal is instant. It could also be a more tractable problem with how we're setting up queues or command pools etc., so I've been intending to investigate it for a while.

Either way, 3% seems like a very large impact given what I've been able to measure. The transparent bin shouldn't mean the modified code was running more often than in my tests, as at least one of them included rebinding the pipeline and every descriptor set between each draw, just like a renderbin would. I don't think as much as 3% of instructions executed or memory accessed would be things that weren't happening before based on what I saw when looking at the assembly etc., but there are confounding factors like whether we're pushing things out of the cache that the driver was using, and that's not going to be consistent between hardware vendors.

I'll investigate further with translucent drawables and see if I can get different results from my previous testing.

robertosfield:
I'm seeing 3% slow down on a real-world model, it's a consistent enough result to know it's not just noise.  I don't yet know the specific cause of the bottleneck and until I have a performance benchmark written that reproduces it reliably there is no point speculating on it's cause or how to resolve it.

It really sucks that I'm seeing this performance regression because I really could without yet another unpaid task to follow up on before I can get on with tagging the next point release.  However, a 3% regression for resolving a niche issue isn't something I can just let by.  Good performance can die from a thousand paper cuts.

AnyOldName3:
I've redone my testing of this branch, the optimised version based on bitmasks and the original branch from #1438 both are based on, but with everything rebased on master so my results aren't thrown off by the command pool reset issue anymore. However, I'm still only seeing sub-half-a-percent overhead from either approach (I can't get a more precise figure as the variance between runs is more than that), so it means the command pool issue can't have been what was causing me to be unable to replicate your results. That's with and without alpha blending forcing objects into a bin, too, so it can't just be that.

One more possibility is that maybe some of your test scenes use `vsg::BindDescriptorSets` instead of just `vsg::BindDescriptorSet`. The check, particularly without the bitmask optimisation, is more complicated than the one for binding a single descriptor set, so would be expected to be more expensive. I've not included this in my testing as binding multiple descriptor sets at once is generally a sign that they should be one larger descriptor set, so it's not something I'd expect performance-sensitive apps to do.

Obviously, it'll be less hassle for both of us going forward if we can figure out why your measurements are typically massively different to mine, and I was hoping that fixing the command pool reset issue would be the silver bullet that made things consistent.

By the way, you mentioned in #1538 that you'd merged the version of this with the bitmask optimisation, but it's not clear from your phrasing whether you just meant into the https://github.com/vsg-dev/VulkanSceneGraph/compare/optimized-layout-compatibility-checks branch or whether that branch was supposed to be merged into master a while ago and you forgot it hadn't been. I'm not sure whether the 3% figure was only for the vector-of-bool implementation and the `uint32_t` implementation was sufficiently fast or whether they're both too slow and I'm still supposed to be investigating further optimisation possibilities.

robertosfield:
As a quick note, I'm really busy with client work right now so can't look at this again right away.  When I get back to it I'll complete by vsgperformance testbed to reproduce the performance issues with procedural generated scenes, after than I can start comparing the different branches to acess the performance impact and if any other optimizations are warranted.

PR #1467, opened 2025-05-16
Add self sufficient header check to CI

AnyOldName3:
# Pull Request Template

## Description

We've had a few problems with headers not being self-sufficient. CMake 3.24 added a feature to detect this automatically (for certain cases - it doesn't cover dependencies of uninstantiated templates). This uses that and enables it in CI so we know about the problem when it occurs, and also fixes all the instances of the problem that were detected.

It needs the CMake fileset system to be in charge of headers, otherwise it doesn't know which files are headers that need checking for self-sufficiency. That means there's now some conditional CMake to use that system if it's available, while leaving the old approach for systems with older CMake versions. It turns out that the new system handles a few things like installation of headers and setup of include directories automatically, so quite a bit of old code ends up disabled and can eventually be removed.

## Type of change

## How Has This Been Tested?

I built the new target locally on Windows with MSVC and fixed all the detected problems. I then added it to CI, pushed it to a branch, and saw it ran in CI, but still detected problems with GCC and libstdc++. I then fixed the problems it detected locally with GCC and libstdc++ on Windows, and pushed the fixes, which resolved the remaining issues detected by CI.


robertosfield:
I am uncomfortable with both C++ and CMake scripts having multiple code paths that do the same thing for different versions.   Having these multiple paths makes things more complicated and harder to maintain with different code paths 
being tested at different times on different platforms, errors can creep in unseen.  It's just a mess long term.

Self sufficiency of headers is good thing, so detecting it and fixing them early as possible is good thing so can we come up with a changes that allow the checks to happen without the scripts ending up a tangled mess of if else's?

The C++ header fixes would be best applied as a separate PR to any potential CMake changes so this side isn't controversial and mostly can be merged as-is.  I did spot one area where an extra include made one of the existing includes redundant (Inherit.h includes Object.h etc.), but this is just a bit sloppy it's not a problem with maintenance long term.

AnyOldName3:
My preference would be to be more aggressive with bumping the minimum CMake version, but 3.24 is from 2022, which is a little too recent for me to be comfortable forcing on everyone if they might complain. Kitware have they're going to be removing features six years after they're deprecated now, so we can expect to reliably have 3.24 available by 2028 even for the most conservative distros unless they want to be incompatible with fast-moving distros.

Failing that, sometimes I'd suggest lifting a script from CMake's repo which would make a feature available in older versions, but the stuff here relies on C++ changes, so it can't be backported without laboriously reimplementing things.

Other than that, the most practical option that doesn't involve conditional compilation would be merging the C++ changes on their own and keeping the CMake changes on a branch which is rebased when new headers are added. That keeps the mess hidden, but doesn't avoid any of the options that made it count as a mess in the first place.

robertosfield:
We have cppcheck target that we can run that does static analysis of the codebase, this is a separate target and doesn't spill over into the rest of the CMake scripts.  Might something similar be possible without slashing all the self sufficient header additions across the scripts?

AnyOldName3:
The bit that actually enables the self-sufficient header check is just
```cmake
set_target_properties(vsg PROPERTIES VERIFY_INTERFACE_HEADER_SETS ON)
```
which is pretty simple. It's sort of feasible to replace it with our own macro like `vsg_add_target_cppcheck`, but it would be more complicated than the cppcheck macro and less robust than the built-in CMake feature. There are maintenance advantages to using something Kitware maintain instead of something we've created ourselves to do the same job. That's pretty well demonstrated by the fact that the cppcheck target doesn't work on Windows due to relying on the shell to do wildcard expansion, which is opt-in on Windows (inherited from CP/M for backwards compatibility), and cppcheck hasn't opted in (unless extra arguments are passed to tell it to).

The code that supports it is that the headers are added to the target with `FILE_SET HEADERS`, but then that triggers an error if the same option isn't passed to `install(TARGETS vsg ...)` as it makes CMake think we've forgotten to install our public headers, and then adding that option makes a large part of the `CMakeLists.txt` that explicitly installs headers redundant as CMake takes over that task.

Most of the diff (excluding the C++ changes) is that I've disabled the redundant parts on CMake versions where they're redundant. I could change it so it didn't do that, and everything would work with a much smaller diff, but those lines still wouldn't do anything on newer CMake versions, it would just be less explicit that they're not doing anything. The `if()`s could be changed to comments to indicate that the lines are redundant, but to my eye, that's not any prettier than the PR as-is.

The one remaining line is changing `set_property(TARGET vsg PROPERTY CXX_STANDARD 17)` to `target_compile_features(vsg PUBLIC cxx_std_17)`, which should probably be used with or without the rest of this PR. VSG's headers won't compile with C++14 or older, so anything including them needs to use C++17 or newer, and that's what this enforces.

AnyOldName3:
*I thought I posted this yesterday, but it looks like I got distracted by the cppcheck tangent*

The bit that actually enables the self-sufficient header check is just
```cmake
set_target_properties(vsg PROPERTIES VERIFY_INTERFACE_HEADER_SETS ON)
```
which is pretty simple. It's sort of feasible to replace it with our own macro like `vsg_add_target_cppcheck`, but it would be more complicated than the cppcheck macro and less robust than the built-in CMake feature. There are maintenance advantages to using something Kitware maintain instead of something we've created ourselves to do the same job. That's pretty well demonstrated by the fact that the cppcheck target doesn't work on Windows (although now I've made a PR to fix that).

The code that supports it is that the headers are added to the target with `FILE_SET HEADERS`, but then that triggers an error if the same option isn't passed to `install(TARGETS vsg ...)` as it makes CMake think we've forgotten to install our public headers, and then adding that option makes a large part of the `CMakeLists.txt` that explicitly installs headers redundant as CMake takes over that task.

Most of the diff (excluding the C++ changes) is that I've disabled the redundant parts on CMake versions where they're redundant, although I didn't have to because they wouldn't cause an error. I could change it so it didn't do that, and everything would work with a much smaller diff, but those lines still wouldn't do anything on newer CMake versions, it would just be less explicit that they're not doing anything. The `if()`s could be changed to comments to indicate that the lines are redundant, but to my eye, that's not any prettier than the PR as-is.

The one remaining line is changing `set_property(TARGET vsg PROPERTY CXX_STANDARD 17)` to `target_compile_features(vsg PUBLIC cxx_std_17)`, which should probably be used with or without the rest of this PR. VSG's headers won't compile with C++14 or older, so anything including them needs to use C++17 or newer, and that's what this enforces.

AnyOldName3:
*This is the second day in a row that I've come to this thread and seen an old version of this comment listed as an unposted draft rather than something published, so yet again, I'm editing it to be up to date and posting it again.*

The bit that actually enables the self-sufficient header check is just
```cmake
set_target_properties(vsg PROPERTIES VERIFY_INTERFACE_HEADER_SETS ON)
```
which is pretty simple. It's sort of feasible to replace it with our own macro like `vsg_add_target_cppcheck` (I've got that on a local branch), but it would be more complicated than the cppcheck macro and less robust than the built-in CMake feature. There are maintenance advantages to using something Kitware maintain instead of something we've created ourselves to do the same job. That's pretty well demonstrated by things like the cppcheck target not working on Windows.

However, as it's a newish CMake feature, it relies on other newish CMake features. The key thing that supports it is that the headers are added to the target with `FILE_SET HEADERS`, but then that triggers an error if the same option isn't passed to `install(TARGETS vsg ...)` as it makes CMake think we've forgotten to install our public headers, and then adding that option makes a large part of the `CMakeLists.txt` that explicitly installs headers redundant as CMake takes over that task.

Most of the diff (excluding the C++ changes) is that I've disabled the redundant parts on CMake versions where they're redundant. I could change it so it didn't do that, and everything would work with a much smaller diff, but those lines still wouldn't do anything on newer CMake versions, it would just be less explicit that they're not doing anything. The `if()`s could be changed to comments to indicate that the lines are redundant, but to my eye, that's not any prettier than the PR as-is.

The one remaining line is changing `set_property(TARGET vsg PROPERTY CXX_STANDARD 17)` to `target_compile_features(vsg PUBLIC cxx_std_17)`, which should probably be used with or without the rest of this PR. VSG's headers won't compile with C++14 or older, so anything including them needs to use C++17 or newer, and that's what this enforces.

robertosfield:
I'm going to close this PR as the relevant  header changes and fixes are applied and the CMake changes a just  too convoluted for the functionality provided.  Happy to review paired down changes.

AnyOldName3:
The posts that weren't showing up are all here now, so that's annoying. Anyway, I've got a couple of alternatives.

https://github.com/vsg-dev/VulkanSceneGraph/compare/master...AnyOldName3:emulated-verify-interface-header-sets backports the 3.24 feature to work in older versions of CMake, although it needs using a little differently as file sets don't exist. I don't think it's as reliable as the native CMake feature, but it does the same thing.

https://github.com/vsg-dev/VulkanSceneGraph/compare/master...AnyOldName3:VulkanSceneGraph:self-sufficient-header-checks-2 uses the CMake feature like this PR does, but avoids conditional code wherever possible. It started with a much smaller diff than this PR, but adding an exclusion for `TracyInstrumentation.h` made it grow a few extra lines. Most of the difference is just that it's no longer explicit which lines are redundant, which I personally don't think is an advantage.

robertosfield:
I have done a quick review of your twp braches and the emulated implementation looks cleaner and more self contained.  I can't say I've review the implementation details so can't comment on that.

The hacks to work around the failure around TracyInstrumentation.h point to a problem in running the vertify code in both implementations.  Is it simply that you don't have TracyInstrumentation.h on your paths to it's failing to find the Tracy headers?  

The TracyInstrumentation.h header is self container and only ever supposed to be included into user applications that won't to use Tracy and have everything they need already installed.  It might be that one should just ignore that header in the verify code if Tracy isn't on the system.   Is there a way to turn off the verify with a pragma?

AnyOldName3:
If we were following CMake best practice, they'd probably want us to split the header off into its own header-only library which explicitly depended on Tracy. Adding it as a source file for the VSG when the VSG itself doesn't have a dependency on Tracy confuses CMake's dependency management.

The file fails to include Tracy's headers on my machine because that's normal on Windows. There's no such thing as a system-wide include directory like `/usr/include`, so having Tracy installed doesn't mean that the VSG build system can see it - only vsgExamples has been told where Tracy is, so that's the only thing that can use it.

A pragma to disable the check from within the header itself isn't possible as from the compiler's perspective, it's just compiling a normal C++ file, so it doesn't know it's doing a self-sufficient header check and wouldn't know to look for a self-sufficient-header-check opt-out pragma if one existed.

robertosfield:
That's a flaw of the CMake dependency support, creating a whole separate library for a single self contained header like TracyInstrumentation.h is just stupid.  Trying to coding around flaws in tooling is just bad programming.

These self sufficient header checks really aren't a game changer that are worthy us messing up code for, they are marginal value for a small number of programmers.  If we can support it with minimal hassle then all good, but really not worth jumping through hoops for.

AnyOldName3:
The header file *isn't* self-contained, as it depends both on the rest of the VSG and on Tracy, otherwise the check wouldn't have identified it.

Either way, moving the header to its own library isn't mandatory, which is why just putting it in its own file set within the same library worked.

PR #1468, opened 2025-05-20
Fix cppcheck on Windows

### LINKED ISSUE #1092: cppcheck-vsg does not compile with Windows10 + visual studio 2019
### Discussed in https://github.com/vsg-dev/VulkanSceneGraph/discussions/1091

<div type='discussions-op-text'>

<sup>Originally posted by **usr007** February 21, 2024</sup>
Hello, 
There is error message when compile project cppcheck-vsg with windows10 and visual studio 2019.
cppcheck: error: could not find or open any of the paths given, how did you resolve it?
It looks like the paths "include/vsg/*/*.h" and "vsg/*/*.cpp" passed to cppcheck is not expected.
It can be resolved by list all files and pass them to cppcheck within my test.
Regards.</div>

AnyOldName3:
Fixes #1092 

The gist of the problem is that as-is, the `vsg_add_target_cppcheck` macro generates a command that looks something like this:
```
/usr/bin/cppcheck -j 24 --quiet --enable=style --language=c++ /path/to/vsg/cmake/cppcheck-suppression-list.txt /path/to/vsg/include/vsg/*/*.h /path/to/vsg/include/vsg/*/*.h
```
Unix-derived platforms will expand wildcards in commands before passing them to the process, so the process gets passed a bit list of filepaths. Windows is derived from CP/M, which just passed the literal command verbatim to the process, and let it split the arguments and expand wildcards however it wanted, so that's what Windows does, too.

cppcheck *is* able to deal with wildcards, but to do that, it needs to be passed a directory instead of a file and a `--file-filter` to say which files to use from the directory. Making that change works for me on both Windows and Ubuntu, so I think it's a safe bet that it's reliable everywhere.

This might not be the best way to set cppcheck up, though. It's got built-in support for analysing CMake projects (provided you're using a generator that supports `EXPORT_COMPILE_COMMANDS`, so Makefiles or Ninja) and VS solutions (so the most popular non-`EXPORT_COMPILE_COMMANDS` generator). In that mode, it can figure out the include paths and preprocessor defines from the project, which might give better results.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -117,8 +117,10 @@ vsg_add_target_cppcheck(
     SUPPRESSIONS_LIST
         "${VSG_SOURCE_DIR}/cmake/cppcheck-suppression-list.txt"
     FILES
-        ${CMAKE_SOURCE_DIR}/include/vsg/*/*.h
-        ${VSG_SOURCE_DIR}/src/vsg/*/*.cpp
+        ${CMAKE_SOURCE_DIR}/include/
+        --file-filter=${CMAKE_SOURCE_DIR}/include/vsg/*/*.h
+        ${VSG_SOURCE_DIR}/src/
+        --file-filter=${VSG_SOURCE_DIR}/src/vsg/*/*.cpp
         -I ${VSG_SOURCE_DIR}/include/
 )
 vsg_add_target_uninstall()


robertosfield:
I wrote the cppcheck to be invoked periodically by me under Linux so haven't attempted to test under Windows so not surprised you've come across problems.

I don't know how much I want others using cppcheck as most of the output is pretty dubious with examples of what the cppcheck authors have decided is "best practice" but make for less readable and less easy to maintain code.  This means I have to write a fuck load of cppcheck suppressions to see the actual useful stuff from the forest of crap it spits out.

I'd be concerned that others might not have the experience in managing software projects long term to being able to spot the bad cppcheck stuff from the good stuff so might end up with crappy submissions trying to quieten cppcheck nonesense and introducing bugs or less maintainable code.

So... I'm not too concerned about others not being able to run cppcheck under Windows, and make the CMake script less readable for this privileged doesn't seem like a good tradeoff.  If we can make cppcheck stuff cleaner that I'd be all for that.

PR #1469, opened 2025-05-20
Cherry picked include fixes

AnyOldName3:
This should just be the uncontroversial parts of #1467

### DIFF
--- a/include/vsg/vk/Queue.h
+++ b/include/vsg/vk/Queue.h
@@ -12,6 +12,7 @@ THE SOFTWARE IS PROVIDED "AS IS", WITHOUT WARRANTY OF ANY KIND, EXPRESS OR IMPLI
 
 </editor-fold> */
 
+#include <vsg/core/Inherit.h>
 #include <vsg/core/Object.h>
 #include <vsg/vk/vulkan.h>
 
--- a/include/vsg/core/IntrusiveAllocator.h
+++ b/include/vsg/core/IntrusiveAllocator.h
@@ -16,6 +16,7 @@ THE SOFTWARE IS PROVIDED "AS IS", WITHOUT WARRANTY OF ANY KIND, EXPRESS OR IMPLI
 
 #include <list>
 #include <vector>
+#include <string>
 
 namespace vsg
 {
--- a/include/vsg/io/read_line.h
+++ b/include/vsg/io/read_line.h
@@ -10,6 +10,7 @@ THE SOFTWARE IS PROVIDED "AS IS", WITHOUT WARRANTY OF ANY KIND, EXPRESS OR IMPLI
 
 </editor-fold> */
 
+#include <cstdint>
 #include <istream>
 
 namespace vsg
--- a/include/vsg/threading/FrameBlock.h
+++ b/include/vsg/threading/FrameBlock.h
@@ -15,6 +15,8 @@ THE SOFTWARE IS PROVIDED "AS IS", WITHOUT WARRANTY OF ANY KIND, EXPRESS OR IMPLI
 #include <vsg/threading/ActivityStatus.h>
 #include <vsg/ui/ApplicationEvent.h>
 
+#include <condition_variable>
+
 namespace vsg
 {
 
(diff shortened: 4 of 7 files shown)


robertosfield:
Ooo, I think some of these extra header includes shines a light on some bugs :-)

So adding the vec3.h header to math.h just hides the bug of vec3 return types being used in stead of vec2:

~~~ cpp
template<typename T>
    t_vec3<T> operator*(const t_mat2<T>& lhs, const t_vec2<T>& rhs)
    {
        return t_vec3<T>((lhs[0][0] * rhs[0] + lhs[1][0] * rhs[1]),
                         (lhs[0][1] * rhs[0] + lhs[1][1] * rhs[1]));
    }

    template<typename T>
    t_vec3<T> operator*(const t_vec2<T>& lhs, const t_mat2<T>& rhs)
    {
        return t_vec3<T>(lhs[0] * rhs[0][0] + lhs[1] * rhs[0][1],
                         lhs[0] * rhs[1][0] + lhs[1] * rhs[1][1]);
    }
~~~

I will do a full code review of the changes and see if I can spot other cases like this.

include/vsg/utils/TracyInstrumentation.h
```diff
@@ -12,12 +12,16 @@ THE SOFTWARE IS PROVIDED "AS IS", WITHOUT WARRANTY OF ANY KIND, EXPRESS OR IMPLI
 
 </editor-fold> */
 
+#ifdef TRACY_ENABLE
```
robertosfield, line 15: This header will only ever be included by applications that use Tracy so there is do point add TRACY_ENABLE, adding it inappropriate like this would break the build of their application if they actually used TRACY_ENABLE.

I think this is an example of trying to quieten false positives and breaking things in the process.

include/vsg/maths/mat2.h
```diff
@@ -13,6 +13,7 @@ THE SOFTWARE IS PROVIDED "AS IS", WITHOUT WARRANTY OF ANY KIND, EXPRESS OR IMPLI
 </editor-fold> */
 
 #include <vsg/maths/vec2.h>
+#include <vsg/maths/vec3.h>
```
robertosfield, line 16: This one hides a bug! Now I've spotted the problem I've fixed it with commit: fc3fd403d93b1adedbac63de60ac95edf8f5c21a which makes this include inappropriate.

robertosfield:
Is there a github way of removing files from a PR?  There are changes which I'm happy with an want to merge and there others that are inappropriate that shouldn't be merged (TracyIntrumentation.h and mat2.h).

timoore:
> Is there a github way of removing files from a PR? There are changes which I'm happy with an want to merge and there others that are inappropriate that shouldn't be merged (TracyIntrumentation.h and mat2.h).

If it's the PR author or if you have permission to write to the PR branch, just make the change locally and push to the remote. That will update the PR.

Otherwise I'd manipulate the PR as a branch locally, merge it master, then close the PR by hand:
```
git fetch origin pull/1489/head:include-fixes
...
git checkout master
git merge include-fixes
git push origin @
```

robertosfield:
I know how to make changes through merging and branch then making changes, but was curious if I could just select what files to merge and what ones to merge based on the PR directly via the github web interface.

include/vsg/utils/TracyInstrumentation.h
AnyOldName3: The other parts of this file that use Tracy are already fenced off behind the same preprocessor check, so it's always required this define to be set to work, and never done anything productive when it's unset. I hadn't seen that Tracy's own headers disable most of their declarations when this is unset, and assumed it came from VSG.

It sounds like this file needs an exclusion if we're going to have a check in CI.

AnyOldName3:
I've made the relevant changes.

include/vsg/vk/Queue.h
```diff
@@ -12,6 +12,7 @@ THE SOFTWARE IS PROVIDED "AS IS", WITHOUT WARRANTY OF ANY KIND, EXPRESS OR IMPLI
 
 </editor-fold> */
 
+#include <vsg/core/Inherit.h>
 #include <vsg/core/Object.h>
```
robertosfield, line 16: The Object.h include is now redundant as it Inherit.h includes it.  I'll merge this PR and remove remove the Object.h header.

robertosfield: I have merged this PR and removed the no longer required include: 51cff49917cdaa1b58ab8dd8001a56d15d5a22d8

PR #1487, opened 2025-06-04
use rendergraph renderArea as scissor instead of viewport

theodoregoetz:
In our application we frequently have views which are partially outside of the window's render area. In this case, the viewport is used to determine aspect ratio and should be allowed to go outside of the rendering area. The scissor is used to clip the rendered scene to the smaller visible region and should not change any rendering with respect to the camera. 

The original code sets the viewport and scissor to the render graph's renderArea during record which ignores any potential difference. This PR fixes that to only set the scissor.

While this change fixes our immediate issue, I suspect there is a better solution that will be more general or more complete. I'm just not sure what's best.

This is rework of the effort to accomplish the same thing with older VSG structure around viewport state
See these PRs: 

* #1431
* #1399
* #1390
* #1331

## Type of change

- [X] Breaking change (fix or feature that would cause existing functionality to not work ~as expected~ as it did before)

## How Has This Been Tested?

In our application we have viewport states where the viewport is partially outside the window area. The scissor is then brought to be within the window area. The render graph's renderArea should then be the scissor and not the viewport when recording the render commands.

## Checklist:

### DIFF
--- a/src/vsg/app/RenderGraph.cpp
+++ b/src/vsg/app/RenderGraph.cpp
@@ -49,7 +49,8 @@ RenderGraph::RenderGraph(ref_ptr<Window> in_window, ref_ptr<View> in_view) :
         renderArea.extent = window->extent2D();
     }
 
-    viewportState->set(renderArea.offset.x, renderArea.offset.y, renderArea.extent.width, renderArea.extent.height);
+    viewportState->scissors.resize(1);
+    viewportState->scissors[0] = renderArea;
 
     // set up the clearValues based on the RenderPass's attachments.
     setClearValues(window->clearColor(), VkClearDepthStencilValue{0.0f, 0});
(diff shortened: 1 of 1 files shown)


AnyOldName3:
You should be able to get the behaviour you want without modifying the VSG by just having an explicit `vsg::View` and setting its camera's viewport state appropriately. We definitely want to have the default setup keep the viewport and scissor in sync as that's what most applications will want.

AnyOldName3:
That use case might be better served with a single rendergraph for a renderpass with multiple subpasses, each of which has its own view. Renderpasses aren't particularly cheap, and I think subpasses are supposed to be cheaper if you can live with their limitations. I think the main one is that you can only access the same pixel from previous subpasses, so it wouldn't work for something like a blur, but for OIT, I wouldn't expect that to be a problem.

PR #1513, opened 2025-06-25
Disable extra-qualification warnings with AppleClang

TP-David:
This removes the excessive extra-qualification warnings on Mac OS when using VSG_type_name macro in a vsg namespace.


TP-David:
> Now I have reverted the type_name changes I presume this PR is now not required?

Sure, if reverting had no other side effects, this PR is no longer relevan and can be closed.

PR #1524, opened 2025-07-10
add ubyteArray in intersector

mp3butcher:
I found that intersector and so on polytope don't draw indexed triangles if its indices is ubyte...Is it a mistake or a desired behavior?


robertosfield:
With Vulkan ubyte indices aren't recommend for performance reasons, legal but may not be fast paths so generally you might want to avoid them.  So... the VSG had focused on ushort and uint indices so far.

I'll merge this PR as it looks sensible as ubyte indices are legal even if they aren't recommended.

PR #1527, opened 2025-07-14
Added support for using multiple texture coordinate with built-in ShaderSets.

robertosfield:


robertosfield:
@timoore I'm open to suggestions.  

Right now we have the built-in ShaderSets that are generated from vsgshadersets utility found in vsgExamples, with the shaders in vsgExamples/data/shaders. Originally the VSG didn't include built-in shaders with vsgXchange/osg2vsg/vsgExamples all providing the required shaders, then ShaderSets came about and things have ended up more interwoven.

Originally I envisaged that the VSG source repository to just contain a single library, no utility programs or source files.  We've stuck to this.  Added shader source and possible utilities for generating pre-compiled ShaderSets is a possibility but would take us away from the simply single library build.

However, all this is a separate topic for discussion than this PR.  A thread on the discussion forum would be the best place for further discussions.

PR #1529, opened 2025-07-16
Fix typo in INSTALL.md

L0LC0D3:
In the process of reviewing the VulkanSceneGraph library, I discovered this typo. 
Why didn't anyone notice right away?


robertosfield:
FYI, I have dyslexia so spotted typos isn't my strong point, thankfully the community usually spot issues that I don't :-)

For the code we have compilers that do a good jog on picking up on inconsistencies but for documentation it's easier for typos to creep in, for that we need human eye balls.

Issue #1534, opened 2025-07-21
Allow shader defines to be created with an optional value


rms7326:
**Is your feature request related to a problem? Please describe.**
Using shader specialization constants to size global arrays doesn't work on the Mac.

We have a need in our vertex shader to size both a vertex attribute array and an array of `gl_clipDistance` by a defined value at shader compile time. We were using shader specialization constants to achieve this and on Linux and Windows with NVidia and AMD we seemed to have success even though the "Arrays inside a block" discussion in section "4.4.x Specialization-Constant Qualifier" of [ARB_gl_spirv.txt](https://registry.khronos.org/OpenGL/extensions/ARB/ARB_gl_spirv.txt) made it sound like using specialization constants to size these array members in the shader is not supported. When we tried out the same code on the Mac we had very unreliable results and were not able to get clipping to work correctly if the vertex array or the `gl_ClipDistance` array was sized by a specialization constant.

**Describe the solution you'd like**
We would like to expand the definition of a shader compiler define in VSG to allow for an optional value to be associated with the define name. Here are several proposed solutions:
1. Continue with the existing `std::string` and `std::set<std::string>` interfaces for managing defines which allow client code to add valueless defines with a simple upper case string such as, `"VERTEX_NORMALS"`. In addition to valueless defines that are currently supported, allow a define to have a `"NAME=VALUE"` syntax. This would allows a defined to be constructed as, `"NUM_PRIMITIVE_CLIPS=3"`. There are several places in `ShaderCompiler.cpp` where the name portion of the define would need to be extracted when comparing with the list of `#pragma import_defines`, and when the define is expanded to the header stream. If an equal sign is found in a define it would be replaced with a space and then added to the `headerstream` just like it is today.
2. Add a new type, `vsg::CompileDefine`, that is used in place of the `std::string` interface and maintain a set of defines with `std::set<vsg::CompilerDefine>`. The constructor for `vsg::CompileDefine` would allow for valueless or name/value construction. The changes required to VSG would be much more intrusive and we might need special handling for VSG serialization. This makes option 2 less attractive than option 1.

**Describe alternatives you've considered**
We attempted to use shader specialization constants but it wasn't super clear if they are even allowed to be used to size global arrays and we had inconsistent behavior on the Mac. Additionally, even on Linux and Windows with NVidia and AMD we had to make a default value for the shader constant that was the maximum array size. In the end we determined that using specialization constants for this purpose was not correct.

**Additional context**
[...]

PR #1539, opened 2025-07-22
Extend shader compiler define syntax to allow name=value instead of just valueless defines

rms7326:
## Description

Using shader specialization constants to size global arrays doesn't work on the Mac.

We have a need in our vertex shader to size both a vertex attribute array and an array of gl_clipDistance by a defined value at shader compile time. We were using shader specialization constants to achieve this and on Linux and Windows with NVidia and AMD we seemed to have success even though the "Arrays inside a block" discussion in section "4.4.x Specialization-Constant Qualifier" of [ARB_gl_spirv.txt](https://registry.khronos.org/OpenGL/extensions/ARB/ARB_gl_spirv.txt) made it sound like using specialization constants to size these array members in the shader is not supported. When we tried out the same code on the Mac we had very unreliable results and were not able to get clipping to work correctly if the vertex array or the gl_ClipDistance array was sized by a specialization constant.

Currently, valueless shader compiler defines are added to a `std::set<std::string>` container via the shader module's hints. This PR extents the syntax to allow `name=value` without changing the client API thereby providing an option for defines with simple values for use the in the shader code.

Fixes # (issue)
#1534 

## Type of change
Core VSG change

## How Has This Been Tested?
I extended the vsgclip example in the vsgExamples repository. See pull request 

## Checklist:

### DIFF
--- a/src/vsg/utils/ShaderCompiler.cpp
+++ b/src/vsg/utils/ShaderCompiler.cpp
@@ -10,6 +10,7 @@ THE SOFTWARE IS PROVIDED "AS IS", WITHOUT WARRANTY OF ANY KIND, EXPRESS OR IMPLI
 
 </editor-fold> */
 
+#include <vsg/core/Exception.h>
 #include <vsg/core/Version.h>
 #include <vsg/io/Logger.h>
 #include <vsg/io/Options.h>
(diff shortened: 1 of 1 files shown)


robertosfield:
Thanks for the changes.  I have done a first pass review.  It's good see how localized the changes are.

However, I think it's a bit too clever for it's own good, with the capabilities being hidden in plain sight.  You and I will know that the defines are now capable of being used this way but it won't be obvious to others without them reading an example.

I think the right solution will be to modify the public interface of defines to make it clear that it can be used with just a keyword or a <keyword, value>.  The natural implementation would be to use a std::map<std::string, std::string> but as this would break backwards compatibility we'd need to provide a mechanism to provide that capability.

Rather implement something in the short term then come back an implement it in a different way further down the line I will undertake this work myself, but I think it needs to happen after 1.1.11.

rms7326:
OK, thanks for considering it, Robert.

Issue #1538, opened 2025-07-22
Crash on record with nested views.


theodoregoetz:
Commit 32262f8 which resets the state stack during record when it visits a View made the assumption that views were not nested. In our use-cases, we make extensive use of nested views for many different reasons. I consider this commit to be a regression on established behavior and should be reverted in favor of a solution to the original problem described in Issue #1501 that allows for nested views in the scene graph.

I didn't find an example in vsgExamples right off that had nested views but maybe there is one somewhere. If not I could try to build up an example so a regression like this does not happen again.

robertosfield:
Shadows uses nested View's, but so far I haven't been aware of any regressions, but it doesn't depend upon inheriting state.

Could you create a example that illustrates the usage case of nested Views in the way that you use them and is now broken?

@rolandhill as the PR #1502 is from you, could do a chip in with the process of creating an example, or perhaps another example so we have a test for each usage so when we settle upon a final solution it works for each usage case.

rolandhill:
I don't have an example prepared and the code I'm using is part of a much bigger project. Also, the issue only presents if the state is left a certain way after the previous RecordTraversal.

I think it can be fixed easily though. The regression is occurring with nested views, so we can count the views as the RecordTraversal passes through the graph and only reset state for the top level view. I'll put a PR together for testing.

robertosfield:
@rolandhill, I can't help be feel that the _state->reset() change in your original PR is a workaround for a problem that actually lies elsewhere.

If the lazy state updating is breaking your applications usage of sibling View's then there must be some state that is not being tracked correctly such that updates that should be happening aren't, or updates are happening that aren't appropriate or required.

These issues may well be lurking in the VSG or client applications and might pop up with odd symptoms or not noticed for a very long time, but when they do pop up could be very perplexing and hard to characterize.

As the _state->reset() workaround caused problems with @theodoregoetz's usage case, and the workaround to this workaround to get this to work for that usage case is just adding an extra layer of complexity of a workaround of a workaround that likely isn't necessary at all because the actual problem is elsewhere.   

So... @rolandhill if you application has a usage case that reliably has issues with the lazy state updating then I think it would be really valuable to have a VSG example that illustrates this usage case so that we can then use it as a test case for a final proper fix rather than a workaround.  

And for the same reasons I think it would test example of @theodoregoetz's usage case so that we can make sure any changes also work for that usage case.

[...]

PR #1540, opened 2025-07-23
Reset State for top-level Views only

rolandhill:
Reset state every time the RecordTraversal encounters a top-level View. 

State needs to be reset between sibling top-level Views, otherwise it is left in an undefined status, which can lead to state not being recorded when using lazy state updating. However, some implementations using nested Views require state to persist from parent View to child View.

This PR modifies https://github.com/vsg-dev/VulkanSceneGraph/pull/1502 so that only top-level State is reset instead of resetting it for all Views. This addresses https://github.com/vsg-dev/VulkanSceneGraph/issues/1538

### DIFF
--- a/include/vsg/app/RecordTraversal.h
+++ b/include/vsg/app/RecordTraversal.h
@@ -183,6 +183,9 @@ namespace vsg
         int32_t _minimumBinNumber = 0;
         std::vector<ref_ptr<Bin>> _bins;
         ref_ptr<ViewDependentState> _viewDependentState;
+
+        // Increment and decrement the view counter so we know if we are traversing a top-level view.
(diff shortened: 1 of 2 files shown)


robertosfield:
I don't think this PR is the right approach so I'll close this PR.

I feel that using State::dirtyStateStacks() is more appropriate: 7dfa8333e3a530c9596bf73a657a1c86bed73143

If the change to dirtyStateStacks() doesn't work for your app we can investigate further, this will require creating an test program to reproduce the issues you're seeing.

rolandhill:
State::dirtyStateStacks() works for me. I'm happy to go with that if it also works for @theodoregoetz .

PR #1542, opened 2025-07-23
Adressed implicit conversion issues in quat.h

cmeyer42:
Implicit conversions in quat.h from constants (e.g. 1.0 which is a double) into value_type break compilation in VisualStudio with compiler settings "/W3 /WX".

These changes also improve performance by allowing the compiler to use the value_type variants of sin and cos.

# Pull Request Template

## Description

Motivation: see above.

We have strict compiler settings to catch conversion issues in our application code on windows (/W3 /WX). The include of vsg/maths/quat.h and subsequent use of vsg::quat (aka t_quat<float>) result in compiler warnings and subsequent errors due to the /WX setting.

The proposed fix not only adresses the compiler warnings but also enables the compiler to choose the appropriate value_type specific version of sin() and cos() which in turn might improve performance.

The applied solution is the most straight-forward one i could think of. Another solution could be the specialization of the template class for at least double- and float-types. That would then increase maintenance costs.

No dependencies.

## Type of change

## How Has This Been Tested?

The problem occured in the compilation of the CsWorldview app (from the csgCs project) with our compiler settings.

A repeat case is:
`           
#include <vsg/all.h>
int main()
{
            vsg::quat hourRot;
            hourRot.set(-1.0f, vsg::vec3(0.0f, 0.0f, 1.0f));
            return 0;
}
`
- [ ] Test A
- [ ] Test B

**Test Configuration**:
VisualStudio with compiler flags "/W3 /WX".

## Checklist:

- [? ] My code follows the style guidelines of this project
- [ x] I have performed a self-review of my own code
- [ x] I have commented my code, particularly in hard-to-understand areas
- [x ] My changes generate no new warnings
- [ x] I have added tests that prove my fix is effective or that my feature works
- [ x] New and existing unit tests pass locally with my changes
- [ x] Any dependent changes have been merged and published in downstream modules

### DIFF
--- a/include/vsg/maths/quat.h
+++ b/include/vsg/maths/quat.h
@@ -155,7 +155,7 @@ namespace vsg
 
         explicit operator bool() const noexcept
         {
-            static constexpr value_type zero = static_cast<value_type>(1.0);
+            static constexpr value_type zero = static_cast<value_type>(0.0);
             return value[0] != zero || value[1] != zero || value[2] != zero || value[3] != zero;
         }
     };
--- a/include/vsg/maths/quat.h
+++ b/include/vsg/maths/quat.h
@@ -93,7 +93,13 @@ namespace vsg
 
         void set(value_type angle_radians, const t_vec3<value_type>& axis)
         {
-            const value_type epsilon = 1e-7;
+            //@todo epsilon might have to be made value_type dependent
+            static constexpr value_type epsilon = static_cast<value_type>(1e-7);
+            // avoid implicit conversions. 
+            // a performance optimization for the sin / cos calls below which otherwise operate in double precision.
+            static constexpr value_type one = static_cast<value_type>(1.0);
+            static constexpr value_type half = static_cast<value_type>(0.5);
+
             value_type len = length(axis);
             if (len < epsilon)
             {
@@ -102,9 +108,9 @@ namespace vsg
                 return;
             }
 
-            value_type inversenorm = 1.0 / len;
-            value_type coshalfangle = cos(0.5 * angle_radians);
-            value_type sinhalfangle = sin(0.5 * angle_radians);
+            value_type inversenorm =  one / len;
+            value_type coshalfangle = cos(half * angle_radians);
+            value_type sinhalfangle = sin(half * angle_radians);
 
             x = axis.x * sinhalfangle * inversenorm;
             y = axis.y * sinhalfangle * inversenorm;
@@ -114,7 +120,12 @@ namespace vsg
 
         void set(const t_vec3<value_type>& from, const t_vec3<value_type>& to)
         {
-            const value_type epsilon = 1e-7;
+            //@todo epsilon might have to be made value_type dependent
+            static constexpr value_type epsilon = static_cast<value_type>(1e-7);
+            // avoid implicit conversions.
+            // a performance optimization for the sin / cos calls below which otherwise operate in double precision.
+            static constexpr value_type one = static_cast<value_type>(1.0);
+            static constexpr value_type half = static_cast<value_type>(0.5);
 
             value_type dot_pd = vsg::dot(from, to);
             value_type div = std::sqrt(length2(from) * length2(to));
@@ -132,9 +143,9 @@ namespace vsg
 
             double angle_radians = acos(dot_pd / div);
 
-            value_type inversenorm = 1.0 / len;
-            value_type coshalfangle = cos(0.5 * angle_radians);
-            value_type sinhalfangle = sin(0.5 * angle_radians);
+            value_type inversenorm = one / len;
+            value_type coshalfangle = cos(half * angle_radians);
+            value_type sinhalfangle = sin(half * angle_radians);
 
             x = axis.x * sinhalfangle * inversenorm;
             y = axis.y * sinhalfangle * inversenorm;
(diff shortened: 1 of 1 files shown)


include/vsg/maths/quat.h
```diff
@@ -132,17 +143,21 @@ namespace vsg
...
             x = axis.x * sinhalfangle * inversenorm;
             y = axis.y * sinhalfangle * inversenorm;
             z = axis.z * sinhalfangle * inversenorm;
             w = coshalfangle;
         }
 
-        explicit operator bool() const noexcept { return value[0] != 0.0 || value[1] != 0.0 || value[2] != 0.0 || value[3] != 0.0; }
+        explicit operator bool() const noexcept
+        {
+            static constexpr value_type zero = static_cast<value_type>(1.0);
```
robertosfield, line 158: Perhaps this might be best written as:

static constexpr value_type zero = static_cast<value_type>(0.0);

:-)

include/vsg/maths/quat.h
```diff
@@ -93,7 +93,13 @@ namespace vsg
 
         void set(value_type angle_radians, const t_vec3<value_type>& axis)
         {
-            const value_type epsilon = 1e-7;
+            //@todo epsilon might have to be made value_type dependent
```
robertosfield, line 96: Might it be worth having these epsilon, zero, half, one values be members of quat/vec2/vec3/vec4/mat2/mat3/mat4?

cmeyer42: to offer a solution for the whole codebase we might consider to introduce something like type_traits. e.g.
`vsg::numbers<value_type>::zero
vsg::numbers<value_type>::one
vsg::numbers<value_type>::half etc.`
One could add vsg::PI/PIf into that traits as well...

robertosfield: I can see value in that :-)

robertosfield: I will create a branch to experiment with the suggested vsg::numbers approach.

robertosfield:
@cmeyer42 I have merged the vsg::numbers with with PR #1546 so I'll close this PR as I believe it now superfluous. Thanks with your helping pointing the direction to a clean, albeit a bit less readable, solution.

AnyOldName3:
> Implicit conversions in quat.h from constants (e.g. 1.0 which is a double) into value_type break compilation in VisualStudio with compiler settings "/W3 /WX".

In general, you shouldn't expect other people's code to build if you apply your own warning flags to it and then enable warnings-as-errors. If your problem is just that your own project includes VSG headers and you want those warning flags for your own code, then you should instead add the VSG include directory as a system include directory so it ignores your error flags. That's done with `/external:I path/goes/here` instead of `/I path/goes/here` with MSVC and `-isystem path/goes/here` instead of `-i path/goes/here` with GCC/Clang, or have CMake do it for you. If you're using CMake 3.25 or newer, it defaults to treating imported targets as system ones, so uses these flags for their headers (it doesn't seem to be controlled by a policy, so I think it's based on the CMake version you're using rather than the `cmake_minimum_required` arguments), and with older versions, you can manually set the `INTERFACE_SYSTEM_INCLUDE_DIRECTORIES` target property to match the value of `INTERFACE_INCLUDE_DIRECTORIES `.

cmeyer42:
> > Implicit conversions in quat.h from constants (e.g. 1.0 which is a double) into value_type break compilation in VisualStudio with compiler settings "/W3 /WX".
> 
> In general, you shouldn't expect other people's code to build if you apply your own warning flags to it and then enable warnings-as-errors. If your problem is just that your own project includes VSG headers and you want those warning flags for your own code, then you should instead add the VSG include directory as a system include directory so it ignores your error flags. That's done with `/external:I path/goes/here` instead of `/I path/goes/here` with MSVC and `-isystem path/goes/here` instead of `-i path/goes/here` with GCC/Clang, or have CMake do it for you. If you're using CMake 3.25 or newer, it defaults to treating imported targets as system ones, so uses these flags for their headers (it doesn't seem to be controlled by a policy, so I think it's based on the CMake version you're using rather than the `cmake_minimum_required` arguments), and with older versions, you can manually set the `INTERFACE_SYSTEM_INCLUDE_DIRECTORIES` target property to match the value of `INTERFACE_INCLUDE_DIRECTORIES `.

Well, vsg compiles fine with its CMakeLists.txt. We don't modify those compiler settings. It is our code which includes vsg headers and which warns and throws errors bc of the /WX setting. Reasons are conversions which happen in the vsg header files.

We explicitly instrumented our VS compiler with those settings to catch performance- and design issues.

In vsg::quat I found one side effect of those conversions is that the compiler is forced to use a `double cos(double)` function instead of a `float cos(float)`. So analysing the conversion problem has revealed a potential performance problem in vsg.

I'm aware of possibilities to suppress those warnings. I just don't want to apply the suppression to vsg as vsg is an important building block of our application and we are very interested in its performance.

AnyOldName3:
At the moment, the VSG sets no warnings for MSVC (https://github.com/vsg-dev/VulkanSceneGraph/blob/master/cmake/vsgMacros.cmake#L28-L34), so if avoiding this kind of thing is important, you might be able to persuade Robert to change that. In my experience, MSVC is *much* more likely to fire a `C4244` warning than GCC is to fire a `-Wnarrowing` warning (which is enabled for the VSG via `-Wall`) when a narrowing conversion happens due to GCC defining a bunch of extra conversions as safe as a language extension, so I'd expect a lot of changes would be necessary, though.

Issue #899, opened 2023-08-04
typeid / type_info equality check fails for clang/libc++ when using VSG in dynamic libraries


martinweber:
## Issue found

I am posting this here to discuss the following behavior I found:

I have been running into issues when using VSG in a dynamic library, where the cast<> function on `vsg::Object` did return a `nullptr` even when the type was correct. This is caused by how clang/libc++ is generating `type_info.hash_code()`and the related type_info comparison operator. 

For example, in a `vsg::Visitor`:
(Note: this also happens in another area where we are using `vsg::Object::cast()`)

```
    void apply(vsg::Object& obj) override
    {
        auto* matrix_node = obj.cast<vsg::MatrixTransform>();
        // ...
    }
```

This always returned a nullptr even when the object was of type `vsg::MatrixTransform`. Using `dynamic_cast<vsg::MatrixTransform*>(&obj)` instead returned a pointer to the `vsg::MatrixTransform`.

I then logged the values for `std::type_info` in `vsg::Inherit::is_compatible()`: 

```
Subclass: N3vsg15MatrixTransformE, 4737813731 ? is_compatible: N3vsg15MatrixTransformE, 4577409791
Subclass: N3vsg9TransformE, 4737813807 ? is_compatible: N3vsg15MatrixTransformE, 4577409791
Subclass: N3vsg5GroupE, 4737813148 ? is_compatible: N3vsg15MatrixTransformE, 4577409791
Subclass: N3vsg4NodeE, 4737813255 ? is_compatible: N3vsg15MatrixTransformE, 4577409791
type_info object         : 4737813731  // at callsite of vsg::Object::cast()
type_info MatrixTransform: 4737813731 // at callsite of vsg::Object::cast()
```
(Note: the type after "? is_compatible:" is from the `type` parameter.)

* The `type_info.name()` returned the same value (N3vsg15MatrixTransformE)
* The `type_info.hash_code()` returned different values even though the type (`type_info.name()`) was identical
* The type_info comparison operator returns false for matching types

The `type_info.hash_code()` is identical at the call site, but differs in the type's `is_compatible()` function. The call site is in a different dynamic library than VSG, which is linked as static library into a different dynamic library.

This seems to be an issue with clang/libc++ when using dynamic libraries. I have found discussions about this [here](https://github.com/pybind/pybind11/issues/912) and [here](https://www.reddit.com/r/cpp/comments/rj4gob/difference_in_implementation_of_type_info_for/).

[...]

PR #1575, opened 2025-09-02
vsg::Object cast<T>() sometimes returns nullptr on macOS Sonoma arm64 Release builds

hearga:
## Description

- Addressed an issue where vsg::Object::cast<T>() would sometimes return nullptr on macOS Sonoma (Apple Silicon, arm64) in Release builds.
- Updated the casting logic to use dynamic_cast<T*>(), matching the approach already used  in the previous function " template<class T>
      ref_ptr<T> readObject(const char* propertyName)"

## Type of change

### DIFF
--- a/include/vsg/io/Input.h
+++ b/include/vsg/io/Input.h
@@ -219,7 +219,8 @@ namespace vsg
         {
             if (!matchPropertyName(propertyName)) return;
 
-            arg = read().cast<T>();
+            ref_ptr<Object> object = read();
+            arg = ref_ptr<T>(dynamic_cast<T*>(object.get()));
         }
 
         /// read a value of particular type


robertosfield:
We need to get to the bottom of why cast<> is failing, not go around relacing all of it's usage across the VSG.

Is that are particular object type that is failing?

robertosfield:
Could you have a look at this thread, I believe it touches upon the issue you are seeing:

https://github.com/vsg-dev/VulkanSceneGraph/issues/899

It looks like an issue with clang generating different type_info results for different libraries in certain circumstances.  At the end of the above thread there are suggestions for solutions.

hearga:
It works using explicit template instantiation. It resolves the different type_info hash IDs during casting.

robertosfield:
Thanks for the changes, I've done a first past review and it feels like a rather heavy handed approach, forcing lots of extra code that we don't need other than for working around a problem with a specific compiler's build settings.

The changes would also need to be rolled out to all the other use of templates such as vsg::Value, Array2D, Array3D and perhaps even use of vsg::Inherit.

Did you test building as dynamic library?

hearga:
Thanks for the review, I’ve reworked this to a minimal fix that targets the actual Clang issue (hidden visibility causing divergent &typeid(Array<T>) across TUs):

- Export the class: Array<T> to stabilize typeid.
- No explicit instantiations: removed ArrayInstantiations.cpp.
- Opt-in escape hatch: VSG_USE_DYNAMIC_CAST switches cast<T>() to dynamic_cast for those who prefer it.

If this approach looks good, we can apply the same small change to Value, Array2D/Array3D, etc., as needed, no project-wide mass changes required.

Shared libs: not tested, sorry, I have limited time to work on this; in shared builds, we already define the import/export macro, so the type info should be default-visible and typeid stable—no additional changes expected.

src/vsg/CMakeLists.txt
```diff
@@ -426,6 +426,14 @@ target_link_libraries(vsg ${LIBRARIES})
 
 if (BUILD_SHARED_LIBS)
     target_compile_definitions(vsg INTERFACE VSG_SHARED_LIBRARY)
+else()
+    if(CMAKE_CXX_COMPILER_ID MATCHES "Clang")
+        if (VSG_USE_DYNAMIC_CAST)
```
robertosfield: Where in CMake scripts is VSG_USE_DYNAMIC_CAST set & controlled?  Is there a missing CMake option for this?

include/vsg/core/Array.h
```diff
@@ -33,7 +33,7 @@ namespace vsg
 {
 
     template<typename T>
-    class Array : public Data
+    class VSG_DECLSPEC Array : public Data
```
robertosfield: If Array requires  VSG_DECLSPEC then Value, Array2D and Array3D will also require it.

robertosfield:
After a crazy busy period of client work I'm now back catching up with PRs, sorry for taking so long to get back to this.

I've done another code review and added comments into the various changes.  I think some of the changes might be sensible to merge separately so if there are ones that aren't directly dependent on the dynamic_cast<> usage then it would be good to get these fixed as separate commits/PRs.  I'm happy to do some of this, just point me in the right direction.

robertosfield:
After review the CMake and Version.h.in I have changed to use CMake option and changed the capitalization of the define to be a more direct about what it maps too: 0d7cc5ea8d9fa3df4b155a58b2b4e5fd0230a63e

PR #1600, opened 2025-11-01
utils: ComputeBounds: Consider vertex offset

blobfish:
HEADS UP: This commit breaks ABI. Might want to handle differently.

I was experimenting with some different buffer and vertex index draw configurations when i noticed compute bounds doesn't consider the VertexIndexDraw::vertexOffset member and returns incorrect results.


robertosfield:
I think it would make sense to order the method members in the same order a VertexIndexDraw.

PR #1615, opened 2025-12-02
Allow specialisation of vsg::Inherit for sublasses, alternative implentation

AnyOldName3:
An alternative to https://github.com/vsg-dev/VulkanSceneGraph/pull/1609

The original PR's description, which is almost entirely relevant to this one, too:

> This allows default implementations of members to be provided, prevents copying and pasting of lots of trivial functions, and makes it possible to generate a compiler error if a subclass fails to provide a particular member without making it a pure virtual function.
> 
> This is a pretty close equivalent to how in the OSG, there was a `META_Object` macro that did roughly what the VSG's Inherit template does, but also there were specialised macros like `META_Node` and `META_StateAttribute` to do extra things for particular kinds of object.
> 
> I've only used this for `vsg::ArrayState` so far as that was the place where the problem came up in my work (I needed to add extra `cloneArrayState` overloads and ensure subclasses overrode them). I think it's a fairly safe bet that there are other places where using this would be helpful, as I know I've been annoyed by the lack of such a feature in the past, although I can't remember when or why.
> 
> This isn't the prettiest code in the world, ~~and has a bonus extra annoying feature that it makes the Visual Studio debugger add an extra level of indentation to loads of things~~, but it's the best that can be done until P3469 https://www.open-std.org/jtc1/sc22/wg21/docs/papers/2024/p3469r0.pdf is ratified and becomes part of C++ (it's too late to get into C++26) and finally makes CRTP fully obsolete.

Unlike the original PR:
* for classes that don't use this feature, nothing about them changes as all the templates expand to exactly what was used before.
* for classes that do use this feature, there's no need to specify complicated template conditions as that's handled automatically.
* it relies on more basic template features.

### DIFF
--- a/include/vsg/core/Inherit.h
+++ b/include/vsg/core/Inherit.h
@@ -20,16 +20,29 @@ THE SOFTWARE IS PROVIDED "AS IS", WITHOUT WARRANTY OF ANY KIND, EXPRESS OR IMPLI
 
 namespace vsg
 {
+    template<class ParentClass, class Subclass, class Enable = void>
+    struct InheritHelper
+    {
+        using Type = ParentClass;
+    };
+
+    template<class ParentClass, class Subclass>
+    struct InheritHelper<ParentClass, Subclass, typename ParentClass::template InheritExtras<ParentClass, Subclass>>
+    {
+        using Type = typename ParentClass::template InheritExtras<ParentClass, Subclass>;
+    };
 
     /// Inherit<> uses the Curiously Recurring Template Pattern
     /// to provide the classes versions of create, accept(..), RTTI and sizeofObject()
     template<class ParentClass, class Subclass>
-    class Inherit : public ParentClass
+    class Inherit : public InheritHelper<ParentClass, Subclass>::Type
     {
     public:
+        using BaseClass = typename InheritHelper<ParentClass, Subclass>::Type;
+
         template<typename... Args>
         Inherit(Args&&... args) :
-            ParentClass(std::forward<Args>(args)...) {}
+            BaseClass(std::forward<Args>(args)...) {}
 
         template<typename... Args>
         static ref_ptr<Subclass> create(Args&&... args)
@@ -47,11 +60,11 @@ namespace vsg
         std::size_t sizeofObject() const noexcept override { return sizeof(Subclass); }
         const char* className() const noexcept override { return type_name<Subclass>(); }
         const std::type_info& type_info() const noexcept override { return typeid(Subclass); }
-        bool is_compatible(const std::type_info& type) const noexcept override { return typeid(Subclass) == type || ParentClass::is_compatible(type); }
+        bool is_compatible(const std::type_info& type) const noexcept override { return typeid(Subclass) == type || BaseClass::is_compatible(type); }
 
         int compare(const Object& rhs) const override
         {
-            int result = ParentClass::compare(rhs);
+            int result = BaseClass::compare(rhs);
             if (result != 0) return result;
 
             size_t startOfSubclass = sizeof(ParentClass);
--- a/include/vsg/state/ArrayState.h
+++ b/include/vsg/state/ArrayState.h
@@ -27,6 +27,28 @@ namespace vsg
     class VSG_DECLSPEC ArrayState : public Inherit<ConstVisitor, ArrayState>
     {
     public:
+        template<class ParentClass, class Subclass>
+        class InheritExtras : public ParentClass
+        {
+        public:
+            using ParentClass::ParentClass;
+
+            explicit InheritExtras(const ArrayState& rhs, const CopyOp& copyop = {}) :
+                ParentClass(rhs, copyop)
+            {
+            }
+
+            ref_ptr<ArrayState> cloneArrayState() override
+            {
+                return Subclass::create(static_cast<Subclass&>(*this));
+            }
+
+            ref_ptr<ArrayState> cloneArrayState(ref_ptr<ArrayState> arrayState) override
+            {
+                return Subclass::create(static_cast<Subclass&>(*arrayState));
+            }
+        };
+
         ArrayState() = default;
         ArrayState(const ArrayState& rhs, const CopyOp& copyop = {});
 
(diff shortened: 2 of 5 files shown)


robertosfield:
Similar story. Closing this PR as it complicates the code base with little useful benefit.

AnyOldName3:
I don't think a reasonable person would say this complicates anything overall. This PR removes twice as many lines as it adds, and the added lines are all basic C++ that should be obvious to anyone with a reasonable amount of experience with C++. It might take someone a moment to work out how the partial template specialisation is working, but anyone who knows enough about templates to be aware that they're different to Java's generics etc. should be able to recognise that it *is* partial template specialisation, which is enough to reason about what it's doing and how to use it. CRTP is a *much* more advanced template idiom than anything used here, and that's the whole deal with `Inherit.h`, so the changes are trivial compared to the surrounding code.

The code that's replaced, however, despite looking simple, is pretty insidious, as without this PR, it's mandatory, but there's no machine-enforced check that it exists or is correct. It's bad for exactly the same reasons as a world without `vsg::Inherit` where everyone had to write their own `create` and `className` implementations would be bad. As `ArrayState` is a class that applications might want to subclass, the existing subclasses including the necessary boilerplate doesn't mean that it's safe to ignore the latent footgun here, even if we assume that there'll never be another use case for this that will crop up later.

robertosfield:
Complexity isn't just lines of code.  It's added an extra layer of inheritance.

ArrayState is not in it's final state.  It's part of the VSG that solves an awkward problem - how to provide a C++ mapping of shader behavior, but it solves it in an awkward and sub-optimally efficient way.  For me it's part of the scene graph design that isn't yet properly solved. 

So at some point in the future I'll return to awkward issue of how to do something like ArrayState but improve it's design and implementation so it's easier to extend and more efficient.  It's likely that much of the old ArrayState approach will go along with it the current methods that you are trying to streamline.

I don't know how it'll look, I just need to spend time with the code and how the functionality is implemented and used and just bang my head against a wall until something cleaner drops out.  I've done this quite a few times through the life of OpenSceneGraph and VulkanSceneGraph projects, I instinctively know when there is a better solution lurking out there.

I also instinctively know when a something is proposed that takes us in the wrong direction.  These instincts aren't something I was born with, they have been developed over decades of working on code myself and working with the code of others.  

If solutions get merged that aren't right then cause more problems down the line, as project lead the buck stops with me, it doesn't matter whether I created the problems or others it's down to me to make sure fixes are made to the design and implementation to address them.  This rather sharpens ones view of changes that are going to be manageable long term and ones that take one in the wrong direction.

I hope you develop these same instincts.  You are bright enough not to fear complexity and at times don't see complexity where others might so you dive in and go in directions that my spidey sense make me pause and think twice.  

I'm just as bright as you but have this instinctive drive towards trying to simplify things where possible and not to entirely settle till I've got things right.  Most of the time we don't get to those perfect solutions on first try, often we have to make do with imperfect while we learn more about the problem domain then return to the area with a new perspective in order to make that leap.

If I reject PRs, it's because we can and must do better.

AnyOldName3:
> Complexity isn't just lines of code. It's added an extra layer of inheritance.

I never said it was only lines of code, I asserted that the removed lines were worse than the added ones, too. Inheritance is a basic C++ feature, and this PR uses it in a trivial way to do something obvious that there's no need to think about unless you decide to opt in. The removed code looks simple, but people interacting with the class have to realise what it's doing and why and that it's mandatory that they copy it to avoid problems that would look like they're caused by another system, e.g. culling or intersection. You or I would know to suspect a `cloneArrayState` implementation was missing, but anyone new to the VSG wouldn't. If code is complicated to use correctly, then it's complicated.

As I alluded to in the original PR description, there have been other, non-`ArrayState` use cases where having this available would have allowed me to make something more simple and/or reliable than it ended up being. I waited until the same problem had come up repeatedly before I put time into coming up with a reusable solution.

> ArrayState is not in it's final state. It's part of the VSG that solves an awkward problem - how to provide a C++ mapping of shader behavior, but it solves it in an awkward and sub-optimally efficient way. For me it's part of the scene graph design that isn't yet properly solved.
> 
> So at some point in the future I'll return to awkward issue of how to do something like ArrayState but improve it's design and implementation so it's easier to extend and more efficient. It's likely that much of the old ArrayState approach will go along with it the current methods that you are trying to streamline.

That's the kind of thing this PR helps with. My intersection optimisation work already makes some changes to `ArrayState`. User subclasses still compile without the feature from this PR, but they'd misbehave in potentially subtle ways at runtime. With this PR, I can make the compiler either generate the required new methods or emit an error, so there's no false impression that everything should work. Obviously, if enough of the case class changes, then user code will stop compiling, so it might not be especially applicable to `ArrayState`, but it's always bad if user code keeps compiling but stops working. I can all but guarantee there'll be *some* kind of change that needs making in the future that having this available will help limit the impact of.

> I also instinctively know when a something is proposed that takes us in the wrong direction. These instincts aren't something I was born with, they have been developed over decades of working on code myself and working with the code of others.
> 
> If solutions get merged that aren't right then cause more problems down the line, as project lead the buck stops with me, it doesn't matter whether I created the problems or others it's down to me to make sure fixes are made to the design and implementation to address them. This rather sharpens ones view of changes that are going to be manageable long term and ones that take one in the wrong direction.
> 
> I hope you develop these same instincts. You are bright enough not to fear complexity and at times don't see complexity where others might so you dive in and go in directions that my spidey sense make me pause and think twice.

You've given me plenty of reasons over the years not to automatically trust your instincts on some matters more than anyone else's, so to verify things some other way, too, just like I would with my own instincts. I'm not completely unaware of the difficulties maintaining a large project - remember I've been one of OpenMW's maintainers for years, and that's a very large project. I've reached different conclusions to you about what's more likely to cause problems based on what I've seen cause problems in the past, not because I've never seen things cause problems. I have a healthy fear of complexity, but don't seem to see it in exactly the same places as you do, and am willing to accept that complex situations rarely can adequately be handled by simple code.

> I'm just as bright as you but have this instinctive drive towards trying to simplify things where possible and not to entirely settle till I've got things right. Most of the time we don't get to those perfect solutions on first try, often we have to make do with imperfect while we learn more about the problem domain then return to the area with a new perspective in order to make that leap.
> 
> If I reject PRs, it's because we can and must do better.

Most of the time I find that parts of the problem domain aren't handled, and try and add anything that handles them, you're dismissive that what I've discovered is really part of the problem domain and don't think the fix is simple enough, but rarely suggest anything as an alternative that, to me, seems not to just be a workaround that pushes a greater amount of complexity somewhere else.

If I'm opening a PR without explicitly saying it's only a proof of concept, then it's because I think it's close to as good as we're going to get (often after looking up what typically ends up working in similar situations - not everything the VSG does is completely new ground with no prior art). If I can't come up with something better, and you don't come up with something better, then there's no compelling reason to believe that there's a better approach out there that we can reasonably expect to discover.

PR #1624, opened 2025-12-13
Enforce using C++17 when linking against vsg in CMake

vikhik:
## Description

Use target_compile_features to set C++17 standard, with PUBLIC scope.
This is the modern CMake preferred way to set a required C++ standard.
This now means that users do not need to set 17 for themselves, but will automatically be set when using `target_link_libraries` as a _minimum_ of 17. Users can still choose to set higher standards if desired.

Fixes # (issue)

## Type of change

- [x] This change requires a documentation update

## How Has This Been Tested?

- `cmake -B build . --G "Ninja Multi-Config"

**Test Configuration**:

* `cmake version 3.31.6`


robertosfield:
Changing the minimum cmake version is not without risks of breaking support for those unable to upgrade.

The cmake 3.7 version requirement was set purely for backwards compatibility with older dristos version of cmake. Historically some users have been tied to older distros and restricted on the dependencies that are able to install dependencies.

So the question now is what restrictions the community still have.  We haven't polled the community recently so tomorrow I'll write a post asking for feedback on OS, CMake, Compiler restrictions.

robertosfield:
I have created a thread to poll for platforms and tools folks are using: #1627

PR #1628, opened 2025-12-18
Fixes a misspelling in validateInstancelayerNames

cubicool:
# Pull Request Template

## Description

I think this is just another minor misspelling propagated by autocomplete. I'm assuming `validateInstanceLayerNames` was the intended camel-case to be consistent with everything else?

Fixes # (issue)

## Type of change


robertosfield:
Separate PRs are fine, the only thing I'd tweak would be the naming of the PR/commit so that if it's spelling typo then label it as such, but if it's just a clean up capitalization then make this clear so I know what to expect.  Thanks.

PR #1631, opened 2025-12-19
fix: building projects that use VulkanSceneGraph as a dynamic library…

Vic-Min:
… with VS2022

# Pull Request Template

## Description

Using 'VSG_DECLSPEC' with the Array, Array2D, Array3D, and Value templates causes the error "C2491: definition of dllimport function not allowed" when building vsgXchange, vsgExamples, and vsgQt.

Missing 'VSG_DECLSPEC' for State causes the error "LNK2019: unresolved external symbol" when building vsgExamples/vsgcompute.

## Type of change

## Checklist:

### DIFF
--- a/include/vsg/core/Array.h
+++ b/include/vsg/core/Array.h
@@ -32,7 +32,7 @@ THE SOFTWARE IS PROVIDED "AS IS", WITHOUT WARRANTY OF ANY KIND, EXPRESS OR IMPLI
 namespace vsg
 {
     template<typename T>
-    class Array : public Data
+    class VSG_TEMPLATE_DECLSPEC Array : public Data
     {
     public:
         using value_type = T;
--- a/include/vsg/core/Array2D.h
+++ b/include/vsg/core/Array2D.h
@@ -30,7 +30,7 @@ namespace vsg
 {
 
     template<typename T>
-    class Array2D : public Data
+    class VSG_TEMPLATE_DECLSPEC Array2D : public Data
     {
     public:
         using value_type = T;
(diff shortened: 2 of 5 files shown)


robertosfield:
Um.... the addition of VSG_DECLSPEC to Array etc. was added with PR #1575  in order to address Issue #899, just removing them will then break this PR and bring back the Issue it resolved.

So.... ideas @hearga and @Vic-Min?  I have zero interest in ping-ponging between two incompatible fixes.  We need a solution that doesn't break the build/execution for Windows vs macOS.

hearga:
@Vic-Min and @robertosfield,  what if Array.h, Arra2D.h, Array3D.h and Value.h replace `VSG_DECLSPEC ` by `VSG_TEMPLATE_DECLSPEC`

and then in Export.h

```
#if defined(WIN32)
#    if defined(vsg_EXPORTS)
#        define VSG_DECLSPEC __declspec(dllexport)
#    elif defined(VSG_SHARED_LIBRARY)
#        define VSG_DECLSPEC __declspec(dllimport)
#    else
#        define VSG_DECLSPEC
#    endif
#    define VSG_TEMPLATE_DECLSPEC
#else
#    if defined(VSG_SHARED_LIBRARY) || defined(VSG_EXPORTS)
#        define VSG_DECLSPEC __attribute__((visibility("default")))
#        define VSG_TEMPLATE_DECLSPEC __attribute__((visibility("default")))
#    else
#        define VSG_DECLSPEC
#        define VSG_TEMPLATE_DECLSPEC
#    endif
#endif
```

PR #1636, opened 2025-12-23
Simplified find package glslang so that glslang 16.x is supported out of the box.

robertosfield:


AnyOldName3:
This *might* cause problems on distros that provide glslang 13 or older, as it'll be picked up by `find_package`, and then fail to build. Hopefully, it's old enough that no one's using it anymore.

PR #1657, opened 2026-01-21
Fix vkCmdBeginRenderPass SYNC-HAZARD-WRITE-AFTER-READ error

TP-David:
These errors appear when enabling "Synchronization" Layer Configuration in Vulkan Configurator and running vsgdraw and vsgtext:

`Validation Error: [ SYNC-HAZARD-WRITE-AFTER-READ ] | MessageID = 0x376bc9df vkQueueSubmit(): WRITE_AFTER_READ hazard detected. vkCmdBeginRenderPass: ...`

Fix in vsg::createRenderPass:
Add **VK_ACCESS_COLOR_ATTACHMENT_WRITE_BIT** to colorDependency.srcAccessMask and depthDependency.dstAccessMask
Add **VK_PIPELINE_STAGE_COLOR_ATTACHMENT_OUTPUT_BIT** to depthDependency.srcStageMask and depthDependency.dstStageMask

Fix in vsg::createMultisampledRenderPass (for multi-sampled windows):
Add **VK_PIPELINE_STAGE_EARLY_FRAGMENT_TESTS_BIT | VK_PIPELINE_STAGE_LATE_FRAGMENT_TESTS_BIT** to dependency.dstStageMask and dependency2.srcStageMask
Add **VK_ACCESS_DEPTH_STENCIL_ATTACHMENT_READ_BIT | VK_ACCESS_DEPTH_STENCIL_ATTACHMENT_WRITE_BIT** to dependency.dstAccessMask and dependency2.srcAccessMask

OS Windows 11
GPU NVIDIA RTX 2000 Ada Generation Laptop GPU
Driver date 10/12/2025
Vulkan 1.4.328.1

PR #1665, opened 2026-01-28
Enable synchronization validation feature if layer is provided

TP-David:
Required to print synchronization validation errors
https://github.com/vsg-dev/VulkanSceneGraph/pull/1657#issuecomment-3812235074


robertosfield:
I have [generalized](da3f114325308e3c1f756bf108e80811acce2254) the set up of the VkValidationFeaturesEXT to make it easier to add further enabled/disabled validation features, and then merged this with VSG master with PR #1667.

PR #1668, opened 2026-01-31
Window traits clear color

paveltrpn:
# Add clear color field to WindowTraits and use it as window clear color.

## Description

I recently start using VulkanSceneGraph and did not find any way to change clear color of the window.  
This PR add corresponding field to WindowTraits and force to use it as window background clear color.
Default value of clear color stays same as before.

Additionally i add .qtcreator folder (QtCreator setting folder starts from 19.0) to .gitignore file.

### DIFF
--- a/include/vsg/app/WindowTraits.h
+++ b/include/vsg/app/WindowTraits.h
@@ -98,6 +98,8 @@ namespace vsg
         std::any nativeWindow;
         std::any systemConnection;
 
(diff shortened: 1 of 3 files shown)


robertosfield:
FYI, to set the clear colour of a window you simply set via the [window->clearColor() method](github.com/vsg-dev/VulkanSceneGraph/blob/master/include/vsg/app/Window.h#L66) i.e.

~~~ cpp
windowColor->set(r,g,b,a);
~~~

PR #1690, opened 2026-04-11
Fix typo in CompileResult::add()

ArepMM:
We have rare crashes of viewer after switching to other location in scene while it was loading multiple of models with PageLODs. It was hard to investegate, once we found that before crash some models are loaded with wrong or corrupted textures. Then we check the source code of the compilation and adding PagedLODs to the scene, and found this typo in handling of dynamic buffers and images at CompileResult. After building with fix we have no more crashes.


robertosfield:
Well spotted.  Thanks for the fix.

PR #1692, opened 2026-04-13
Fix 'using namespace tracy' at global scope in a header file

Vic-Min:
## Type of change

## How Has This Been Tested?

```
#include <vsg/all.h>
#ifdef TRACY_ENABLE
#    include <vsg/utils/TracyInstrumentation.h>
#endif

struct Color{};

Color color;
```

`error C2872: Color: ambiguous symbol can be "Color" or "tracy::Color"`

## Checklist:

### DIFF
--- a/include/vsg/utils/TracyInstrumentation.h
+++ b/include/vsg/utils/TracyInstrumentation.h
@@ -17,10 +17,10 @@ THE SOFTWARE IS PROVIDED "AS IS", WITHOUT WARRANTY OF ANY KIND, EXPRESS OR IMPLI
 
 #include <vsg/utils/Instrumentation.h>
 
-using namespace tracy;
-
 namespace vsg
 {
+
+    using namespace tracy;
 
     class TracySettings : public Inherit<Object, TracySettings>
     {


robertosfield:
By putting the using namespace tracy inside the vsg scope, the namespace becomes vsg::tracy, which is not what is intended.

What platform are you working on?  What version of tracy?  Does the vsgtracyinstrumentation example compile for you?

Vic-Min:
> By putting the using namespace tracy inside the vsg scope, the namespace becomes vsg::tracy, which is not what is intended.

This is not true, the namespace does not become vsg::tracy:
```
namespace tracy
{
    struct Color{};
};

namespace vsg 
{
    using namespace tracy;
};

tracy::Color color; // Ok
vsg::tracy::Color color; // error: 'tracy' in namespace 'vsg' does not name a type
```
> What platform are you working on?

Win10, VS2022

> What version of tracy?

Last release 0.13.1

> Does the vsgtracyinstrumentation example compile for you?

The `vsgtracyinstrumentation` example compiles.
My code, which defines the Color class, doesn't compile.
MRE - in the first post.

I proposed this solution as a minimal one that solves the problem.
In my opinion, a better solution would be to add `using namespace tracy` to the beginning of the `enter`/`leave` methods and replace `VkCtx` with `tracy::VkCtx`.
I'm ready to submit a PR for this.

robertosfield:
I have just updated my old tracy checkout to 0.13.1 and the library compiles fine but the profiler is broken due to a new Wayland dependency.  I'd guess my Kubuntu 24.04 systems Wayland libs are older than the latest Tracy is built against.  Arrgg what a pain.

In generally I prefer not using using namespace in headers, when I originally wrote the VSG tracy support I followed 3rd party examples and Tracy itself for inspiration.  That was a while back and haven't touched it much since.

robertosfield:
To test I added:

~~~ cpp
struct Color{};

Color color;
~~~

to vsgtracyinstrumentation and then commented out the using namespace tracy and fixed build issues by putting in explict tracy:: in the TracyInstrumentation.h header.  With these changes the above compiles in vsgtracyinstrumentation: 24f21dab153448337115672c5f543d521b423521

I will close this PR as VSG master should now work fine for you.

PR #1697, opened 2026-04-16
Refactor how GraphicsPipelineCongfigurator::compare() handles the shaderHints->defines and descriptorConfigurator->defines together to avoid failure to match compatible configurations.

robertosfield:


swp-ariaci:
This change appears to introduce a regression.

While testing the updated comparison logic, I ran into a crash in previously working code caused by dereferencing `GraphicsPipelineConfigurator::descriptorConfigurator` without a null check.

I’ve added a minimal reproducible example and details here:
https://github.com/vsg-dev/VulkanSceneGraph/issues/1682#issuecomment-4296883651

In short, `descriptorConfigurator` is accessed unconditionally in `compare()`, but it can be null in practice, which leads to a crash during `SharedObjects::share(..., init)`.

It looks like a null check is needed before accessing it.

Issue #1711, opened 2026-05-15
PagedLOD / DatabasePager can result in continual GPU memory growth


jamierobertson1:
Since the recent changes to the DatabasePager and new MemoryBufferPools resource mechanism, I've noticed my streaming PagedLOD pointclouds cause GPU memory to grow continually when zooming in and out until it eventually runs out and the database pager fails to compile nodes anymore.

It looks like the frequent loading and unloading of large number of tiles cause the MemoryPools become increasing fragmented, causing new pools to be frequently created when loading tiles. Pointcloud databases are probably particularly bad for this as tiles can vary a lot in size, and the tiles are loaded / unloaded frequently.

I've done quite a bit of testing and I don't think there's actual bugs causing the memory growth, inactive tiles are correctly unloaded, and slots in existing pools are reclaimed correctly when they have enough space. 

Possibly it might help batching the geometry in my pointcloud databases into fixed size arrays?
  
**To Reproduce**
Open a reasonable sized vsgPoints paged LOD database and zoom in and out, so tiles are repeatedly loaded / unloaded and monitor GPU memory.

**Expected behavior**
GPU memory to stabilise at some ceiling, as happened with previous vsg versions. Possibly targetMaxNumPagedLODWithHighResSubgraphs was a bit more ruthless when removing nodes.

**Desktop (please complete the following information):**
 - OS: Windows 11

robertosfield:
Over the last couple of months I have done lots of work to help avoid the crashes due to memory growth when paging, and added some basic load balancing into the DatabasePager/CompileManager to enable more aggressive reuse when memory limits are hit. So... I was hoping these issues would become less problematic.

In testing I have been using large 3D Tiles database that were previously over-consuming GPU memory and causing crashes, and these now work better, but not yet as good as I'd like w.r.t reusing of expired subgraphs, and like you I suspect memory fragmentation as an issue.  

To help refine the DatabasePager/CompileManager/vsgPoints further I'll need a test case that I can run.  Potentially a synthetic test in vsgpoints might be a way of doing, or just picking a big public dataset that we can all download and test against.  

Recording a camera animation path that recreates problems is what I find really helpful when debugging these issues, so creating a shared one would be useful as well.

Essentially I'd like to be able to have myself and others run the same test, get rests, refine the code, re-run it across multiple systems and make sure the results are going in the right direction.

[...]

PR #1712, opened 2026-05-16
Fix Device::availableMemory crash on Vulkan 1.0 instances

mbait:
## Description

`Device::availableMemory()` unconditionally calls `vkGetPhysicalDeviceMemoryProperties2` with a chained `VkPhysicalDeviceMemoryBudgetPropertiesEXT`. Both have preconditions that are never checked:

- `vkGetPhysicalDeviceMemoryProperties2` requires Vulkan 1.1 (or `VK_KHR_get_physical_device_properties2`, promoted to core in 1.1).
- `VkPhysicalDeviceMemoryBudgetPropertiesEXT` requires the `VK_EXT_memory_budget` device extension.

When a VSG application creates a `vsg::Instance` with `VK_API_VERSION_1_0` (still the default of `vsg::Instance::create`), this function dispatches an unresolved entry point and feeds the driver a struct it never advertised support for. Behaviour is undefined. On Mesa Lavapipe it segfaults inside a driver worker thread during normal scene compilation (`MemoryBufferPools::reserveBuffer` -> `Device::availableMemory`). The Khronos validation layer flags it as:

> `vkGetPhysicalDeviceMemoryProperties2()`: Attempted to call with an effective API version of 1.0.0 … but this API was not promoted until version 1.1.0.

## Fix

Gate the 1.1 path on both `supportsApiVersion(VK_API_VERSION_1_1)` and `supportsDeviceExtension(VK_EXT_MEMORY_BUDGET_EXTENSION_NAME)`. When either is missing, fall back to the Vulkan 1.0 `vkGetPhysicalDeviceMemoryProperties` and treat `memoryHeaps[i].size` as the budget with zero live usage. The fallback returns a conservative upper bound — buffer pool sizing degrades gracefully instead of crashing.

No public API changes. The 1.1 + extension code path is unchanged for users who request it.

## Type of change

## How Has This Been Tested?

Reproduced and verified on Mesa 25.2.8 Lavapipe (`lvp_icd.json`) in a GPU-less Linux container with a minimal headless VSG application (`vsg::Instance::create(..., VK_API_VERSION_1_0)`, offscreen framebuffer, one `Builder::createSphere` frame).

- Before the patch: SIGSEGV inside `libvulkan_lvp.so` during `viewer->compile()`, validation layer reports the API-version violation above.
- After the patch: clean run, no validation errors, expected image written.

**Test Configuration**:
- OS: Ubuntu 24.04 (LXC container, no GPU)
- Vulkan loader: 1.3.275
- Driver: Mesa 25.2.8 Lavapipe (LLVM 20.1.2)
- Compiler: GCC 13.3.0

## Checklist

- [x] My code follows the style guidelines of this project (clang-format clean)

### DIFF
--- a/include/vsg/vk/Device.h
+++ b/include/vsg/vk/Device.h
@@ -85,6 +85,9 @@ namespace vsg
         /// return true if Device was created with specified extension
         bool supportsDeviceExtension(const char* extensionName) const;
 
+        /// commonly checked extensions cached at construction
+        const bool memory_budget = false; // VK_EXT_memory_budget usable on this Device
+
         /// return the amount of remaining memory, compatible with specified flags, available that can be allocated.
         VkDeviceSize availableMemory(VkMemoryPropertyFlags memoryPropertiesFlags, double allocatedMemoryLimit = 1.0) const;
 
--- a/src/vsg/vk/Device.cpp
+++ b/src/vsg/vk/Device.cpp
@@ -163,6 +163,11 @@ Device::Device(PhysicalDevice* physicalDevice, const QueueSettings& queueSetting
     }
 
     _extensions = DeviceExtensions::create(this);
+
+    // Cache extension/version probes used on hot paths. VK_EXT_memory_budget piggy-backs on
+    // vkGetPhysicalDeviceMemoryProperties2 which requires Vulkan 1.1 (or the KHR variant).
+    const_cast<bool&>(memory_budget) = supportsApiVersion(VK_API_VERSION_1_1) &&
+                                       supportsDeviceExtension(VK_EXT_MEMORY_BUDGET_EXTENSION_NAME);
 }
 
 Device::~Device()
@@ -208,17 +213,24 @@ bool Device::supportsDeviceExtension(const char* extensionName) const
 
 VkDeviceSize Device::availableMemory(VkMemoryPropertyFlags memoryPropertiesFlags, double allocatedMemoryLimit) const
 {
-    VkPhysicalDeviceMemoryBudgetPropertiesEXT memoryBudget;
-    memoryBudget.sType = VK_STRUCTURE_TYPE_PHYSICAL_DEVICE_MEMORY_BUDGET_PROPERTIES_EXT;
-    memoryBudget.pNext = nullptr;
+    VkPhysicalDeviceMemoryProperties memoryProperties;
+    VkPhysicalDeviceMemoryBudgetPropertiesEXT memoryBudget = {};
 
-    VkPhysicalDeviceMemoryProperties2 dmp;
-    dmp.sType = VK_STRUCTURE_TYPE_PHYSICAL_DEVICE_MEMORY_PROPERTIES_2;
-    dmp.pNext = &memoryBudget;
+    if (memory_budget)
+    {
+        memoryBudget.sType = VK_STRUCTURE_TYPE_PHYSICAL_DEVICE_MEMORY_BUDGET_PROPERTIES_EXT;
 
-    vkGetPhysicalDeviceMemoryProperties2(*(getPhysicalDevice()), &dmp);
+        VkPhysicalDeviceMemoryProperties2 dmp = {};
+        dmp.sType = VK_STRUCTURE_TYPE_PHYSICAL_DEVICE_MEMORY_PROPERTIES_2;
+        dmp.pNext = &memoryBudget;
 
-    auto& memoryProperties = dmp.memoryProperties;
+        vkGetPhysicalDeviceMemoryProperties2(*(getPhysicalDevice()), &dmp);
+        memoryProperties = dmp.memoryProperties;
+    }
+    else
+    {
+        vkGetPhysicalDeviceMemoryProperties(*(getPhysicalDevice()), &memoryProperties);
+    }
 
     VkDeviceSize availableSpace = 0;
     for (uint32_t i = 0; i < memoryProperties.memoryTypeCount; ++i)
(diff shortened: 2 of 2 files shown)


robertosfield:
Are you able to recreate the problem in other ways?  

I'd like to create the problem on my system so I can test your PR and other possible approaches.  I've just tried on my Kubuntu 26.04 + AMD8700G system but selecting the Lavapipe driver using vsgdeviceselection --select 2 and it works.

~~~ sh
$ vsgdeviceselection models/openstreetmap.vsgt --select 2
vkEnumerateInstanceVersion() 4211029
VK_API_VERSION = 1.4.341.0
Selected vsg::PhysicalDevice ref_ptr<vsg::PhysicalDevice>(vsg::PhysicalDevice 0x78deb10f1838) llvmpipe (LLVM 21.1.8, 256 bits) deviceType = 4
~~~

Other approaches I'm curious about is making changing the default of VK_API_VERSION_1_0 to VK_API_VERSION_1_1 when creating the Vulkan instance.  Another change would be to make the supported check something that is done within vsg::Device on creation rather than on every call to availableMemory.  

I am also inclined towards to have separate code blocks for the memory available check, one with the memory budget extension and one without, rather than having ? usage sprinkled through the code.   The later is more compact but harder to interpret what is going on in each case.

mbait:
Honestly, I'm not sure how to reproduce the bug on a physical system. I've come accross the issue building my project within a CI instance that's an LXC container without a physical GPU passed through. I think running a headless VSG application within any unprivileged Linux container system will reproduce the problem.

mbait:
I'm not so familiar with the internal of either Vulkan or VulkanSceneGraph. If you say that `availableMemory` is a hot path, then sure - the check needs to be in a different place. I'm open to suggestions and hints. But in general I think that the ability to run VSG applications in headless mode, within containers is a great feature that helps projects to increase the level of confidence when producing automated builds, so I look forward for fixing the current issue that blocks that.

robertosfield:
I have been experimenting with runntime checks of VK_EXT_memory_budget extension with the following code:

~~~ cpp
   vsg::info("Device supportsApiVersion(VK_API_VERSION_1_1) = ", supportsApiVersion(VK_API_VERSION_1_1));
    vsg::info("VK_EXT_MEMORY_BUDGET_EXTENSION_NAME = ", VK_EXT_MEMORY_BUDGET_EXTENSION_NAME,", supportsDeviceExtension(VK_EXT_MEMORY_BUDGET_EXTENSION_NAME) = ", supportsDeviceExtension(VK_EXT_MEMORY_BUDGET_EXTENSION_NAME));

~~~

When I run it I see:

~~~ sh
info: Device supportsApiVersion(VK_API_VERSION_1_1) = 1
info: VK_EXT_MEMORY_BUDGET_EXTENSION_NAME = VK_EXT_memory_budget, supportsDeviceExtension(VK_EXT_MEMORY_BUDGET_EXTENSION_NAME) = 0
~~~

But... when I run `vsgdeviceselection --extensions `it lists VK_EXT_memory_budget:

~~~ sh
$ vsgdeviceselection --extensions | grep VK_EXT_memory_budget
    extensionName = VK_EXT_memory_budget, spec = 1
~~~

I don't know the reason for this discrepancy.

robertosfield:
OK, I've figured out the discrepancy,  VK_EXT_memory_budget isn't enabled by default, so the Device::supportsDeviceExtension(..) method doesn't return true.

However, this extension is working fine on my system without it being enabled so I presume the extension has been promoted or enabled by default on my NVidia drivers.

robertosfield:
I have implemented an alternative approach:

https://github.com/vsg-dev/VulkanSceneGraph/tree/availableMemory_checks

@mbait Could these this branch and if it works for your usage case I'll merged it with VSG master.

mbait:
No, that doesn't work for me - I still experience segfaults running the test app in a headless container. 

The branch's check is one half of the precondition (device extension) but misses the other (instance API version /
  KHR companion). My PR also gates on supportsApiVersion(VK_API_VERSION_1_1), which is why it survives the 1.0 case.

For a short period of time I can provide you with a remote container you will be able to ssh into if you need - just let me know. Or you can have one at your own machine with Docker or Podman. The idea is to have the Lavapipe as the default driver, because manual selection might not work as you expected.

mbait:
Reworked to follow the caching pattern from #1711 / branch `availableMemory_checks`:

- Adds `const bool Device::memory_budget` set once at the end of `Device::Device()` and re-used on every `availableMemory()` call.
- Same field name and type as `availableMemory_checks` so reconciling the two changes is mechanical.
- The precondition stored in the cache combines **both** `supportsApiVersion(VK_API_VERSION_1_1)` and `supportsDeviceExtension(VK_EXT_memory_budget)`. The API-version half is what fixes the original crash — without it, a `vsg::Instance::create(..., VK_API_VERSION_1_0)` on a driver that supports `VK_EXT_memory_budget` (e.g. Mesa Lavapipe) still dispatches `vkGetPhysicalDeviceMemoryProperties2` against an instance that never advertised support for it and crashes the driver. Verified against the same headless PoC: green with this PR, still SIGSEGVs on `availableMemory_checks` alone.

robertosfield:
The need for the API check is an oddity.  If the driver declares it supported then it should work, not have extra conditions, so I suspect the driver has a bug.

Adding a version check isn't difficult, and as I'd expect the VSG to be used with Vulkan-1.1 or later, as this is a per-requisite it shouldn't cause problems.  

However, in your version you don't request the support of the version so the supportsDeviceExtension(VK_EXT_memory_budget) fails.

After merging my changes yesterdays I decided that it was better to add the cached extension result into vsg::DeviceExtensions so I'll make that change this morning and add the API version check.

robertosfield:
I have made my changes: 61c94f5daa330a5602fee468dcf07d36ad61ee6c

@mbait Could you test the https://github.com/vsg-dev/VulkanSceneGraph/tree/availableMemory_checks branch.  If this works fine I'll merge with VSG master.

robertosfield:
I believe my changes should work to address the issue that inspired this PR so I have merged them with VSG master and will close this PR.

https://github.com/vsg-dev/VulkanSceneGraph/pull/1713

@mbait If the issue still persists with VSG master please create an Issue for it, thanks.

mbait:
Confirmed that branch `availableMemory_checks` at 61c94f5d fixes the original `vsg::Instance::create(..., VK_API_VERSION_1_0)` + Mesa Lavapipe crash that this PR was opened for. Closing as superseded — thanks @robertosfield.

PR #1737, opened 2026-08-19
Fix Descriptor count comparison when VkDescriptorSetLayoutBinding.descriptorCount is not 1

rolandhill:
A VkDescriptorSetLayoutBinding can have multiple descriptors, so the descriptorCount needs to be taken account of when comparing to the size of vsg::Descriptors. Without this, valid texture arrays get rejected. 

The current code assumes that VkDescriptorSetLayoutBinding.descriptorCount is always 1.

This fix has been tested on Kubuntu and Windows 11 with AMD 9070XT.

### DIFF
--- a/src/vsg/state/DescriptorSet.cpp
+++ b/src/vsg/state/DescriptorSet.cpp
@@ -76,9 +76,14 @@ void DescriptorSet::compile(Context& context)
             throw Exception{"Error: invalid DescriptorSet as no setLayout assigned.", VK_INCOMPLETE};
         }
 
-        if (setLayout->bindings.size() != descriptors.size())
+        size_t numDescriptors = 0;
+        for(auto& binding : setLayout->bindings)
         {
-            throw Exception{make_string("Error: invalid DescriptorSet as setLayout bindings size (", setLayout->bindings.size(), ") and descriptors size ", descriptors.size(), ") not equal."), VK_INCOMPLETE};
+            numDescriptors += binding.descriptorCount;
+        }
+        if (numDescriptors != descriptors.size())
+        {
+            throw Exception{make_string("Error: invalid DescriptorSet as setLayout bindings size (", numDescriptors, ") and descriptors size ", descriptors.size(), ") not equal."), VK_INCOMPLETE};
         }
 
         // make sure all the contributing objects are compiled


robertosfield:
While testing I found the vsgtexturearray example in vsgExamples now crashes, reverting back to commit before merging this PR fixes the crash.

The crash is in BindDescriptorSet assocaited with the pipeline layout not being assigned.  Outwardly this would seem completely unconnected to this PR, but I suspect something has gone astray with the state setup for the texture arrays example since applying this PR.  

I am now investigating.

robertosfield:
I have found the the change in checks in the DescriptorSet.cpp was trigging throwing on exception which is why vsgtexturearray was failing.  I don't think there is an error in vsgtexturearray so it looks like the new checks aren't correct for this usage case.

vsgtexturearray works fine including nothing reported from Vulkan Validation Layer with it as is, so I think the state setup is fine, so the checks you change seem inappropriate.

This leaves me wondering whether your app is also OK with it's different approach or perhaps is working regardless on being setup slightly wrong.  Running your app with Vulkan Validation Layer enabled would worth doing.

As it's not clear what the correct checks are, and that we have the Vulkan Validation Layer that can tell us what is actually wrong then using that would be better than throwing exceptions liberally so I've removed the checks completely and checked this into VSG master.

rolandhill:
There are no vulkan validation warnings. I even forced an error to check the validation layer is working.

vsgtexturearray is doing something a bit different to my application. I'm accessing all the textures for every fragment to blend them together and the textures are all different, so need different descriptors. I'm going through it now to make sure I understand.

robertosfield:
A DescriptorImage has a lit of ImageInfo, each ImageInfo has a Sampler and ImageView, so it should be possible to have a single descriptor that manages multiple textures each with their own samplers.  This is the approach vsgtexturearray takes.  Should be possible in your case as well.

rolandhill:
aarhh, an ImageInfo**List**. Suddenly it makes more sense!

 I've been creating a DescriptorImage for each texture and specifying the dstArrayElement in the constructor, then pushing them all into the Descriptors. I've been doing this for years, it has always worked on all the different GPUs I've tested and it doesn't throw any validation errors. I think it is also a valid way to specify Descriptors because the facility to specify the dstArrayElement is there.

If using a ImageInfoList is more common practice then I might try to switch to that. I don't think doing it the way I describe should throw an exception though (and I see that it no longer does in this dev release).

Thanks for the explanation.

PR #1738, opened 2026-08-20
Drag and drop back end

rolandhill:
This PR provides the VSG back end to enable drag and drop from the operating system (Linux and Windows). There is still work required within applications to take advantage of this, but you can't do it at the application level without VSG passing the events through. Applications can just ignore this and it won't affect them.

Working in Linux and Windows within the Spatial Integration application.


robertosfield:
I have merged the changes as the branch:

https://github.com/vsg-dev/VulkanSceneGraph/tree/rolandhill-DragAndDrop

I spotted that ObjectFactory.cpp hadn't been updated so I've add that with commit: bf89412c98aa434ac0c183e4cadc1e85d129644f

